PR #9, opened 2018-10-17
sys-kernel: fix COREOS_SOURCE_NAME to linux-4.19.0-flatcar

dongsupark:
Jenkins job `package-matrix` failed with the following build error:
    
```
cp: cannot stat '/build/amd64-usr/usr/lib/modules/4.19.0-coreos/build/.config': No such file or directory
* ERROR: sys-kernel/coreos-kernel-4.19::coreos failed (prepare phase):
```

Actually there's no directory `4.19.0-coreos`, only `4.19.0-flatcar`.
We need to replace `coreos` with `flatcar` in a eclass variable `$COREOS_SOURCE_NAME`.


dongsupark:
Merged upstream master as of today, 2018-10-25, which is based on kernel 4.19 release instead of -rc8.
Updated an eclass variable `COREOS_SOURCE_NAME` to fix the build error from package-matrix.

PR #11, opened 2018-10-30
sys-apps: fix baselayout commit to flatcar-master branch

dongsupark:
So far the Flatcar SDK has installed following packages from the upstream master, not the flatcar-master branch. It should actually install the `flatcar-master` branch, to avoid build errors.

* baselayout: without this fix, build fails when trying to copy `env.d/99flatcar_ldpath` from the baselayout repo.

* update_engine: without this fix, build fails with errors when trying to access to `/usr/bin/postinst`, which is symlink to `/usr/sbin/flatcar-postinst`.

* bootengine: without this fix, the systemd service `initrd-setup-root.service` fails when `/sbin/initrd-setup-root` in initrd tries to run `/usr/sbin/coreos-tmpfiles` instead of `/usr/sbin/flatcar-tmpfiles`.

### DIFF
--- a/coreos-base/update_engine/update_engine-9999.ebuild
+++ b/coreos-base/update_engine/update_engine-9999.ebuild
@@ -9,7 +9,7 @@ AUTOTOOLS_AUTORECONF=1
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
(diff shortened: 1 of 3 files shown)


dongsupark:
Pushed another commit to fix a similar issue in update_engine.

dongsupark:
Yesterday I pushed another commit to fix a similar issue in bootengine, and tested it with Flatcar SDK. With that, the SDK finally worked fine without errors.
Ready for review.

PR #14, opened 2018-11-07
sys-kernel: make KV_OUT_DIR use again COREOS_SOURCE_NAME

dongsupark:
`FLATCAR_SOURCE_NAME` does not exist. Use only `COREOS_SOURCE_NAME`.


dongsupark:
FTR, this fix was already included in the beta (1939.1.0) build in the last week.
Apparently the build went well with this PR included. I've tested the beta image, it seems to run fine with Kernel 4.14.79.

PR #17, opened 2019-01-15
profiles: add CFLAGS to prevent the Spectre v2

dongsupark:
To be able to prevent the Spectre v2, we should add the following CFLAGS `-mindirect-branch=thunk` and `-mindirect-branch-register` for the general profile.

This PR has to be tested together with https://github.com/flatcar-linux/portage-stable/pull/2.

See also:
https://security.googleblog.com/2018/01/more-details-about-mitigations-for-cpu_4.html
https://groups.google.com/forum/#!topic/linux.gentoo.user/6xUuPccmxtU


dongsupark:
It's included instead in the edge version 2121.99.0.

PR #18, opened 2019-02-19
eclass: trim suffix REVISION from COREOS_SOURCE_VERSION

dongsupark:
In case of a suffix `$COREOS_SOURCE_REVISION` expliticly given by the user, like `-r1`, `$COREOS_SOURCE_NAME` should not include the revision suffix. That's because the individual coreos-kernel*ebuild will make use of `$COREOS_SOURCE_NAME` as a base of `$KV_OUT_DIR`.

Otherwise the build will fail because `/build/amd64-usr/lib/modules/4.14.96-flatcar-r1` does not exist. More precisely, that's because `$KV_FULL` given by the upstream eclass does not include the suffix, e.g. `/build/amd64-usr/lib/modules/4.14.96-flatcar`,


dongsupark:
@iaguis Yes, that's exactly what I mean.
My local build passed with the change. Hopefully it will fix the Jenkins build as well.

PR #19, opened 2019-02-26
dev-lang/perl: fix segfault in perl-cross

dongsupark:
In developer containers or SDK, there has been a bug in cross-compiled perl that it crashes with segmentation fault. It has two different reasons:

The first issue is that perl depends on perl-cross 1.1.4, which has a bug with gcc 7.x or newer, so that the necessary CFLAGS `-fwrapv` was not appended. This issue was fixed in perl-cross 1.1.9 or newer. So we just need to upgrade perl-cross to 1.2.2, the newest release.

```
  CROSS_VER=1.2.2
```

The other issue is that the CFLAGS `-O2` was given when building cross-compiled perl. We should replace the optimization flag with `-O0`, when it's built by the cross compiler. Without that, the segfault will not be fixed at all.

```
  if tc-is-cross-compiler; then
      replace-flags -O? -O0
  fi
```

How to test inside the Flatcar SDK environment:

```
LD_PRELOAD=/build/amd64-usr/usr/lib/libperl.so.5.24.1 /build/amd64-usr/usr/bin/perl -V
```

See also https://github.com/coreos/bugs/issues/2369
https://github.com/coreos/bugs/issues/2545
https://github.com/arsv/perl-cross/issues/60


iaguis:
Cool debugging!

Have you checked if updating to the [latest upstream ebuilds](https://github.com/flatcar-linux/coreos-overlay/pull/19) and upgrading perl works? It seems perl 5.28.0 uses perl-cross 1.2.2 and that approach seems a bit cleaner to me than moving perl to coreos-overlay.

I also see that the current perl ebuild doesn't have the `-O0` fix, so maybe upstream's ebuild doesn't work. In that case it'd be nice to contribute a fix upstream. What do you think?

Related: https://github.com/gentoo/gentoo/pull/10976

dongsupark:
I've already tried to upgrade perl to 5.26.
I didn't work well, because so many other perl packages depend on the exact version of perl, 5.24.1-r2. Updating all of these would make this PR much bigger.
I just wanted to fix only the particular issues without touching other parts.

dongsupark:
Filed a PR against upstream Gentoo, to turn off `-O?` options. https://github.com/gentoo/gentoo/pull/11171

On the other hand, upstream Container Linux already includes a recent version of perl 5.26.2 with keyword `amd64` with `CROSS_VER=1.1.9`, while the ebuild in our flatcar-master branch still stays with keyword `~amd64`.

So it would be a better approach to simply merge upstream master into flatcar-master, without changing downstream perl. I'm not sure what that change would suddenly break other builds. Anyway the critical releases were just gone, so it might be a good moment to be in sync with upstream, right? :-)

dongsupark:
So I created 2 draft PRs that merge upstream into flatcar-master, and update perl packages.

* https://github.com/flatcar-linux/portage-stable/pull/3
* https://github.com/flatcar-linux/coreos-overlay/pull/20

With the 2 PRs, we can build an image with perl 5.26.2. On top of that, if the upstream Gentoo PR got merged, it's possible to fix the segfault issue.

Unfortunately It's not possible to avoid downstream changes completely. Some perl packages still rely on old version of perl.

So I'm not sure if that's a better approach.

dongsupark:
Not needed any more.

PR #20, opened 2019-02-28
Merge upstream 2019-02-28 + update perl packages

dongsupark:
To make the build chain rely on perl 5.26.2 instead of perl 5.24, we should merge upstream.

As several perl subpackages depend on perl 5.24.x, we need to make them depend perl 5.26.2, to make the whole build work.

(Draft PR. Let's not merge yet.)


dongsupark:
Not needed any more

PR #23, opened 2019-04-17
edge: add cgroupid and patch runc for OCI hooks

alban:
- New package: https://github.com/kinvolk/cgroupid
- Updated runc & docker-runc: with [static hooks](https://github.com/kinvolk/runc/tree/alban/static-hooks)


alban:
It compiles fine for me in the SDK.

PR #32, opened 2019-05-08
sys-kernel: add wireguard patch

iaguis:
For edge.


dongsupark:
Oh, are we doing it? Finally? :+1: 
I was thinking about upgrading kernel to 5.1. But anyway I can deal with that after merging this PR.
Will try to test this PR in this week.

PR #33, opened 2019-05-09
app-emulation/cri-o: bump cri-o to 1.13.7

dongsupark:
Bump cri-o to 1.13.7 to fix problems with the upstream tarball 1.13.5.

Now that we follow the way of upstream Gentoo to remove `-i` from go build commands, we don't need to keep a separate patch file.


dongsupark:
Tested under the local SDK env. Works.

PR #34, opened 2019-05-09
sys-kernel: bump kernel to 5.1.0

dongsupark:
Upgrade kernel to 5.1.0, and adjust corresponding patches.


dongsupark:
Tested under the local SDK env. Works.

PR #35, opened 2019-05-10
net-vpn: add a userspace tool for wireguard

dongsupark:
In addition to the wireguard kernel modules, we need to add wireguard tool in userspace as well.
Import its upstream ebuilds, and remove unnecessary parts like kernel modules and docs.


dongsupark:
Tested.

iaguis:
Hm, I wanted to bring up whether it makes sense to have this as a part of Flatcar or if just using a container would be fine for users (see for example https://blog.jessfraz.com/post/installing-and-using-wireguard/). Always hesitant about adding tools for the sake of adding them, it might make sense in this case but I'd like to have a discussion about it.

iaguis:
Also, systemd-networkd has support for wireguard since v237.

dongsupark:
Ok, if you think so, we can remove the wireguard tool from the image in the future. Either way I'm fine.

PR #36, opened 2019-05-27
sys-kernel: fix kernel build path issues to build coreos-firmware

dongsupark:
In order to build `coreos-firmware` correctly, we need to set `$KERNEL_DIR` and `$KBUILD_OUTPUT` before doing `inherit linux-info`. Otherwise `$KV_FULL` ends up being an invalid path, and as a result, `coreos-firmware` becomes an empty package.

Also we should make `coreos-firmware` depend on `coreos-sources` and `coreos-modules`, to be able to get a correct list of kernel modules.

We should exclude files that does include a space character `' '` from the firmwares list, to avoid kola tests errors like the following.

```
    --- FAIL: cl.filesystem/blacklist (0.20s)
             files.go:210: Blacklisted files or directories found:
    /usr/lib64/firmware/brcm/brcmfmac43430a0-sdio.ONDA-V80 PLUS.txt
```

### DIFF
--- a/sys-kernel/coreos-firmware/coreos-firmware-99999999.ebuild
+++ b/sys-kernel/coreos-firmware/coreos-firmware-99999999.ebuild
@@ -2,6 +2,10 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI="6"
+
+# Tell linux-info where to find the kernel source/build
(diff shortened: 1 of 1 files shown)


dongsupark:
Added another commit to fix kola tests failing because of unnecessary text files under the firmware directory.

PR #41, opened 2019-06-26
app-emulation/containerd: make containerd listen on localhost

dongsupark:
In the past there had been an issue of containerd listening on a public address with TCP port 10010.
The issue was fixed in all channels, by [disabling the gRPC plugin of containerd](https://github.com/coreos/bugs/issues/2524).
However, Flatcar edge [enabled the gRPC plugin again](https://github.com/flatcar-linux/coreos-overlay/commit/8184af2518634c115dd6ef623959da5948be4cca), to support cgroup v2 for containerd.
As a result, the original issue of containerd happened again on Flatcar edge.

To fix that, we should make containerd listen on localhost 127.0.0.1, and a TCP port 10010. See also [containerd/cri docs](https://github.com/containerd/cri/blob/53c2230ec09d3672dbfb39756954b1a4afdb1ab8/docs/config.md).
Also bump containerd to 1.1.2-r3.


dongsupark:
Wait. I've just realized, it's not a good idea to use the dynamic allocated port. If the port number becomes different, the [kola test to check for network listeners](https://github.com/flatcar-linux/mantle/blob/3db1be562bd9fe814b80ca33a13f2f4edbdfca50/kola/tests/misc/network.go#L126) cannot work. The current kola test requires a static port number.
I will change `stream_server_port` back to `10010`.

PR #53, opened 2019-07-26
app-emulation: disable temporarily selinux for runc v1.0.0-rc8

dongsupark:
To fix regression caused by selinux with runc v1.0.0-rc8, we should temporarily disable selinux checks for runc.


app-emulation/docker-runc/files/0001-temporarily-disable-selinux.GetEnabled-error-checks.patch
```diff
@@ -0,0 +1,58 @@
...
+-	selinux "github.com/opencontainers/selinux/go-selinux"
+ )
+ 
+ type Validator interface {
+@@ -99,9 +98,6 @@ func (v *ConfigValidator) security(config *configs.Config) error {
+ 		!config.Namespaces.Contains(configs.NEWNS) {
+ 		return fmt.Errorf("unable to restrict sys entries without a private MNT namespace")
+ 	}
+-	if config.ProcessLabel != "" && !selinux.GetEnabled() {
+-		return fmt.Errorf("selinux label is specified in config, but selinux is disabled or not supported")
```
invidian: `config.ProcessLabel` hmm, maybe we could modify the configuration rather than changing the code?

pothos: Maybe that would be just removing the `+selinux` line when building, don't know and it takes long to test such a patch. Actually it would be good to enable SELinux in Flatcar.

pothos: By keeping the support compiled, we can try if it works when the files get the correct labels.

dongsupark: About changing the config, I'm not sure how to do that, because the runtime config is generated outside of runc, in this case by docker.
So we need to somehow patch docker as well. Not so simple.

Removing selinux build tags should be also done for docker as well.

So for now I have no better idea, I just needed a simple fix. ;-)

invidian: Thanks for additional context @dongsupark, your patch seems reasonable then, if it doesn't break other things :+1:

PR #54, opened 2019-07-26
kubelet-wrapper: replace rkt with docker for stable/beta/alpha

dongsupark:
Replace rkt with docker in kubelet-wrapper, and adjust corresponding options, to avoid recent security issues.

* change rkt volume/mount options to docker option `--mount`, and adjust bind propagation options accordingly.
* Update the default `KUBELET_IMAGE` to `gcr.io/google-containers/hyperkube`, to be up-to-date with recent kubernetes versions.
* append docker run options to add capability & SELinux options.
* Update systemd unit files to make them runnable with docker.
  * Use simple type for systemd unit, as docker cannot work with notify.
  * Give a specific container name like `etcd-docker` by using `--name=` option, as docker does not have the uuid option like rkt.
  * Use simple type as docker does not run with notify.
* Update RDEPEND to include docker.

This PR should be merged together with https://github.com/flatcar-linux/toolbox/pull/1.


dongsupark:
> Which channel is this going to and how did you test?

My original plan was to just get this in for all channels from stable to edge.
I tested with simple use cases with stable, but it was July. We should test again with latest versions.

> does it work with OpenEBS volumes? 
> I see some rprivate mounts

Good points. We should look into the cases of OpenEBS and mount propagation modes.

dongsupark:
Fixed bugs in RDEPEND, which should actually not include docker, as only torcx should install docker-related binaries.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -31,59 +31,48 @@ if [[ -n "${KUBELET_VERSION}" ]]; then
...
+	echo RKT_RUN_ARGS environment variable is deprecated, this script uses Docker now, please use DOCKER_RUN_ARGS instead and pass Docker options
 fi
 
 KUBELET_IMAGE_TAG="${KUBELET_IMAGE_TAG:-${KUBELET_VERSION}}"
 
 require_ev_one KUBELET_IMAGE KUBELET_IMAGE_TAG
 
-KUBELET_IMAGE_URL="${KUBELET_IMAGE_URL:-${KUBELET_ACI:-quay.io/coreos/hyperkube}}"
+KUBELET_IMAGE_URL="${KUBELET_IMAGE_URL:-${KUBELET_ACI:-gcr.io/google-containers/hyperkube}}"
 KUBELET_IMAGE="${KUBELET_IMAGE:-${KUBELET_IMAGE_URL}:${KUBELET_IMAGE_TAG}}"
```
alban: It seems that this patch is applied on Flatcar Edge.

With Lokomotive, we have:
```
$ cat /etc/kubernetes/kubelet.env
KUBELET_IMAGE_URL=docker://k8s.gcr.io/hyperkube
KUBELET_IMAGE_TAG=v1.16.2
```

See [worker.yaml.tmpl](https://github.com/kinvolk/lokomotive-kubernetes/blob/4754b5b3256b069b2651c03c536e7823423005d8/aws/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl#L102-L103)

Notice the "docker://" prefix.

Then, the kubelet fails to start:

```
Nov 24 15:33:25 ip-10-0-32-204 kubelet-wrapper[9825]: + exec /usr/bin/docker run --cap-add SYS_ADMIN --security-opt label:disable --mount type=bind,bind-propagation=rprivate,source=/etc/kubernetes,target=/etc/kubernetes --mount type=bind,bind-propagation=rprivate,source=/etc/ssl/certs,target=/etc/ssl/certs,readonly --mount type=bind,bind-propagation=rprivate,source=/usr/share/ca-certificates,target=/usr/share/ca-certificates,readonly --mount type=bind,bind-propagation=rslave,source=/var/lib/docker,target=/var/lib/docker --mount type=bind,bind-propagation=shared,source=/var/lib/kubelet,target=/var/lib/kubelet --mount type=bind,bind-propagation=rprivate,source=/var/log,target=/var/log --mount type=bind,bind-propagation=rprivate,source=/usr/lib/os-release,target=/usr/lib/os-release,readonly --mount type=bind,bind-propagation=rprivate,source=/run,target=/run --mount type=bind,bind-propagation=rprivate,source=/sys,target=/sys --mount type=bind,bind-propagation=rprivate,source=/lib/modules,target=/lib/modules,readonly --mount type=bind,bind-propagation=rprivate,source=/etc/machine-id,target=/etc/machine-id,readonly docker://k8s.gcr.io/hyperkube:v1.16.2 /kubelet --anonymous-auth=false --authentication-token-webhook --authorization-mode=Webhook --client-ca-file=/etc/kubernetes/ca.crt --cluster_dns=10.3.0.10 --cluster_domain=cluster.local --cni-conf-dir=/etc/kubernetes/cni/net.d --config=/etc/kubernetes/kubelet.config --exit-on-lock-contention --kubeconfig=/etc/kubernetes/kubeconfig --lock-file=/var/run/lock/kubelet.lock --network-plugin=cni --node-labels=node.kubernetes.io/node --pod-manifest-path=/etc/kubernetes/manifests --read-only-port=0 --volume-plugin-dir=/var/lib/kubelet/volumeplugins
Nov 24 15:33:25 ip-10-0-32-204 kubelet-wrapper[9825]: /run/torcx/bin/docker: invalid reference format.
Nov 24 15:33:25 ip-10-0-32-204 kubelet-wrapper[9825]: See '/run/torcx/bin/docker run --help'.
Nov 24 15:33:25 ip-10-0-32-204 systemd[1]: kubelet.service: Main process exited, code=exited, status=125/n/a
Nov 24 15:33:25 ip-10-0-32-204 systemd[1]: kubelet.service: Failed with result 'exit-code'.
```

because docker does not accept the "docker://" prefix in the command line.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -31,59 +31,48 @@ if [[ -n "${KUBELET_VERSION}" ]]; then
...
+	echo KUBELET_ACI environment variable is deprecated. This script uses Docker now, please use KUBELET_IMAGE_URL instead and pass a Docker image name
 fi
 
 if [[ -n "${RKT_OPTS}" ]]; then
-	echo RKT_OPTS environment variable is deprecated, please use the RKT_RUN_ARGS instead
+	echo RKT_OPTS environment variable is deprecated, this script uses Docker now, please use DOCKER_RUN_ARGS instead and pass Docker options
+fi
+
+if [[ -n "${RKT_RUN_ARGS}" ]]; then
+	echo RKT_RUN_ARGS environment variable is deprecated, this script uses Docker now, please use DOCKER_RUN_ARGS instead and pass Docker options
```
alban: Lokomotive also fails with this change because $RKT_RUN_ARGS is ignored. Symptoms:
```
Nov 24 16:55:30 ip-10-0-45-152 kubelet-wrapper[21293]: RKT_RUN_ARGS environment variable is deprecated, this script uses Docker now, please use DOCKER_RUN_ARGS instead and pass Docker options
...
Nov 24 16:55:31 ip-10-0-45-152 kubelet-wrapper[21293]: F1124 16:55:31.580198       1 kubelet.go:1413] failed to start OOM watcher open /dev/kmsg: no such file or directory
Nov 24 16:55:31 ip-10-0-45-152 systemd[1]: kubelet.service: Main process exited, code=exited, status=255/EXCEPTION
Nov 24 16:55:31 ip-10-0-45-152 systemd[1]: kubelet.service: Failed with result 'exit-code'.
```

I am wondering if Flatcar needs to support both kubelet-wrapper with rkt and Docker for a little while, so that Typhoon, Lokomotive and other Kubernetes that depend on that script have some time to adapt, since it's a breaking change.

Otherwise, Kubernetes distros would need to somehow detect whether to use RKT_RUN_ARGS or DOCKER_RUN_ARGS.

dongsupark:
As for the etcd-wrapper part, please see https://github.com/kinvolk/coreos-overlay/pull/857 instead.

PR #71, opened 2019-09-06
sys-apps/systemd: bump to systemd v242 and adjust ebuilds

dongsupark:
Make the ebuild work with systemd v242.

* Remove unnecessary checks for enabled units under `/etc/systemd/system`, because [upstream commit](https://github.com/systemd/systemd/commit/01d2041e41f4886a6dff16a53a950ae8d5b66bc7) changed the behavior of systemd, so that it does not enable units any more under `/etc/systemd/system`.
* Import some code from upstream Gentoo.
* update commit ID to include v242-flatcar changes

This PR should be merged together with https://github.com/flatcar-linux/systemd/pull/4.


pothos:
When I use this with Lokomotive through libvirt the network does not come up.

pothos:
Hm, I also can't log in when using the qemu script anymore, but I had tested this before, strange.

pothos:
Yes, `systemd-resolved` and `systemd-networkd` are not enabled.

pothos:
I will try a build with this:

```
diff --git a/sys-apps/systemd/systemd-9999.ebuild b/sys-apps/systemd/systemd-9999.ebuild
index dcfed77e7..7b868cfa9 100644
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -381,6 +381,11 @@ multilib_src_install_all() {
        # Don't set any extra environment variables by default
        rm "${ED}/usr/lib/environment.d/99-environment.conf" || die
 
+       for f in systemd-networkd.service systemd-resolved.service
+       do
+               dosym "${f}" "${unitdir}/multi-user.target.wants/${f}"
+       done
+
        # Do not enable random services if /etc was detected as empty!!!
        rm "${ED}$(usex split-usr '' /usr)/lib/systemd/system-preset/90-systemd.preset" || die
        insinto $(usex split-usr '' /usr)/lib/systemd/system-preset
```

pothos:
The above worked, next try is:
```
diff --git a/sys-apps/systemd/systemd-9999.ebuild b/sys-apps/systemd/systemd-9999.ebuild
index dcfed77e7..40ce60459 100644
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -381,6 +381,9 @@ multilib_src_install_all() {
        # Don't set any extra environment variables by default
        rm "${ED}/usr/lib/environment.d/99-environment.conf" || die
 
+       systemd_enable_service multi-user.target.wants systemd-networkd.service
+       systemd_enable_service multi-user.target.wants systemd-resolved.service
+
        # Do not enable random services if /etc was detected as empty!!!
        rm "${ED}$(usex split-usr '' /usr)/lib/systemd/system-preset/90-systemd.preset" || die
        insinto $(usex split-usr '' /usr)/lib/systemd/system-preset
```

pothos:
Interesting, with `systemd_enable_service` they got started and then they terminate. It works but `networkctl` says `unmanaged`.

pothos:
Have force-pushed the fix.

dongsupark:
@pothos Yes, you change looks good to me. Thanks!
I'm still not completely sure if this is the only way.
I have expected the ebuild from Gentoo upstream to work better than it actually is.
Anyway it's only a change in Edge channel, we still have chances for further improvements.
Let's merge it.

PR #74, opened 2019-09-10
sys-apps/dbus: fix authentication bypass issue in D-Bus

dongsupark:
To fix the security issue about authentication bypass, [CVE-2019-12749](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-12749), backport the patch to dbus 1.10.18.

See also https://github.com/coreos/coreos-overlay/pull/3761,
https://security.gentoo.org/glsa/201909-08


dongsupark:
This dbus fix has been already available in the recent beta & alpha releases.
So we should merge it into edge as well to address the security issue.

PR #77, opened 2019-09-16
Use mantle/ignition versions which have the opt/org.flatcar-linux Ignition variable

pothos:
See https://github.com/flatcar-linux/mantle/pull/26 and https://github.com/flatcar-linux/scripts/pull/22 which in combination with this PR do what https://github.com/flatcar-linux/ignition/issues/2 did for stable/beta/alpha.
Because https://github.com/flatcar-linux/ignition/pull/4 is not merged yet, this is effectively a downgrade of Ignition. Consider to delay merging this until it can use the latest commit after https://github.com/flatcar-linux/ignition/pull/4 got merged.


pothos:
Updated to point to https://github.com/flatcar-linux/ignition/pull/4 for testing

pothos:
Changed to point to https://github.com/flatcar-linux/ignition/pull/5

PR #87, opened 2019-09-30
bump cri-o to 1.15.2

simonfuhrer:
should fix kubelet 1.16.0 crio-o grpc connection error


invidian:
:+1:, that seems recommended when running kubelet 1.15.X

dongsupark:
Thanks!

PR #88, opened 2019-10-07
Support for ARM architecture again in 2019

dongsupark:
This PR consists of most changes to support ARM architecture for Flatcar Linux.

A bunch of changes are about reverting old commits that dropped ARM architecture, back in mid 2018.
After that, we had to update packages as well, for example, kernel or rust packages.
It's mainly being tested on Packet.net.

There could be still rooms for improvements.
So please review, test, and let us know.

Should be merged together with https://github.com/flatcar-linux/bootengine/pull/4 https://github.com/flatcar-linux/scripts/pull/23 https://github.com/flatcar-linux/mantle/pull/28.


pothos:
For the amd64 image `coreos-metadata-sshkeys@core.service` failed with `Error: initialization` `coreos-metadata[804]: Caused by: Couldn't find flag 'flatcar.oem.id' in cmdline file (/proc/cmdline)`.

pothos:
I reverted the commit that worked around the ignition problem.

dongsupark:
FTR, yes, we should revert the commit for coreos-metadata-sshkeys.
Though as I tested today, without manually enabling coreos-metadata-sshkeys, ARM instance on Packet does not configure ssh keys at all. It seems initramfs does not read the given ignition config at all.
We need to look into the ignition issue, before we could merge this PR.

pothos:
The OEM packages are not part of the stable PXE images, so we don't need to worry about `coreos-metadata-sshkeys@core.service` not being enabled.
After the iPXE image was used to install to disk with, e.g., `-o packet`, then `coreos-metadata-sshkeys@core.service` is enabled.

dongsupark:
All right. Thanks for testing it.
Then I will squash unnecessary commits, so we can merge this PR.

dongsupark:
Kai, your fixes look good, according to my local ARM builds.
Let's merge this PR next week, when the release waves are gone.

pothos:
Squashed some commits.

PR #101, opened 2019-10-28
Fix CROS_WORKON_COMMIT values for multiple repos

dongsupark:
We still have multiple places where `CROS_WORKON_COMMIT` still points to an old upstream tag. That has been no problem because so far ordinary channels from stable to alpha simply followed upstream changes.

However, since we started directly updating Edge changes by comparing diff from a base branch and flatcar-master, the hidden issue of wrong `CROS_WORKON_COMMIT`s actually started playing a role. That's why Edge channels still have old version of certain packages.

To get rid of potential issues, let's replace every possible `CROS_WORKON_COMMIT` with the latest commit of `flatcar-master`.

Note, some repos are not updated, as the current commit is not different from flatcar-master. We had to also update cargo dependencies, pulled by newer versions of rust programs.


dongsupark:
Locally tested, it looks good.
These changes are needed for the next Edge release.

PR #103, opened 2019-10-31
kubelet-wrapper: replace rkt with docker for Edge

dongsupark:
Replace rkt with docker in kubelet-wrapper, and adjust corresponding options, to avoid recent security issues.

* change rkt volume/mount options to docker option `--mount`, and adjust bind propagation options accordingly.
* Update the default `KUBELET_IMAGE` to `gcr.io/google-containers/hyperkube`, to be up-to-date with recent kubernetes versions.
* append docker run options to add capability & SELinux options.
* Update systemd unit files to make them runnable with docker.
  * Use simple type for systemd unit, as docker cannot work with notify.
  * Give a specific container name like `etcd-docker` by using `--name=` option, as docker does not have the uuid option like rkt.
  * Use simple type as docker does not run with notify.
* Update RDEPEND to include docker.

This PR is the same as https://github.com/flatcar-linux/coreos-overlay/pull/54, except that it's based on Edge.


dongsupark:
Fixed bugs in RDEPEND, which should actually not include docker, as only torcx should install docker-related binaries.

PR #105, opened 2019-11-11
Revert "coreos-base/afterburn: fix CROS_WORKON_COMMIT to flatcar-master"

dongsupark:
This reverts commit 818cf3c6abc996a842d3ec21a448dc8f80b0da0e.

Upgrade of afterburn caused build errors during releases of stable and etc, because the upstream still has dependencies on older versions of crates.
So we need to revert the change to make the ordinary build workflow work again.


coreos-base/afterburn/afterburn-9999.ebuild
```diff
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ ${PV} == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm64"
 else
-	CROS_WORKON_COMMIT="56549117e7ac1941e41ff0274da6950617377fff" # flatcar-master
+	CROS_WORKON_COMMIT="d5de619f19e3f4ad20a9d2e5803ba0353618f00c" # v4.0.0
```
invidian: Hmm, why not update to 4.2.0? https://github.com/coreos/afterburn/blob/v4.2.0/Cargo.lock

dongsupark: In theory I would like to update such dependencies to the newest releases like that. However, it turns out, specific build-xyz branches from upstream appear to depend on older cargo crates. As long as we still follow upstream for stable, beta, and alpha, we need to slow down with version updates.

dongsupark:
Reverted update-ssh-keys as well.

dongsupark:
Oh, yesterday I forgot to increase revisions.
Updated revisions.

PR #123, opened 2019-11-26
net-libs/rpcsvc-proto: make rpcgen work for both architectures

dongsupark:
`rpcsvc-proto` started failing due to a missing binary `rpcgen`, during the Jenkins pipeline for SDK.

We need to have a fallback mechanism for trying both the built-in rpcgen and the host rpcgen.


dongsupark:
Tested. It looks good.

PR #124, opened 2019-11-27
[WIP] profiles: enable sssd only for sudo on amd64 to avoid talloc failures

dongsupark:
_Work-in-progress: still testing_

Jenkins pipeline for arm64 started failing with `sys-libs/talloc`, which depends on the build system `waf`. When running `configure` script, `waf` tries to run a cross-compiled binary, which fails with `Exec format error`.

Unfortunately waf is not able to deal with running tests with cross-built binaries, when no `qemu-aarch64-static` is installed on the host system. Since the Jenkins worker is a minimal Flatcar system, we cannot simply install a qemu static package on the system. Also there is no simple way to disable running such tests in talloc and waf.

So for now, we should enable `sssd` for `sudo` only for `amd64`, not for `arm64`. Then sudo for arm64 will not pull the new dependencies like `talloc`.

### DIFF
--- a/profiles/coreos/amd64/generic/package.use
+++ b/profiles/coreos/amd64/generic/package.use
@@ -15,3 +15,5 @@ app-arch/tar                selinux
 # Only ship microcode currently distributed by Intel
 # See https://bugs.gentoo.org/654638#c11 by iucode-tool maintainer
 sys-firmware/intel-microcode vanilla
+
+app-admin/sudo		    sssd
--- a/profiles/coreos/targets/generic/package.use
+++ b/profiles/coreos/targets/generic/package.use
@@ -1,7 +1,7 @@
 # Copyright (c) 2009 The Chromium OS Authors. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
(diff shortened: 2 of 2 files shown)


pothos:
The alternative would be creating the file `/etc/binfmt.d/qemu-aarch64.conf` on the host with the content:
```
:qemu-aarch64:M::\x7fELF\x02\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\xb7:\xff\xff\xff\xff\xff\xff\xff\x00\xff\xff\xff\xff\xff\xff\xff\xff\xfe\xff\xff:/usr/bin/qemu-aarch64-static:
```
but instead of `/usr/bin/qemu-aarch64-static` using the path to the SDK's `/usr/bin/qemu-aarch64`.

dongsupark:
All right.
I will try setting binfmt like that, after the Jenkins finished building everything. It will take ~10 hours for both amd and arm.

pothos:
It's not urgent but I just expect these things to come up in the future again otherwise.

dongsupark:
This change is probably not needed. Closing.

PR #122, opened 2019-11-26
Update flatcar-master-edge

pothos:
Merge flatcar-master into flatcar-master-edge.

PR #125, opened 2019-11-29
dev-lang/rust: store shell scripts under source dir instead of /tmp

dongsupark:
When building ebuilds under catalyst stage4, emerge cannot write data outside of its chroot, under `/tmp` on the host.
That's the reason why the rust installation from toolchain fails like `/tmp/cc.sh: Permission denied`.

That is no issue when building it under local SDK, because in that case there is no toolchain installation involved.

So we should write the temporary scripts under `${S}`, for example, `/mnt/host/source/src/build/catalyst/tmp/coreos-toolchains/stage4-amd64-2303.2.1/`.

### DIFF
--- a/dev-lang/rust/rust-1.36.0.ebuild
+++ b/dev-lang/rust/rust-1.36.0.ebuild
@@ -219,14 +219,14 @@ src_configure() {
 		fi
 	done
 	if [ -f /usr/bin/aarch64-cros-linux-gnu-gcc ]; then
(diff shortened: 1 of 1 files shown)


pothos:
Have ported it to https://github.com/flatcar-linux/coreos-overlay/pull/122/commits/2aa3b9e01b0e63bcf26e96e37d46dea53bc7e72f

PR #127, opened 2019-12-03
profiles: add libselinux keywords to the base profile

dongsupark:
So far libselinux was not installed for arm64, because it was simply masked.

We need to enable keywords for both libselinux and libsepol, to correctly build arm64 images including SELinux libs and binaries.


pothos:
Note, to enable it we need to duplicate `profiles/coreos/amd64/sdk/package.use` and `profiles/coreos/amd64/generic/package.use` for arm64 or move their content to a shared location.

PR #135, opened 2019-12-04
profiles: Move selinux usage tags from amd64 to all targets

pothos:
Enables SELinux for arm64.

**Note:** Should be done for Edge as well.
Needs qemu-aarch64 on the Jenkins worker.


pothos:
Tried to use `targets/generic` but the use flag didn't get picked up correctly, so I switched to `base` and will test again.

pothos:
Use flags are still not set for arm64, but anyway there is a build issue for `app-admin/setools-3.3.8-r7`

pothos:
Build issue fixed for `app-admin/setools-3.3.8-r7`

pothos:
TODO: The packages are installed but SELinux is still disabled.

PR #136, opened 2019-12-04
profiles: add libselinux keywords to the base profile for edge

dongsupark:
So far libselinux was not installed for arm64, because it was simply masked.
We need to enable keywords for both libselinux and libsepol, to correctly build arm64 images including SELinux libs and binaries.


pothos:
Note: Needs the changes from https://github.com/flatcar-linux/coreos-overlay/pull/135 to actually install SELinux packages.

dongsupark:
Ok, I will not merge this PR for the upcoming release 2345.99.0.

dongsupark:
outdated

PR #154, opened 2019-12-20
app-admin/sudo: bump sudo to 1.8.31 for edge

dongsupark:
Since sudo 1.8.28, every sudo started printing out a warning like `/etc/environment: No such file or directory`, when `/etc/environment` does not exist.

Also sudo <= 1.8.30 is affected by a pwfeedback vulnerability, CVE-2019-18634. https://seclists.org/oss-sec/2020/q1/48

Update sudo to 1.8.31 from upstream Gentoo, to resolve the issues.

See also https://bugs.gentoo.org/698946.


dongsupark:
Updated sudo to 1.8.31.

PR #161, opened 2020-01-28
sys-kernel: enable kernel config CONFIG_DEBUG_INFO_BTF for edge

dongsupark:
Compile kernel with debug info for BTF, BPF Type Format.

Fixes https://github.com/flatcar-linux/Flatcar/issues/13


dongsupark:
Please do not merge, as this PR causes the kernel build to fail like this:
```
+ printf '  %-7s %s\n' BTF '.tmp_vmlinux.btf: pahole (pahole) is not available'
  BTF     .tmp_vmlinux.btf: pahole (pahole) is not available
+ return 1
+ echo 'Failed to generate BTF for vmlinux'
Failed to generate BTF for vmlinux
+ echo 'Try to disable CONFIG_DEBUG_INFO_BTF'
Try to disable CONFIG_DEBUG_INFO_BTF
```

It happens because Flatcar devel environment does not include `pahole`, a part of `dev-utils/dwarves`.
So we need to first get `dev-utils/dwarves` and its dependencies included in the Flatcar SDK. That should be the first release.
After that we need to actually enable `CONFIG_DEBUG_INFO_BTF`, based on the new Flatcar SDK.

dongsupark:
Now that the latest Edge 2387.99.0 includes dwaves/pahole in its [SDK](https://storage.googleapis.com/flatcar-jenkins/sdk/amd64/2387.99.0/flatcar-sdk-amd64-2387.99.0.tar.bz2), we are now able to get this PR merged for the next Edge release.

pothos:
Should we pick this for https://github.com/flatcar-linux/coreos-overlay/pull/175 and close here? Or we rebase this here later on 5.5.

dongsupark:
Interesting, when I simply merged this PR to my testing branch, it did not fail with merge conflicts. Auto merge worked magically, since there is only a single commonconfig file. :-)
I could also see that BTF kernel config was enabled on the Edge VM.
Anyway just to be sure, I am going to close this PR and append the commit to https://github.com/flatcar-linux/coreos-overlay/pull/175.

pothos:
Ah, I assumed there would be conflicts, then the only benefit would be of testing it once together with the 5.5 kernel update.

dongsupark:
Oops, merged it by accident.
Removed the BTF commit from flatcar-master-edge.

PR #162, opened 2020-01-29
app-arch/torcx: Fix wrong profile reference

pothos:
When /etc/flatcar/docker-1.12 contains "no"
the profile for 18.06 was searched for instead
of 19.03. Keep docker-1.12-no.json in sync
with the latest version in app-torcx/docker/files/.

**Note:** Should be cherry-picked for alpha flatcar-build-2387(.0.0) as well.


pothos:
Done.

PR #163, opened 2020-01-30
Use rkt again instead of docker in etcd/flannel-wrappers for Edge

dongsupark:
We replaced rkt with docker via https://github.com/flatcar-linux/coreos-overlay/pull/103 for Edge, and later got it back to docker via https://github.com/flatcar-linux/coreos-overlay/pull/148.
However, we did not revert commits for other wrappers, like etcd-wrapper or flannel-wrapper.
Let's do the revert for other wrappers as well, to unbreak issues around etcd, which happened with Lokomotive clusters recently.


pothos:
(Toolbox stays using Docker.)

PR #170, opened 2020-02-06
Merge upstream 2020-02-06 and revert EOL messages

dongsupark:
Merge upstream `master branch` to `flatcar-master`, and revert EOL commits from upstream.
We need to explicitly do that, to avoid various failures during the Flatcar release process.


dongsupark:
Tested the PR, and confirm it works fine.

pothos:
Ignition's `flatcar-master` and the ebuild commit should be updated to 0.34.0.

dongsupark:
Updated commits for init and igntion.

PR #175, opened 2020-02-10
sys-kernel: bump kernel to 5.5.2 for edge

dongsupark:
Update Kernel to the latest 5.5.2.

* Enabled a kernel module QEDE, just like upstream.
* Fixed K_BASE_VER to 5.5
* Removed invalid kernel configs that are not supported in 5.5.
* Update coreos-firmware to 20200122, to make it work with kernel 5.5.
* Rename symlinks source version from 1.24.3.0 to 1.24.11.0, to make it work with the new firmware version.


pothos:
`sys-kernel/coreos-sources/coreos-sources-5.5.2.ebuild` still has `K_BASE_VER="5.4"`

dongsupark:
Yes, good catch.
Today I found the wrong kernel version as well, and fixed it.
Then it uncovers more issues, as always, so I fixed them too.
Now I am building and testing the whole edge image.

dongsupark:
Updated PR.

* Fixed K_BASE_VER to 5.5
* Removed invalid kernel configs that are not supported in 5.5.
* Update coreos-firmware to 20200122, to make it work with kernel 5.5.
* Rename symlinks source version from 1.24.3.0 to 1.24.11.0, to make it work with the new firmware version.

dongsupark:
Added another commit for BTF, coming from https://github.com/flatcar-linux/coreos-overlay/pull/161

PR #184, opened 2020-02-20
coreos-base/oem-rackspace: Get public and private IP based on routes

pothos:
The OEM package does not set net.ifnames=0 in the kernel command line
(as, for example, oem-ec2-compat does) to ensure that ifnames are ethX.

Regardless of the ifname scheme, get the public and private IPs
based on the routes that get set up.
https://support.rackspace.com/how-to/basic-network-troubleshooting/
shows that the private IPs are 10.x.x.x with 10.176.32.1 as gateway.

Note: Should be picked for alpha and edge, too.


pothos:
This PR is not really related to https://github.com/flatcar-linux/bootengine/pull/6 because this script here runs after the initramfs has finished, so it uses the persistent names anyway unless the interface types are excluded (e.g., because they are virtio interfaces). Nevertheless I think we should not hardcode eth0 and eth1, thus the fix.

Disclaimer: I did not test it on Rackspace directly.

dongsupark:
It needs a rebase

PR #190, opened 2020-02-25
sys-kernel/bootengine: Re-enable persistent ifnames

pothos:
In the initramfs persistent ifnames were disabled.
This caused problems because when the renaming was
done after the initramfs, a race made it fail, as
originally reported in
https://github.com/coreos/bugs/issues/1767

Reverts the booteninge commit
"systemd: add module to disable network device renaming"
and widens the networkd match rule for DigitalOcean.

Note: Should be cherry-picked for alpha and edge.


pothos:
Done

PR #193, opened 2020-02-27
app-emulation: bump docker to 19.03.6 to use runc 1.0-rc10 for edge

dongsupark:
So far it looked like docker-runc's version was 1.0-rc10, which was what `runc --version` showed. However it was not the case, because docker 19.03.5 [still pulls in runc](https://github.com/docker/docker-ce/blob/v19.03.5/components/engine/hack/dockerfile/install/runc.installer#L7). As a result, actually it installed runc 1.0-rc8 instead of rc10. That's why the runc still failed to run with a tight memory requirements, like max 20M.

As the upstream docker 19.03.6 already updated the runc version to [1.0-rc10](https://github.com/docker/docker-ce/blob/v19.03.6/components/engine/hack/dockerfile/install/runc.installer#L7), we just need to update docker to 19.03.6.


dongsupark:
See https://github.com/flatcar-linux/coreos-overlay/pull/228 instead.

PR #201, opened 2020-03-05
sys-apps/systemd: Set interface naming scheme to latest

pothos:
When the initramfs gave a persistent name to
a network interface, renaming it via Name is
not working with the v238 naming scheme even
if NamePolicy is unset.
Switch to the newest network interface naming
scheme to resolve this issue.

https://github.com/flatcar-linux/Flatcar/issues/36

Note: Pick for alpha and edge.


pothos:
Ok, good to know that there are some name changes possible which we need to call out in the release notes.

PR #204, opened 2020-03-06
.github: schedule daily Github actions for Go

dongsupark:
Schedule daily Github actions for creating PRs for upstream Go releases, just like Kernel.

Replaces https://github.com/flatcar-linux/coreos-overlay/pull/186


dongsupark:
Updated, by running a shallow git clone instead of a full clone.

PR #205, opened 2020-03-06
.github: fix a commit message and exit code in Kernel workflows

dongsupark:
In the commit message, we should use a prefix `sys-kernel`, without `coreos-sources`.

Also make the entire Github actions not fail if the current release is already the latest Kernel.


dongsupark:
Fixed another issue about failures if the current release is already the latest Kernel.

PR #216, opened 2020-03-13
.github: schedule daily Github actions for Rust

dongsupark:
Schedule daily Github actions for creating PRs for upstream Rust releases.

The Github workflow will create pull request for `dev-lang/rust` in `coreos-overlay`. At the same time, it will send a repository dispatch event to `flatcar-linux/portage-stable`, to update also `virtual/cargo`. We need to send different event types to distinguish alpha from edge.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/28, https://github.com/flatcar-linux/portage-stable/pull/44, https://github.com/flatcar-linux/portage-stable/pull/45, and https://github.com/flatcar-linux/coreos-overlay/pull/221.

### DIFF
--- a/.github/workflows/rust-apply-patch.sh
+++ b/.github/workflows/rust-apply-patch.sh
@@ -2,13 +2,11 @@
 
 set -euo pipefail
 
-branch="rust-${VERSION_NEW}"
+. .github/workflows/common.sh
 
-git -C ~/flatcar-sdk/src/scripts checkout -B "${BASE_BRANCH}" "github/${BASE_BRANCH}"
-git -C ~/flatcar-sdk/src/third_party/portage-stable checkout -B "${BASE_BRANCH}" "github/${BASE_BRANCH}"
+checkout_branches "rust-${VERSION_NEW}"
 
-pushd ~/flatcar-sdk/src/third_party/coreos-overlay >/dev/null || exit
-git checkout -B "${branch}" "github/${BASE_BRANCH}"
+pushd "${SDK_OUTER_SRCDIR}/third_party/coreos-overlay" >/dev/null || exit
 
 updateNeeded=1
 VERSION_OLD=$(sed -n "s/^DIST rustc-\(1.[0-9]*.[0-9]*\).*/\1/p" dev-lang/rust/Manifest | sort -ruV | head -n1)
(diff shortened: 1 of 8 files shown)


dongsupark:
Now this PR depends on https://github.com/flatcar-linux/coreos-overlay/pull/221, as the symlink setup is done by `setup-flatcar-sdk.sh`.

dongsupark:
It turns out, updating Rust to the latest version like 1.42.0 does not always work.
We need to fix a couple of remaining issues, in the current rust toolchains in portage-stable etc, before turning on the automation.
Please do not merge this PR for now.

dongsupark:
Updated the PR. Ready for review.
Now the PR depends on the following PRs as well:
https://github.com/flatcar-linux/portage-stable/pull/44
https://github.com/flatcar-linux/portage-stable/pull/45

.github/workflows/rust-apply-patch.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
marga-kinvolk, line 1: This file is very similar to `docker-apply-patch.sh`, and I assume there will be more `$package-apply-patch.sh` coming... It seems it should be possible to just have a base script that receives a couple of parameters for each package, no?

dongsupark: All right.
I will then try to create `common-apply-patch.sh` that includes common parts.
And make other scripts include the common script.

dongsupark: Done.
Split out common parts to `common.sh`.

.github/workflows/rust-apply-patch.sh
```diff
@@ -0,0 +1,26 @@
...
+popd >/dev/null || exit
+
+popd >/dev/null || exit
+
+generate_patches dev-lang rust Rust
+
+apply_patches
+
+echo ::set-output name=VERSION_OLD::"${VERSION_OLD}"
+echo ::set-output name=UPDATE_NEEDED::"${updateNeeded}"
```
marga-kinvolk, line 26: The other scripts don't have this last line. I see that only the ymls for rust are using this variable for conditionally creating the patches... Is there a reason why the others don't do this?

.github/workflows/rust-apply-patch.sh
marga-kinvolk: Thanks, this looks much better!

.github/workflows/rust-apply-patch.sh
dongsupark: Right. Only rust has `UPDATE_NEEDED`.
There is no specific reason for that. I have only realized we should check if the update is actually needed, so added it only to Rust.

Other actions do not have it. But it has not been an issue, because so far `peter-evans/create-pull-request` simply succeeds, even if no update is needed.

However, it is different in case of Rust, which triggers both `create-pull-request` and `repository-dispatch`. The former will anyway succeed. But the latter will create an unnecessary repository dispatch event, which will create another round of actions on the side of portage-stable. That could really look confusing.

PR #229, opened 2020-03-19
coreos-base/oem-packet: Fix TTY for ARM64

pothos:
The default for arm64 is ttyAMA0 but the
c2.large.arm needs ttyS1.

Note: Pick for alpha and edge.


pothos:
Ah, it's actually works but some temporary other problem caused the output to be broken.

PR #237, opened 2020-03-24
profiles: build tracepath in iputils

dongsupark:
Enable a USE flag `tracepath` for iputils to get the `/usr/sbin/tracepath` binary file included in Flatcar images.

Note, the `tracepath` flag is not the same as the existing `traceroute` flag, which enables only `tracerout6`.


pothos:
Does this include the `traceroute` binary which was asked for?

dongsupark:
No, this PR does not include `traceroute`.
However, I assume most users can achieve purpose by running `tracepath` instead of `traceroute`.
`tracepath` is even better, because it does not require root privileges like `traceroute` does.

PR #242, opened 2020-03-24
sys-kernel: Revert to version used in stable

pothos:
Brings to kernel to the version in `origin/build-2345`.


pothos:
Thanks, done!

PR #247, opened 2020-03-30
sys-kernel: Upgrade Kernel from 5.5 to 5.6

dongsupark:
Upgrade Kernel from 5.5.13 to 5.6.0.

Enable `CONFIG_BRIDGE_NETFILTER` to fix an issue of not being able to turn on `CONFIG_NETFILTER_XT_MATCH_PHYSDEV`.

Now that vanilla Kernel 5.6 includes wireguard, we can simply remove the 3rd-party wireguard patch.

Also enable `CONFIG_WIREGUARD` to get the upstream wireguard included.

Add `-Wno-error=implicit-fallthrough` to the default CFLAGS, to avoid errors like `this statement may fall through`.

Fixes https://github.com/flatcar-linux/Flatcar/issues/64


dongsupark:
Tested for both amd64 and arm64.
It works well.

PR #275, opened 2020-04-07
sys-kernel: bump kernel to 5.4 for alpha

dongsupark:
Change the default Kernel version of Alpha to 5.4, the most latest LTS Kernel tree.

Also update patches and kernel configs, so it could build with the new Kernel.
Update coreos-firmware to 20191022, to be compatible with Kernel 5.4.

See also https://github.com/flatcar-linux/Flatcar/issues/73.


dongsupark:
Updated to Kernel 5.4.32.

PR #276, opened 2020-04-07
Adapt the sed command to the new file contents

marga-kinvolk:
With the latest systemd version, the lines that sed is removing now
include an exclamation mark, so adapt the sed command to accept both
formats.

These lines need to be removed so that the configuration set in
/usr/lib/tmpfiles.d/pam.conf file can be used (duplicate entries will
lead to systemd not doing anything).

Fixes: flatcar-linux/Flatcar#75

### DIFF
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -422,8 +422,8 @@ multilib_src_install_all() {
 	# Do not ship distro-specific files (nsswitch.conf pam.d)
 	rm -rf "${ED}"/usr/share/factory
 	sed -i "${ED}"/usr/lib/tmpfiles.d/etc.conf \
-		-e '/^C \/etc\/nsswitch\.conf/d' \
-		-e '/^C \/etc\/pam\.d/d'
+		-e '/^C!* \/etc\/nsswitch\.conf/d' \
+		-e '/^C!* \/etc\/pam\.d/d'
 }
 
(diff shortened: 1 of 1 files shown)


pothos:
Can you please prefix the first commit message line with the package name you edit (`sys-apps/systemd: `)?
It would be good to state in the second commit that https://github.com/flatcar-linux/init/blob/flatcar-master/configs/tmpfiles.d/issuegen.conf defines `/etc/issue`.

pothos:
And `pam.conf` seems to come from https://github.com/flatcar-linux/coreos-overlay/blob/flatcar-master/sys-libs/pam/pam-1.2.1-r1.ebuild#L139 – would be good to state that in the first commit message.

marga-kinvolk:
PTAL. I've reworded both commits as requested.

pothos:
Once merged, both commits need to be picked for `flatcar-master-alpha`, `flatcar-master-beta`, and `flatcar-master-edge`.

pothos:
Ah, and cherry-picked for `flatcar-build-2466.0.0` and `flatcar-build-2466.99.0` to go into the currently ongoing Alpha and Edge release.

marga-kinvolk:
I've now cherry picked this change to:
- flatcar-master-edge
- flatcar-master-alpha
- flatcar-master-beta
- flatcar-build-2466.99.0
- flatcar-build-2466.0.0

PR #279, opened 2020-04-08
app-admin/flannel-wrapper: set MACAddressPolicy=none for flannel

marga-kinvolk:
When setting up flannel interfaces, use MACAddressPolicy=none, so that
the MAC Address used is the one set by flannel and not the one assigned
by systemd.

See https://github.com/coreos/flannel/issues/1155 for more information.

This fixes the issue we saw with the vxlan test

### DIFF
--- a/app-admin/flannel-wrapper/files/50-flannel.network
+++ b/app-admin/flannel-wrapper/files/50-flannel.network
@@ -3,3 +3,4 @@ Name=flannel*
 
 [Link]
 Unmanaged=yes
+MACAddressPolicy=none


pothos:
Thanks, small correction for the commit message: `is the one set by flannel` → `is the initial one set by the kernel`

marga-kinvolk:
Done

marga-kinvolk:
This has now been cherry-picked to:
- flatcar-build-2466.99.0
- flatcar-build-2466.0.0
- flatcar-master-edge
- flatcar-master-alpha
- flatcar-master-beta

PR #282, opened 2020-04-09
app-admin/flannel-wrapper: Create a 50-flannel.link

marga-kinvolk:
When setting up flannel interfaces, use MACAddressPolicy=none, so that
the MAC Address used is the initial one set by the kernel and not the
one assigned by systemd.

See coreos/flannel#1155 for more information.

In #279 we tried adding the MACAddressPolicy=none setting to the
existing 50-flannel.network file. But the change should have been in a
.link file, not a .network file.

# Testing done

I manually built the edge image, deployed it in packet and verified that ping on the flannel interface worked.

### DIFF
--- a/app-admin/flannel-wrapper/files/50-flannel.link
+++ b/app-admin/flannel-wrapper/files/50-flannel.link
@@ -1,0 +1,5 @@
+[Match]
+OriginalName=flannel*
+
+[Link]
+MACAddressPolicy=none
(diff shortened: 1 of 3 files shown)


marga-kinvolk:
This has now been cherry-picked to:
- flatcar-build-2466.0.0
- flatcar-build-2466.99.0
- flatcar-master-alpha
- flatcar-master-beta
- flatcar-master-edge

PR #320, opened 2020-04-27
Set oniguruma USE flag for the jq package

samm-git:
See https://github.com/flatcar-linux/Flatcar/issues/101

This PR adds ONIGURUMA regex libary to support regular expressions in jq tool

# How to use

See https://github.com/flatcar-linux/Flatcar/issues/101 for the testcase

# Testing done

Tested with local SDK setup. After rebuild i see that it is linked with libonig library:
```
builder@localhost ~/trunk/src/scripts $ ldd /mnt/host/source/chroot/build/amd64-usr/usr/bin/jq
	linux-vdso.so.1 (0x00007ffe54d57000)
	libjq.so.1 => /usr/lib64/libjq.so.1 (0x00007efc495b8000)
	libc.so.6 => /lib64/libc.so.6 (0x00007efc493e0000)
	libm.so.6 => /lib64/libm.so.6 (0x00007efc49296000)
	libonig.so.5 => /usr/lib64/libonig.so.5 (0x00007efc49205000)
	/lib64/ld-linux-x86-64.so.2 (0x00007efc49614000)
```
Also test from my testcase is now passing:
```
builder@localhost ~/trunk/src/scripts $ echo '["xabcd", "ABC"]'| /mnt/host/source/chroot/build/amd64-usr/usr/bin/jq '.[] | test("a b c # spaces are ignored"; "ix")'
true
true
```

### DIFF
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -100,3 +100,6 @@ sys-apps/man-db -nls
 
 # Disable zstd to avoid adding it to prod images until something needs it
 sys-fs/btrfs-progs -zstd
+
+# enable regular expression processing in jq
+app-misc/jq oniguruma


dongsupark:
Thanks for creating a pull request! :+1: 

However, this PR alone is not enough for enabling the oniguruma library for jq.

1. You need to enable a specific USE flag `oniguruma` for `app-misc/jq`. The flag is turned off by default. Without the flag, jq builds will simply ignore oniguruma. So you should probably need to add additional lines to coreos-overlay's [profiles](https://github.com/flatcar-linux/coreos-overlay/tree/flatcar-master/profiles) directory.

2. So far the oniguruma library was never included in CoreOS/Flatcar, neither in coreos-overlay nor in [portage-stable](https://github.com/flatcar-linux/portage-stable). It means, you probably need to also pull in the new dependencies including oniguruma into portage-stable. You can simply copy the ebuilds from upstream Gentoo. Or you can also make use of an existing helper script [update_ebuilds](https://github.com/flatcar-linux/scripts/blob/flatcar-master/update_ebuilds).

3. Your PR is based on `flatcar-master`, which means it is meant for all channels, stable to edge. If it is about critical bug fixes, security fixes, or urgent needs by users, then we should get it included in stable. However, I am not sure if oniguruma in jq is such a case. So I would suggest adding the tool first to Alpha and Edge. It means the PR needs to be based on `flatcar-master-alpha` or `flatcar-master-edge`.

samm-git:
@dongsupark thank you for reply and for the hints, never been adding packages before. Do you know is there any kind of jenkins or other ci scenario to test the build myself?

dongsupark:
No, at the moment the only CI is an internal Jenkins for us Kinvolk, which covers the whole build chain.
No CI is integrated into individual pull requests.
In the future we could enable Github Actions for building individual packages.

So currently you should set up a local SDK, following [the SDK guide](https://docs.flatcar-linux.org/os/sdk-modifying-coreos/). Then under the SDK environment, you can test individual builds on your own.

samm-git:
Hi @dongsupark. I think it should be fine now. I removed direct dependency on dev-libs/oniguruma  in the coreos package, as it is already done by USE flag in jq. Tested with local SDK setup, as you recommended and rebased against `flatcar-master-alpha`. Please let me know if anything else needed and thank you for suggestions.

dongsupark:
Thanks a lot for updating the PR.
In general it looks good.

However, there is one thing about our workflow.
`coreos-overlay` is for packages that need customization for Flatcar, while [portage-stable](https://github.com/flatcar-linux/portage-stable) is for packages that came directly from upstream Gentoo without customization. So in this case, `dev-libs/oniguruma` should actually be located under `portage-stable`, as it looks the same as upstream Gentoo.

Sorry, I admit, the workflow might not be straightforwards for newcomers.

samm-git:
@dongsupark thank you for review :) I think some document for the newcomers would help a lot, anyway - changes are now done. I moved all ONIGURUMA code to the PR for the portage-stable PR - https://github.com/flatcar-linux/portage-stable/pull/64. Here i left only USE flag.

PR #346, opened 2020-05-07
net-vpn/wireguard-tools: rename wireguard to wireguard-tools for edge

dongsupark:
Since `wireguard` in upstream Gentoo was renamed to `wireguard-tools`, we
need to rename the ebuild.
Files were simply copied from `flatcar-master-alpha` branch.


sayanchowdhury:
I think we would also need to add `net-vpn/wireguard-modules` here.

dongsupark:
No, we do not need to add `net-vpn/wireguard-modules`, which is DKMS for Kernel 5.5 or older.
Edge has Kernel 5.6, and its vanilla Kernel already has a wireguard kernel module.
See also https://github.com/gentoo/gentoo/blob/9cad245879c2e9c311452bcc81a72baf49fb933f/net-vpn/wireguard-modules/wireguard-modules-1.0.20200506.ebuild#L36..L42 .

For Alpha with Kernel 5.4, we could add wireguard-modules. However, at the moment we already have a third-party patch for Wireguard in `sys-kernel/coreos-sources`, so we do not need another DKMS for the Kernel module.

PR #362, opened 2020-05-14
sys-kernel: Bump the kernel to 5.4 for Beta

sayanchowdhury:
# sys-kernel: Bump the kernel to 5.4 for Beta

This PR bumps the kernel to 5.4 in Beta and also updates the firmare to 20191022

# Testing done
```bash
core@localhost ~ $ uname -a
Linux localhost 5.4.38-flatcar #1 SMP Mon May 11 11:43:39 -00 2020 x86_64 AMD EP
YC 7401P 24-Core Processor AuthenticAMD GNU/Linux
```

PR #363, opened 2020-05-14
.github: update Kernel version to 5.4 for beta

dongsupark:
Now that we updated Kernel to 5.4 for Beta, we also need to set Kernel
version to 5.4 in Github Actions.

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/362.


dongsupark:
@sayanchowdhury Now we need to merge this PR as well.

PR #365, opened 2020-05-16
Add Amazon SSM manager to the EC2 images

samm-git:
# Add AWS SSM manager

AWS Systems Manager Agent (SSM Agent) is Amazon software that can be installed and configured on an EC2 instance, an on-premises server, or a virtual machine (VM). SSM Agent makes it possible for Systems Manager to update, manage, and configure these resources. The agent processes requests from the Systems Manager service in the AWS Cloud, and then runs them as specified in the request. 

SSM Agent is preinstalled, by default, on the following Amazon Machine Images (AMIs):

- Amazon Linux
- Amazon Linux 2
- Ubuntu Server 16.04 
- Ubuntu Server 18.04 Amazon ECS-Optimized 

Also for the RPM and DEB based systems Amazon provides binary packages. Flatcar Linux does not have support for it [yet], so it is required to use 3rd party script for the integration via `userdata`

See https://github.com/flatcar-linux/Flatcar/issues/107 for the request

# How to use

## Target state (not here yet)

- [x] Package for the `amazon-ssm-agent` added to the `coreos-overlay`.
- [x] Dependency added to the AWS EC2 package
- [x] Integration with ignition/systemd
- [ ] Testing EC2 AMI with the agent built and enabled

# Testing done

- [x] Ebuild tested with a recent Flatcar SDK
- [ ] Test package on the live coreos installation
- [ ] Build EC2 AMI and test on it


samm-git:
@dongsupark would be great if you can suggest how to integrate it with systemd. My understanding was to create systemd file from ignition, as i cant (?) put one from the oem package out of the oem dir.

pothos:
Thanks for your work on it. I'm not sure about the build commands… and if we want to include this tool in the current way of vendor tool distribution.
But to answer your question, the service would need to be set up in `coreos-base/oem-ec2-compat/files/base/base-ec2.ign`.

samm-git:
@pothos thank you for feedback. Build commands are +- replicating what makefile is doing. I will work on ignition config, thank you for the hint.

samm-git:
@pothos @dongsupark i think patch is ready to land - i added ignition/systemd support and did some minor cleanup/comments. Would be nice to see it merged, happy to answer any questions.

app-emulation/amazon-ssm-agent/amazon-ssm-agent-2.3.1205.0.ebuild
```diff
@@ -0,0 +1,61 @@
...
+  ln -s ${PWD}/vendor/src/*  ${PWD}/vendor/
+}
+
+src_compile() {
+  # this is replication command from the vendor makefile
+  # but without network activity during build phase
+  local GO_LDFLAGS="-s -w -extldflags=-Wl,-z,now,-z,relro,-z,defs"
+  export GOPATH="${WORKDIR}/${PN}-${PV}"
+  # set release version
+  BRAZIL_PACKAGE_VERSION=${PV} go run ./agent/version/versiongenerator/version-gen.go
```
dongsupark, line 36: Hmm, what is `BRAZIL_PACKAGE_VERSION`?
I don't see it in the code of amazon-ssm-agent.

app-emulation/amazon-ssm-agent/amazon-ssm-agent-2.3.1205.0.ebuild
```diff
@@ -0,0 +1,61 @@
...
+SLOT="0"
+KEYWORDS="amd64 arm64"
+
+S="${WORKDIR}/${PN}-${PV}/src/${EGO_PN}"
+
+PATCHES=(
+  # patch generated by `goimports -w ./agent` command
+  "${FILESDIR}"/${P}-goimports.patch
+)
+
```
dongsupark, line 24: For consistency with other ebuilds, can you please use TAB for indentation, instead of 2 spaces?

app-emulation/amazon-ssm-agent/amazon-ssm-agent-2.3.1205.0.ebuild
samm-git: @dongsupark this is the most crazy and funny thing i ever seen in this code (which, btw, is very bad quality). Variable is used by https://github.com/aws/amazon-ssm-agent/blob/master/agent/version/versiongenerator/version-gen.go#L60. This way they are setting version number in the releases, as [version file](https://github.com/aws/amazon-ssm-agent/blob/master/VERSION) is always showing `2.3.0.0`. I found that when i was analyzing how releases are published.

app-emulation/amazon-ssm-agent/amazon-ssm-agent-2.3.1205.0.ebuild
samm-git: @dongsupark done. Also removed updated to fit on OEM partition

samm-git:
@dongsupark now it fits to 128Mb, tested with `./image_to_vm.sh --format ami`

app-emulation/amazon-ssm-agent/amazon-ssm-agent-2.3.1205.0.ebuild
dongsupark: OMG, it is indeed funny.

Why does it translate a base64-encoded string to a name for a specific env variable? It looks like a deliberately cryptic code. How could any newcomer guess the prefix `BRAZIL_`? :joy:  I have tried to find out its reason, but could find nothing from https://github.com/aws/amazon-ssm-agent/commit/57316ab84030b004eaecd19296806020a2f6a502, and am not able to log in their internal issue tracker https://cr.amazon.com/r/5572760/. Of course no docs at all about that.

Anyway thanks for the explanation.

samm-git:
@dongsupark thank you for review. squash is done, let me know if anything else needed.

samm-git:
@dongsupark one thing in my mind - by design this tool creates new uid called ssm-user and grants sudo to it. But we already do have `core` user for +- the same. I think that may be we should replace default user from `ssm-user` to the `core` to make it more aligned. What do you think about it?

samm-git:
@dongsupark i updated agent to the latest (2.3.1319.0) version and rebased it. Would be great to see if merged eventually

sayanchowdhury:
Thanks @samm-git

dongsupark:
@sayanchowdhury You would probably want to cherry-pick it to edge.

samm-git:
Yeah, thank you!

PR #367, opened 2020-05-18
Update OpenSSL to 1.0.2u for GLSA 202004-10

pothos:
Port changes from upstream https://github.com/coreos/coreos-overlay/pull/3947

Note: Pick for all channels and `flatcar-build-2513.0.0`.


pothos:
Done, had to revert some Edge commits that updated it as well but to an earlier version.

PR #382, opened 2020-05-26
app-crypt/gnupg: add patches for accepting without UIDs for edge

dongsupark:
When the GnuPG keyserver is set to `keys.openpgp.org`, `gpg --recv-keys` occasionally fails with the following error:

```
gpg: key E52F0DB391453C45: no user ID
```

We need to make GnuPG accept keys even without UIDs.
Original patches come from https://salsa.debian.org/debian/gnupg2/tree/f292beac1171c6c77faf41d1f88c2e0942ed4437/debian/patches/import-merge-without-userid .
See also https://dev.gnupg.org/T4393 and https://github.com/flatcar-linux/Flatcar/issues/117.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/74.


app-crypt/gnupg/gnupg-2.2.20.ebuild
```diff
@@ -0,0 +1,167 @@
...
+S="${WORKDIR}/${MY_P}"
+
+DOCS=(
+	ChangeLog NEWS README THANKS TODO VERSION
+	doc/FAQ doc/DETAILS doc/HACKING doc/TRANSLATE doc/OpenPGP doc/KEYSERVER
+)
+
+PATCHES=(
+	"${FILESDIR}/${PN}-2.1.20-gpgscm-Use-shorter-socket-path-lengts-to-improve-tes.patch"
+	"${FILESDIR}/${PN}-allow-import-of-previously-known-keys-even-without-UI.patch"
```
sayanchowdhury, line 54: Can we add the comment that these two patches are flatcar specific?

dongsupark:
Done.
Imported the upstream changes, and added Flatcar comments.

PR #400, opened 2020-06-04
sys-apps/systemd: Manually enable the getty service

sayanchowdhury:

# sys-apps/systemd: Enable the getty if not already enabled

With systemd 243, few of the services from whom the symlinks were created automatically are no longer created and a `systemctl preset-all` is recommended to run after the installation

From what I saw in the [Gentoo ebuild](https://github.com/flatcar-linux/coreos-overlay/blob/flatcar-master-alpha/sys-apps/systemd/systemd-9999.ebuild#L499-L500) - In case of installation, if the package is [replacing an outdated package](https://github.com/flatcar-linux/coreos-overlay/blob/flatcar-master-alpha/sys-apps/systemd/systemd-9999.ebuild#L497), it then enables the `getty` and `remote-fs`and then recommends running `preset-all`

On the other hand, I checked Fedora where it runs `systemctl preset-all` upon installation.

For fixing the issue flatcar-linux/Flatcar#131, rather than running `systemctl preset-all`, just enable the `getty` service.

# How to use

```bash
./build_packages
```

UPDATE: The mentioned fix does not create the appropriate location. Even though the SDK seems to contain the symlinks, the final image does not. Even tried with running `systemd preset-all` but that does not do anything. 

After doing a round of debugging realized, realized that disabling the `split-usr` use flag also fixes the issue, as with `split-usr` enabled it creates the files in `/lib` or else it creates them in `/usr/lib`. 

I believe the reason `systemd preset-all` is not working is that we are [removing the preset file that contains all the actions](https://github.com/flatcar-linux/coreos-overlay/blob/flatcar-master-alpha/sys-apps/systemd/systemd-9999.ebuild#L389). But that's just theory.

So, IMO we can merge this PR and create a seperate issue to track the non-working of `systemd preset-all`.

# Testing done

```
./image_to_vm --format=iso
```

### DIFF
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -385,6 +385,10 @@ multilib_src_install_all() {
 	systemd_enable_service multi-user.target systemd-resolved.service
 	systemd_enable_service sysinit.target systemd-timesyncd.service
 
+	# enable getty manually
+	mkdir --parents "${ED}/usr/lib/systemd/system/getty.target.wants"
+	dosym ../getty@.service "/usr/lib/systemd/system/getty.target.wants/getty@tty1.service"
+
 	# Do not enable random services if /etc was detected as empty!!!
 	rm "${ED}$(usex split-usr '' /usr)/lib/systemd/system-preset/90-systemd.preset" || die
 	insinto $(usex split-usr '' /usr)/lib/systemd/system-preset


sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -492,9 +492,15 @@ pkg_postinst() {
...
 
+	if type systemctl &>/dev/null; then
+		if [ ! systemctl --quiet --root="${ROOT:-/}" is-enabled getty@.service ]; then
+			systemctl --root="${ROOT:-/}" enable getty@.service || FAIL=1
+		fi
+	fi
+
 	if [[ -z ${REPLACING_VERSIONS} ]]; then
 		if type systemctl &>/dev/null; then
-			systemctl --root="${ROOT:-/}" enable getty@.service remote-fs.target || FAIL=1
```
pothos, line 497: It looks like `remote-fs.target` should be moved out of `REPLACING_VERSIONS`, too?

A general note for testing: We should also check if the service is still enabled when the rootfs is cleaned and the instance rebooted or when a new rootfs is created with Ignition. This is the case because the service symlinks end up under `/usr`, or?

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -385,6 +385,10 @@ multilib_src_install_all() {
 	systemd_enable_service multi-user.target systemd-resolved.service
 	systemd_enable_service sysinit.target systemd-timesyncd.service
 
+	# enable getty manually
+	mkdir --parents "${D}/usr/lib/systemd/system/getty.target.wants"
```
pothos, line 389: In the lines below and above `ED` is used instead of `D`; I would stick to it to include the `EPREFIX` in case this is needed.

sayanchowdhury:
> Can you modify the PR description to explain _why_ you're enabling getty instead of running preset-all? The description just says that you chose this, but there's no explanation as to why.

Agreed. I will update the description and title with more metadata.

pothos:
Restoring the preset action can only be done if it doesn't impact the behavior when `/etc` doesn't exist at runtime because the rootfs was created by Ignition. I don't know if we could use it to generate the symlinks and then move them to `/usr` in the ebuild. I think it's ok to keep the current way of explicitly enabling services with symlinks in `/usr`.

pothos:
Should be picked for Edge, or?

PR #411, opened 2020-06-09
profiles: enable bind-tools again for ARM for alpha

dongsupark:
`net-dns/bind-tools` has been disabled since a long time, probably because of build errors around cross-compilation for ARM. However, bind-tools binaries should be at least included in ARM images.
So enable `bind-tools` again for ARM without `gssapi` included.

To do that, disable gssapi for bind-tools only in the ARM profile, and enable gssapi only in the AMD profile.

Now that bind-tools are built with gssapi only for AMD, without gssapi for ARM, we need to get the USE flag for `sys-auth/sssd` requirement relaxed. Profile for each architecture will instead choose whether to use gssapi.

## How to use

```
emerge-arm64-usr net-dns/bind-tools sys-auth/sssd
```

### DIFF
--- a/profiles/coreos/arm64/package.use
+++ b/profiles/coreos/arm64/package.use
@@ -1,5 +1,8 @@
 # arm64 use
 
+# Disable gssapi for arm64 to avoid build errors
+net-dns/bind-tools -gssapi
+
 # FIXME: why isn't this set by default???
 sys-libs/ncurses unicode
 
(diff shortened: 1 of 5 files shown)


profiles/coreos/arm64/package.use
```diff
@@ -1,5 +1,7 @@
 # arm64 use
 
+net-dns/bind-tools -gssapi
```
marga-kinvolk, line 3: Maybe add a comment here pointing to the issue or at least a comment explaining that enabling gssapi causes build failures?

dongsupark: Added comments as suggested

PR #412, opened 2020-06-09
profiles: enable bind-tools again for ARM for edge

dongsupark:
`net-dns/bind-tools` has been disabled since a long time, probably because of build errors around cross-compilation for ARM. However, bind-tools binaries should be at least included in ARM images.
So enable `bind-tools` again for ARM without `gssapi` included.

To do that, disable gssapi for bind-tools only in the ARM profile, and enable gssapi only in the AMD profile.

Now that bind-tools are built with gssapi only for AMD, without gssapi for ARM, we need to get the USE flag for `sys-auth/sssd` requirement relaxed. Profile for each architecture will instead choose whether to use gssapi.

## How to use

```
emerge-arm64-usr net-dns/bind-tools sys-auth/sssd
```

### DIFF
--- a/profiles/coreos/amd64/generic/package.use
+++ b/profiles/coreos/amd64/generic/package.use
@@ -16,3 +16,6 @@ app-arch/tar                selinux
 # Only ship microcode currently distributed by Intel
 # See https://bugs.gentoo.org/654638#c11 by iucode-tool maintainer
 sys-firmware/intel-microcode vanilla
+
+# Enable gssapi only for amd64, to avoid build errors in arm64.
+net-dns/bind-tools           gssapi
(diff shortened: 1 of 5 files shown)


dongsupark:
Added comments as suggested.

> I'm a bit confused as to why these are two separate PRs instead of one PR and one cherry pick. Doesn't that make it harder to keep track of changes?

It is just my habit.
I do so because I am paranoid enough to not trust my capability of cherry-picking a chunk of code to branches for other channels. I have made so many embarassing failures when cherry-picking across channels. It would be no big issue with small PRs like this one. However, it is not that straightforwards to do so for a huge PR.
On the other hand, I agree with what you said about difficulties of tracking changes. I will try to create a single PR if possible.

PR #431, opened 2020-06-19
sys-libs/glibc: update glibc to 2.30-r8 for alpha

dongsupark:
Update glibc to 2.30-r8 to sync with upstream Gentoo, to address security issues reported as [GLSA 202006-04](https://security.gentoo.org/glsa/202006-04).

Apply Flatcar changes, by adapting the original coreos patch [eb07324f4de3](https://github.com/flatcar-linux/coreos-overlay/commit/eb07324f4de3) (`sys-libs/glibc: Apply CoreOS changes`) into glibc 2.30.

## How to use

```
emerge-amd64-usr sys-libs/glibc
./build_packages
```


dongsupark:
cherry-picked to edge

PR #433, opened 2020-06-22
sys-kernel/coreos-sources: Update kernel from 5.6 to 5.7

sayanchowdhury:
# sys-kernel/coreos-sources: Update kernel from 5.6 to 5.7

This PR updates the kernel from 5.6 to 5.7 and also applies the fix for the lockup patch

# How to use

```bash
./build_packages
./build_image
```

# Testing done
Not yet. Will update the PR with testing updates.


dongsupark:
The bug fix for the crypto issue is [included](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-5.7.y&id=f6f785bdf6043f9a37fad5ffeaf5bb53e5b8afe2) in Kernel v5.7.6.
If we upgrade this PR to v5.7.6, we can get rid of the third-party patch.

sayanchowdhury:
Sure. I'll update the PR.

PR #444, opened 2020-06-25
profiles: force to use ssl USE flag for wget

dongsupark:
When catalyst tries to fetch a file via https, wget sometimes fails to do so, with the following messages:

```
https://www.kernel.org/pub/software/scm/git/git-2.24.1.tar.xz: HTTPS
support not compiled in.
!!! Couldn't download 'git-2.24.1.tar.xz'. Aborting.
```

That probably happens because wget in some catalyst stages are compiled without `ssl` USE flag.
If a catalyst stage is lucky enough to rebuild wget with `ssl` before actually fetching a file, it would work well.
Though if not, it would fail.
It is not deterministic, and hard to reproduce.

So backport the [fix](https://github.com/gentoo/gentoo/commit/d141380b915d) from upstream Gentoo,  for both amd64 and arm64.
By setting `ssl` for wget in `package.use.force`, it is now not possible to disable `ssl` for wget.

More details: https://bugs.gentoo.org/611072

### DIFF
--- a/profiles/coreos/arm64/package.use.force
+++ b/profiles/coreos/arm64/package.use.force
@@ -1,2 +1,7 @@
 sys-auth/polkit -introspection
 sys-apps/systemd -introspection
+
+# Matt Turner <mattst88@gentoo.org> (2020-03-28)
+# wget is the default FETCHCOMMAND, and most distfiles are distributed via
+# HTTPS. Bug #611072
+net-misc/wget ssl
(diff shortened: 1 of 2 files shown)


pothos:
Will it be enabled in stage1 and does it mean that we can remove https://github.com/flatcar-linux/scripts/blob/flatcar-master/bootstrap_sdk#L57?

dongsupark:
Yeah, I think so, we can probably remove the hack in `bootstrap_sdk`.

Anyway I would prefer to change those gradually.
First enable `ssl` for wget.
Then go through several Jenkins runs with that change.
If there is no wget error like @sayanchowdhury has seen, and there is no regression, then remove the hack in `bootstrap_sdk`.

dongsupark:
cherry-picked to beta, alpha, and edge.

PR #448, opened 2020-06-29
sys-apps/systemd: Update systemd to v245

sayanchowdhury:
Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

# sys-apps/systemd: Update systemd to v245

Should be merged along with

- https://github.com/flatcar-linux/init/pull/23
- https://github.com/flatcar-linux/bootengine/pull/16
- https://github.com/flatcar-linux/portage-stable/pull/86

# How to use

```
emerge-amd64-usr sys-apps/systemd
```

# Testing done

This is a draft PR to get more :eyes:  on the PR early.

Update: Perform a local/dev build of the packages and image.


sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -10,7 +10,7 @@ if [[ ${PV} == 9999 ]]; then
 	# Use ~arch instead of empty keywords for compatibility with cros-workon
 	KEYWORDS="~amd64 ~arm64 ~arm ~x86"
 else
-	CROS_WORKON_COMMIT="171ebfdbcb79b1f42659d111b5a642e72ea02021" # v243-flatcar
+	CROS_WORKON_COMMIT="171ebfdbcb79b1f42659d111b5a642e72ea02021" # v245-flatcar
```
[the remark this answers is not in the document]
sayanchowdhury: I've pushed it now.

sayanchowdhury:
I had a look through it. Thanks for the comments, that would surely make it easier to maintain in the long run. I see you based it on top of `systemd-245.5.ebuild`, but I did it keeping `system-9999` as the base. That should not be a problem though.

Can you also push the changes to this PR so that we can test it.

krnowak:
Pushed, haven't changed the `CROS_WORKON_COMMIT` line, though.

Basing on `systemd-9999.ebuild` shouldn't be a problem, either - currently it only differs in the patches they apply in `src_prepare`. But we drop this patching code altogether, so it doesn't really matter.

sys-kernel/bootengine/bootengine-9999.ebuild
```diff
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ "${PV}" == 9999 ]]; then
```
dongsupark, line 10: How about updating the symlink `bootengine-0.0.38-r6.ebuild` to `bootengine-0.0.38-r7.ebuild`?
The bootengine is notorious for mysterious side-effects caused by garbage cpio files created by previous builds.
So it might be especially a good idea to bump the revision.

sayanchowdhury: Sounds good. Will apply that.

PR #450, opened 2020-06-30
sys-kernel: enable kernel config CONFIG_IKHEADERS

dongsupark:
Enable kernel config [CONFIG_IKHEADERS](https://github.com/torvalds/linux/blob/435faf5c218a47fd6258187f62d9bb1009717896/init/Kconfig#L610-L617
), to make Kernel export kernel headers via `/sys/kernel/kheaders.tar.xz`.
Then bpf-related tools can be used without additional kernel headers in userspace.

As the config is available since Kernel 5.2, we can apply it to Beta, Alpha (both 5.4), and Edge (5.7).

Fixes https://github.com/flatcar-linux/Flatcar/issues/152.

## How to use

```
./build_packages
```


dongsupark:
cherry-picked to alpha and edge.

PR #471, opened 2020-07-14
Update OpenSSL and promote OpenSSH

pothos:
This updates the OpenSSL and OpenSSH packages. Updating OpenSSL requires overwriting a package mask in portage-stable until that is synced to the latest state (currently blocked by glibc).
The update of OpenSSH is basically a promotion from Alpha but not using the Alpha commit and instead follows the CoreOS practice of first creating a sync commit that overwrites all local changes and then introducing the local changes in a second commit.
TODO: Port the split OpenSSH commits to Alpha, and update OpenSSL there.

# How to use

# Testing done


dongsupark:
To update openssl to 1.1.1g, you would probably need to update those:

* `app-crypt/mit-krb5` in portage-stable
* `app-crypt/trousers` in coreos-overlay
* `dev-libs/cyrus-sasl` in coreos-overlay
* (or more?)

Without updating those, openssl build will fail.
`app-crypt/mit-krb5` is the simplest. It will be no issue.
The other two packages though, `trousers` and `cyrus-sasl`, have third-party patches, so it might need some work.

Now I understand why it was not trivial to openssl to 1.1. ;-)

pothos:
Yes^^ I also saw `dev-embedded/u-boot-tools` fail to build

PR #473, opened 2020-07-15
Apply rust changes

pothos:
This collects the downstream crossdev patches we currently need in a single commit and applies it to the upstream Rust ebuild file.

# How to use

Compile Rust in the SDK after setting up an arm64 board

# Testing done

None

Note: Should be adapted for the other channels

### DIFF
--- a/dev-lang/rust/rust-1.44.1.ebuild
+++ b/dev-lang/rust/rust-1.44.1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{6,7} )
+PYTHON_COMPAT=( python3_{6,7,8} )
 
 inherit bash-completion-r1 check-reqs estack flag-o-matic llvm multiprocessing multilib-build python-any-r1 rust-toolchain toolchain-funcs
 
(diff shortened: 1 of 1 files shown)


dev-lang/rust/rust-1.44.1.ebuild
```diff
@@ -1,9 +1,12 @@
+# Difference to upstream ebuild file:
+# - rust-crossdev.patch applied
+#
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{6,7} )
+PYTHON_COMPAT=( python3_{6,7,8} )
```
dongsupark, line 9: It will not work, as our codebase still does not support `python3_8`.

dev-lang/rust/rust-crossdev.patch
```diff
@@ -0,0 +1,52 @@
+diff --git a/dev-lang/rust/rust-1.44.1.ebuild b/dev-lang/rust/rust-1.44.1.ebuild
```
dongsupark, line 1: I am wondering why this patch `rust-crossdev.path` exists.
Maybe you wanted to keep a separate crossdev patch, which can be easily applied in the upcoming releases?

I understand the motivation.
But I am not sure if that is a good idea.
Almost for every rust release we needed to adjust the crossdev patch, simply because of tiny changes from upstream Gentoo ebuilds.
Now with this separate file, we need to update the crossdev patch twice, the first in the actual ebuild, and the second in `rust-crossdev.patch`. Not to mention that the latter has a complicated syntax.

pothos: A commit diff is easily lost, that's why I thought a file would be good. But yes, we don't have it normally and it also has drawbacks, I'll remove it again.

dev-lang/rust/rust-1.44.1.ebuild
pothos: Ok, thanks, I thought this is just a range of possibilities and didn't know it would try to use it.

PR #481, opened 2020-07-20
sys-kernel/coreos-firmware: Change source URL to reliable mirror

pothos:
The caching git web view which was used as source location is not
reliable because the cache can be corrupt, the gzip compression of the
snapshot can change, and the cache is produced by the web server which
is not there to give high security guarantees. We experienced cache
corruption.
Switch to the recommended mirror under
https://mirrors.edge.kernel.org/pub/linux/kernel/firmware/
which also hosts signatures and not having the downsides mentioned
above. This is a downstream change until upstream Gentoo changes the
location.

# How to use

```
emerge-amd64-usr coreos-firmware
```

# Testing done

Builds

Note: Pick for `flatcar-2513`


dongsupark:
I agree with the change itself.

However, with this PR, the build does not pass. `src_unpack` fails like:

```
$ emerge-amd64-usr -1v coreos-firmware
...
mv: cannot move 'linux-firmware-20191022/' to a subdirectory of itself, 'linux-firmware-20191022/linux-firmware-20191022'
```

It was [fixed](https://github.com/gentoo/gentoo/commit/1735a77829830d06694de600daf66c2e1e9e19f3) by upstream Gentoo, but it is not included in coreos-firmware.

So can you please update it like Gentoo ebuilds?

And apparently it is not the only issue in this PR... ;-)

pothos:
> It was fixed by upstream Gentoo, but it is not included in coreos-firmware.
Thanks, the whole `mv` is now obsolete, fixed it by removing.

PR #485, opened 2020-07-21
sys-kernel/coreos-modules: enable virtio modules (mmio and vsock) for…

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

# Enable virtio modules (mmio and socket)

This PR enable in the Flatcar kernel the modules to

- mount Virtio MMIO device when the device is passed in as kernel flag (this is the case for Firecracker)

- enable the module `vmw_vsock_virtio_transport` - this is needed to enable `vsock` (Firecracker uses it)

# How to use

These are needed to be able to run properly Flatcar kernel on Firecracker.

# Testing done

~I still haven't tested this. I would need to create a kernel (and image too?) to test it.~

This has been tested cooking a kernel with these flags on and running Firecracker VMs


mazzy89:
aaaaand we did it 🤠 Thanks a lot @pothos

PR #486, opened 2020-07-21
app-emulation/docker-runc: remove patch cgroups hybrid mode

dongsupark:
Since runc 1.0.0-rc91 or newer has a completely different structure of source code around cgroups hybrid mode for systemd, it is not trivial to rebase the third-party patch on top of 1.0.0-rc91.

Now that other tools like Inspektor Gadget do not rely on the hybrid mode any more, we can hopefully remove the hybrid patch from Flatcar Edge. That will allow us to update runc to 1.0.0-rc91.

## How to use

```
./build_packages
```

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/464.


dongsupark:
Closing, will recreate when necessary.

PR #493, opened 2020-07-23
bootengine: move to version with disk randomization fix

marga-kinvolk:
# Build the latest bootengine

This version of bootengine adds a new script, so on top of moving the CROS commit id pointer, we also need to add execution permissions to the new disk-uuid script.

# How to use / Testing done

Building an image with this change plus the change in flatcar-linux/scripts#82 leads to GRUB booting successfully on a c3.medium.x86 machine followed by the disk UUID getting randomized.

WIP notice: Due to issues with current flatcar-master-alpha, I wasn't yet able to fully test this change on all platforms, I only manually tested it on Packet. Once alpha is fixed, I'll test again on all platforms.

# Coupling

This needs to be submitted after flatcar-linux/bootengine#17 is in, and together with flatcar-linux/scripts#82


marga-kinvolk:
This has now been fully tested as mentioned in flatcar-linux/bootengine#17

PR #497, opened 2020-07-27
Revert glibc update

pothos:
A linker issue hinders glibc to build. Revert https://github.com/flatcar-linux/coreos-overlay/pull/431 until we address the linker issue.

# How to use

# Testing done

Toolchains [build](http://localhost:9091/job/os/job/toolchains/404/console) started


dongsupark:
cherry-picked to edge

PR #498, opened 2020-07-27
metadata: Delete md5-cache

pothos:
The metadata/md5-cache folder is machine-generated based on the
other files in the repository. It causes merge conflicts when at
one time they were not regernated in a commit and then later a
commit does it and includes cache changes which	are incompatible
with later or newer states.
Remove the folder as it	is not necessary to have it and was removed
in upstream Gentoo, too.

# How to use

# Testing done

Running builds with a release base, without a release base, and a full SDK build.


marga-kinvolk:
This last commit seems unrelated to the md5-cache?

pothos:
Yes, will create a separate PR once testing is done.

pothos:
My plan was to retrigger a build but that doesn't work anymore because another branch is deleted, have removed the commit.

PR #504, opened 2020-07-30
Upgrade Rust in main from 1.44.1 to 1.45.0

dongsupark:
Upgrade Rust in main from 1.44.1 to 1.45.0.

It is the same one as https://github.com/flatcar-linux/coreos-overlay/pull/482, but its branch name is `rust-1.45.0-main` instead of `rust-1.45.0-alpha`.


dongsupark:
Closing in favor of https://github.com/flatcar-linux/coreos-overlay/pull/512

PR #505, opened 2020-07-30
sys-apps/systemd: Add tmpfile directive for /var/log/journal/remote

pothos:
The folder /var/log/journal/remote used to be part of the inital rootfs
through a keepdir directive in the build. However, any paths except
/usr are ephemeral and can be deleted at any time and should be recreated
with tmpfile directives. When upstream Gentoo removed the line
"keepdir /var/log/journal/remote" our tests started to fail but in fact
they could have failed earlier if they had tested with Ignition creating
a new root filesystem which lacks the /var/log/journal/remote folder.
Add a directive to create /var/log/journal/remote at runtime in any case.

# How to use

See https://github.com/flatcar-linux/mantle/pull/121

# Testing done

http://localhost:9091/job/os/job/kola/job/qemu/378/console


pothos:
Yes it can be created at build time and also is related to the `"keepdir /var/log/journal/remote"` line which Gentoo upstream removed, I think. But it doesn't fix it, it would just make the test pass because the test only runs on the initial rootfs and not when it was recreated with Ignition. The correct fix is to have a tmpfile directive.

PR #510, opened 2020-08-03
Build amazon-ssm-agent package but disable it for now

pothos:
The amazon-ssm-agent package was never built and caused the vm-matrix
    job to find no binary package.
    Build it as part of build_packages but don't install it on openstack
    or brightbox images. The plan is to add it on EC2 but currently the
    binaries are too large.

# How to use

`./build_packages
…
./build_image
./image_to_vm.sh --format=ami_vmdk …
./image_to_vm.sh --format=openstack …

# Testing done

Images build and AWS AMI runs.

### DIFF
--- a/coreos-base/oem-ec2-compat/files/base/base-ec2.ign
+++ b/coreos-base/oem-ec2-compat/files/base/base-ec2.ign
@@ -7,31 +7,6 @@
       {
         "name": "coreos-metadata-sshkeys@.service",
         "enabled": true
-      },
-      {
-        "name": "amazon-ssm-agent.service",
-        "enabled": true,
-        "contents": "[Unit]\nDescription=amazon-ssm-agent\nAfter=network-online.target\n\n[Service]\nType=simple\nWorkingDirectory=/usr/share/oem\nExecStart=/usr/share/oem/amazon-ssm-agent\nKillMode=process\nRestart=on-failure\nRestartForceExitStatus=SIGPIPE\nRestartSec=15min\n\n[Install]\nWantedBy=multi-user.target\n"
-      }
-    ]
-  },
-  "storage": {
-    "files": [
-      {
-        "filesystem": "root",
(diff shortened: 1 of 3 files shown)


samm-git:
Hi, does it mean that it will not be part of the official AMI-s? I been testing this and it **was** fitting OEM partition.

samm-git:
P.S. i was testing with ./image_to_vm.sh --format=ami_vmdk and it was fitting that partition.

cc: @pothos @sayanchowdhury

samm-git:
In fact that just removing any sense of ssm integration to the coreos. Is there any way this could be fixed?

samm-git:
Ah, so it is affecting only openstack or brightbox images ?  i think ssm here should be not in use anyway

samm-git:
But i think this PR is a bit misleading as it is telling  to not install on ` openstack
or brightbox images`. However, if i am not missing something - aws ec2 ami-s are also affected, no?

pothos:
Hi @samm-git – we had to disable inclusion for AWS EC2 AMIs. I'm also surprised that it doesn't fit and even tried to reduce the binary size but without success. The part about openstack and brightbox was that it got included into those images, too, but that wasn't the plan.

pothos:
I'm in favor of first remodeling how the OEM/vendor tool architecture works in general, so that we are not limited like we are now w.r.t. size limits and no updates being applied.

samm-git:
@pothos thank you for reply. Any way to contribute into that effort?

samm-git:
@pothos about reducing binary size - only way i see is to use golang way to build shared stdlib and later link binaries to it. I am not really sure if we have to do it for number of reasons:

1. It would significantly change way software is built, so we will be far from vendor build, more pain to support
1. Dynamic linking in golang is an experimental feature with very different support level on different archs. 
1. No guarantee that tomorrow it would still fit :)

So yes, +1 to change existing method of OEM. Also i saw that FCOS removed OEM partition entirely, so probably they also decided that it was not the best implementation

PR #513, opened 2020-08-06
.github: Import second signing key

pothos:
The last cork release was signed with a different key.
Import the key to verify the download.


marga-kinvolk:
Why was it signed with a different key? Won't this affect a lot more things as well?

pothos:
The other place where it was used is already updated: https://docs.flatcar-linux.org/os/sdk-modifying-flatcar/
We can return to the original key and then remove this additional step later again.

PR #514, opened 2020-08-07
Update packages for openssl update

pothos:
Together with https://github.com/flatcar-linux/portage-stable/pull/95 this updates openssl and any packages that need a to be updated with it. When updating a package I first reset the state to the Gentoo folder and then reapplied all downstream changes in one commit. I tried to mark ebuild changes with a Flatcar comment. Some downstream patches were dropped if they are not needed but I had to create other new changes to address build issues.

For openssh I reapplied the downstream changes in a separate commit because the existing commit mixed up the upstream update with the downstream changes but there are no actual changes done in this PR.

The new sssd tmpfile permissions are based on those that are installed to the initial root filesystem.

# How to use

Build image and SDK then verify if all things still work.

# Testing done

[No build failures for SDK or images from scratch, most tests completed already, some still running](http://localhost:9091/job/os/job/manifest/755/cldsv/)


pothos:
Picked for flatcar-2592

PR #526, opened 2020-08-13
Update sys-process/tini

krnowak:
This updates tini to match what gentoo has. I bumped the dep on tini in docker 17.03. Should be fine, since a quick scan of the tini's commit log didn't reveal any incompatible changes to me.

# How to use

I did `build_packages` inside chroot.

# Testing done

http://localhost:9091/job/os/job/manifest/792/cldsv/


pothos:
I haven't had a closer look yet. The way downstream patches are applied across the repo as I encountered is first importing the upstream Gentoo file in one commit, overwriting all customization and deleting patch files, and in the next commit applying the downstream patches. This means all changes are in one commit and the `diff A A.orig` is not needed.

krnowak:
> I haven't had a closer look yet. The way downstream patches are applied across the repo as I encountered is first importing the upstream Gentoo file in one commit, overwriting all customization and deleting patch files, and in the next commit applying the downstream patches. This means all changes are in one commit and the `diff A A.orig` is not needed.

If I understood you correctly - this makes it hard to figure out what our changes actually are. This is because the ebuild file can change in upstream gentoo without even bumping any revision. I find searching for the "base" upstream ebuild file annoying, so after I manage to do that, I prefer to store a copy as an `.orig` file and mention from which git commit I took it. Makes it easier later to figure out which changes to pick from upstream, because you have three files then - original "base" ebuild, new ebuild that will become a new "base" and our modified ebuild. That way you can make a more informed decision which our changes should be transplanted to the new "base" ebuild.

app-torcx/docker/docker-17.03.ebuild
```diff
@@ -15,7 +15,7 @@ RDEPEND="
 	~app-emulation/containerd-0.2.6
 	~app-emulation/docker-proxy-0.8.0_p20161019
 	~app-emulation/docker-runc-1.0.0_rc2_p136
-	=sys-process/tini-0.13.2
+	=sys-process/tini-0.18.0
```
dongsupark, line 18: You would probably want to do the same bump in `app-emulation/docker`.

dongsupark:
>> This means all changes are in one commit and the `diff A A.orig` is not needed.

I agree with Kai, I would simply remove A.orig.

> If I understood you correctly - this makes it hard to figure out what our changes actually are. This is because the ebuild file can change in upstream gentoo without even bumping any revision. I find searching for the "base" upstream ebuild file annoying, so after I manage to do that, I prefer to store a copy as an .orig file and mention from which git commit I took it. Makes it easier later to figure out which changes to pick from upstream, because you have three files then - original "base" ebuild, new ebuild that will become a new "base" and our modified ebuild. That way you can make a more informed decision which our changes should be transplanted to the new "base" ebuild.

That sounds like an interesting workflow.
I understand why you would want to keep the `.orig` file. It is indeed annoying to look into the upstream ebuild file every time.

However, what would happen when we need to bump tini to the next version, say, `0.19.0`?

We might need to do:
1) copy the upstream ebuild file into coreos-overlay, not into tini-0.18.0.ebuild but into tini-0.18.0.ebuild.orig.
2) rename tini-0.18.0.ebuild.orig to tini-0.19.0.ebuild.orig,
3) adjust tini-0.18.0.ebuild, so it fits into the new upstream change.
4) rename tini-0.18.0.ebuild to tini-0.19.0.ebuild.

(... and do the same things for all other packages)

In addition, we might need to ensure the committed `.orig` file in coreos-overlay really matches the upstream Gentoo's ebuild, without the suffix `.orig`, at least from time to time.

Sounds like fun. ;-)

pothos:
> this makes it hard to figure out what our changes actually are.

The overall changes are visible in the diff over two commits, like shown in GitHub. Yet first resetting the ebuild in one commit and then applying the downstream patches in another makes the single commits more useful because it allows me to really see the downstream changes in the Git history while otherwise I would need to run an extra command which with `git show HASH:FILE.orig > Aold.orig; git show HASH:FILE > Aold; diff Aold.orig Aold` becomes more complicated.

pothos:
Reviewing on GitHub is easier, too, if the first commit is split up into first resetting/updating and then applying downstream changes.

krnowak:
> > > This means all changes are in one commit and the `diff A A.orig` is not needed.
> 
> I agree with Kai, I would simply remove A.orig.
> 
> > If I understood you correctly - this makes it hard to figure out what our changes actually are. This is because the ebuild file can change in upstream gentoo without even bumping any revision. I find searching for the "base" upstream ebuild file annoying, so after I manage to do that, I prefer to store a copy as an .orig file and mention from which git commit I took it. Makes it easier later to figure out which changes to pick from upstream, because you have three files then - original "base" ebuild, new ebuild that will become a new "base" and our modified ebuild. That way you can make a more informed decision which our changes should be transplanted to the new "base" ebuild.
> 
> That sounds like an interesting workflow.
> I understand why you would want to keep the `.orig` file. It is indeed annoying to look into the upstream ebuild file every time.
> 
> However, what would happen when we need to bump tini to the next version, say, `0.19.0`?
> 
> We might need to do:
> 
>     1. copy the upstream ebuild file into coreos-overlay, not into tini-0.18.0.ebuild but into tini-0.18.0.ebuild.orig.
> 
>     2. rename tini-0.18.0.ebuild.orig to tini-0.19.0.ebuild.orig,
> 
>     3. adjust tini-0.18.0.ebuild, so it fits into the new upstream change.
> 
>     4. rename tini-0.18.0.ebuild to tini-0.19.0.ebuild.
> 

I suppose what I would do is:

1. copy upstream ebuild into coreos-overlay as tini-0.19.0.ebuild{,.orig}, commit
2. apply similar modifications to tini-0.19.0.ebuild which we did to tini-0.18.0.ebuild, commit
  - this requires a good documentation about what our changes actually are and why they are here (maybe some of the modification could be dropped in the new version?)
3. drop 0.18.0 stuff

But I dunno, it likely depends on an ebuild file - in some of them, we do a lot of modifications (like in systemd), so this workflow could be a PITA.

> 
> (... and do the same things for all other packages)
> 
> In addition, we might need to ensure the committed `.orig` file in coreos-overlay really matches the upstream Gentoo's ebuild, without the suffix `.orig`, at least from time to time.

The idea is that if you do an update, then the `.orig` file is basically the first thing you update.

Another change in the workflow could be having a file like `ebuild.orig`, so we don't need to rename it when version changes.

> 
> Sounds like fun. ;-)

For some twisted definition of fun, yes. :P

krnowak:
> > this makes it hard to figure out what our changes actually are.
> 
> The overall changes are visible in the diff over two commits, like shown in GitHub. Yet first resetting the ebuild in one commit and then applying the downstream patches in another makes the single commits more useful because it allows me to really see the downstream changes in the Git history while otherwise I would need to run an extra command which with `git show HASH:FILE.orig > Aold.orig; git show HASH:FILE > Aold; diff Aold.orig Aold` becomes more complicated.

Yet I have an uneasy feeling that this thing is not really often followed and git log often ends up being unhelpful. But I didn't do my share of digging in git logs to back this claim up, so I'll drop the `.orig` file. It's good to have at least a comment how to get the orig ebuild ourselves. But what it needs is a discipline in updating the comment when we update the ebuild again to match a new changes from upstream.

pothos:
Having a link to the original ebuild file makes sense and I'm also not against including it, but I like to keep the practice of one commit that contains all downstream changes. A random example (search for `: Apply (Flatcar|CoreOS) changes`) https://github.com/flatcar-linux/coreos-overlay/commit/640d6894a1cd788f4e3250878f2d31b801728ed3 which even links the previous commit https://github.com/flatcar-linux/coreos-overlay/commit/eb07324f4de3596de53eeb311d7961a71a88d87b as source for the changes but `git log FOLDER` would also lead to it.

app-torcx/docker/docker-17.03.ebuild
krnowak: Done.

krnowak:
Updated.

krnowak:
I suppose that instead of a RATIONALE file, we could have README.md.

t-lo:
@krnowak please consider merging to main.

PR #528, opened 2020-08-14
add qedf, qedi, qla4xxx modules

t-lo:
# Add qedf, qedi, qla4xxx kernel modules

This PR adds support for QLogic qedf, qedi, and qla4xxx SCSI devices via  CONFIG_QEDF, CONFIG_QEDI, CONFIG_SCSI_QLA_ISCSI respectively. Addressess https://github.com/flatcar-linux/Flatcar/issues/165 .

# Testing done
1. built a custom Flatcar Container Linux OS image based on the Alpha 2592.0.0 SDK.
2. in the resulting qemu image, ran `modprobe qedf; modprobe qedi; modprobe qla4xxx
3. ran `dmesg` - no errors reported, and `lsmod` to validate modules were loaded
4. did not verify functionally because of lack of access to hardware
5. Build system acceptance tests are still todo / pending (hence PR is in draft mode)


t-lo:
**cherry-pick candidate for next Alpha and Beta refreshes**

simonello:
Can we let me know when this is going to be released?   I can test these modules on my hardware if needs be.  If so, can you just point me to the iso.

marga-kinvolk:
We are working on a new major Alpha release that will include this change as well as a few others (version 2605). We're still ironing out the last details, but expect it to go out in the next couple of days.

dongsupark:
> I got the error "Target block device /dev/mapper/mpatha is not a full disk". I had to change line 360 in the /usr/bin/coreos-install script and add mpath.

Can you please create a PR to change https://github.com/flatcar-linux/init/blob/flatcar-master/bin/flatcar-install ?

PR #537, opened 2020-08-20
Update audit to 2.8.5

krnowak:
This PR updates audit package to 2.8.5-r1 from gentoo. Also adds a README to explain the reason for having it in coreos-overlay.

# How to use

See if audit message relevant to selinux show up in journalctl.

# Testing done

Downloaded the image and checked the above.


dongsupark (approved):

pothos (commented):
The import commit didn't delete the existing files which makes it hard to view the downstream patches like the added `audit-rules.tmpfiles` files and so on. I would first clean the folder, then import upstream, and finally apply our downstream patches. Otherwise it's really confusing in the future and archaeology has to start on each update – comments alone don't help as much as a`git diff` can do to see the total sum of downstream changes.

dongsupark (approved):

pothos (approved):

PR #542, opened 2020-08-25
README.md: Document how	to update packages

pothos:
The best practice established in this repository is to reset the
package	folder and import a new	version	from upstream without
modifications. In a separate commit the downstream changes are applied.
This makes it clear which downstream changes need to be	ported when
updating to a newer version in the future. Unfortunately this is not
always done which causes confusion and more work. As first step
document the process before we later look into more automation.

# How to use

Follow the guide when updating a package

# Testing done

Ran the `update_ebuilds` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,28 @@ that belong in the OEM partition. Any RDEPENDS from these packages should
 be copied to the RDEPENDS in `board-packages` to ensure they are built.
 
 `coreos-base/coreos-oem-*` are metapackages for OEM specific ACIs. 
+
+# Updating
+
+To update follow the following steps:
+
+* Remove or rename the whole folder of the package to prepare the import from
+  upstream Gentoo, not only resetting the ebuild file but also any additional
+  files like patches or downstream additions under `files`.
+* Run `~/trunk/src/scripts/update_ebuilds --portage_stable . CATEGORY/PACKAGE`
+  in the `coreos-overlay` folder to import a new version from upstream Gentoo.
+  Drop the ebuild files that you don't plan to use.
+* Commit the changes with a message like `CATEGORY/PACKAGE: Sync from Gentoo`,
+  and mention the the commit ID in the body (`git show update_ebuilds/master`).
+* Now find all downstream patches for the package by running
+  `git log CATEGORY/PACKAGE`. If everybody followed the process of resetting
+  before importing an upstream update, you only have to look for the commits
+  after the last update and port them to the new version. Otherwise you have
+  to compare the files manually to their upstream versions from older
+  [portage](https://github.com/gentoo/portage/) revisions.
+* You can combine all old and new downstream patches into a single new commit
+  with the message `CATEGORY/PACKAGE: Apply Flatcar patches` to keep the number of
+  commits to port low, or have separate commits. Make sure that you explain
+  the changes and carry the explanations from old commits over, either in the
+  commit message, through comments in the ebuild file, or through a `README.md`
+  in the folder.


t-lo (approved):
LGTM, thanks!

README.md
```diff
@@ -30,3 +30,25 @@ that belong in the OEM partition. Any RDEPENDS from these packages should
...
+# Updating
+
+To update follow the following steps:
+
+* Remove or rename the whole folder of the package to prepare the import from
+  upstream Gentoo, not only resetting the ebuild file but also any additional
+  files like patches or downstream additions under `files`.
+* Run `~/trunk/src/scripts/update_ebuilds --portage_stable . CATEGORY/PACKAGE`
+  in the `coreos-overlay` folder to import a new version from upstream Gentoo.
+* Commit the changes with a message like `CATEGORY/PACKAGE: Sync from Gentoo`.
```
vbatts, line 43: since we just copy, could we also include a hash to the gentoo repo commit from which it was sync'ed?

README.md
```diff
@@ -30,3 +30,25 @@ that belong in the OEM partition. Any RDEPENDS from these packages should
...
+
+# Updating
+
+To update follow the following steps:
+
+* Remove or rename the whole folder of the package to prepare the import from
+  upstream Gentoo, not only resetting the ebuild file but also any additional
+  files like patches or downstream additions under `files`.
+* Run `~/trunk/src/scripts/update_ebuilds --portage_stable . CATEGORY/PACKAGE`
+  in the `coreos-overlay` folder to import a new version from upstream Gentoo.
```
krnowak, line 42: I admit I never used that script. What I have is I have my own clone for gentoo repo and at this point I do `cp -a "${GENTOO_REPO}/{CATEGORY}/${PACKAGE}" "${CATEGORY}/${PACKAGE}"`, but I guess the script is going to do more or less the same.

But what I do next is that I drop all the ebuilds I don't need, so I only have one ebuild. And I drop all the patches and stuff from the `files` directory that are unused. Having multiple ebuilds is just confusing, especially that we often end up patching only one of them.

README.md
krnowak: I'd commit the synced version first, and then have a commit with a comment about the hash.

README.md
```diff
@@ -30,3 +30,25 @@ that belong in the OEM partition. Any RDEPENDS from these packages should
...
+  `git log CATEGORY/PACKAGE`. If everybody followed the process of resetting
+  before importing an upstream update, you only have to look for the commits
+  after the last update and port them to the new version. Otherwise you have
+  to compare the files manually to their upstream versions from older
+  [portage](https://github.com/gentoo/portage/) revisions.
+* Combine all old and new downstream patches into a single new commit with the
+  message `CATEGORY/PACKAGE: Apply Flatcar patches` to keep the number of
+  commits to port low. Make sure that you explain the changes and carry the
+  explanations from old commits over, either in the commit message or even
+  better through comments in the ebuild file.
```
krnowak, line 54: So far, I haven't followed that - I was mostly splitting the commits into smaller changes. But I think I'd be fine with having a single commit. About the explanations: comments in an ebuild are fine, but sometimes we drop a bunch of lines from an ebuild and a comment might look out of place. So it would be good to have a `README.md` file in the directory that explains the changes. So this file would need to be also updated/copied over every time we do an update, so it does not become stale.

Currently I'm working on adding the `README.md` files to each directory.

krnowak (commented):
I had a bit different process:
- copy ebuild + required files from gentoo
- add a comment about commit ID from gentoo repo this was copied from
- apply flatcar changes (in a single or multiple commits)
- add/update README.md (reason why this package is in coreos overlay, usually them being our own changes)
- drop old ebuild and now-unused files

This is more or less a process I followed in my branches so far ([tini](https://github.com/flatcar-linux/coreos-overlay/commits/krnowak/tini), [audit](https://github.com/flatcar-linux/coreos-overlay/commits/krnowak/audit), [pam](https://github.com/flatcar-linux/coreos-overlay/commits/krnowak/pam), [rules-fixes (moved mdadm to coreos-overlay)](https://github.com/flatcar-linux/coreos-overlay/commits/krnowak/fix-rules)).

README.md
pothos: Yes, will mention that unused ebuild files should be dropped.

README.md
pothos: I found having the explanations in the commit message helps useful but having a them in a README also works.

README.md
pothos: Having it in the body of the commit message would also work, or?
`git show update_ebuilds/master` will give the commit ID

krnowak (approved):

PR #543, opened 2020-08-25
baselayout, systemd: add group "render" for udev

t-lo:
### cherry-pick to: `flatcar-2605`.

# Add group "render" to OS image and to initramfs (for udev)

~This commit adds a dependency on acct-group/render to systemd. The respective group is provided by portage-stable (https://github.com/flatcar-linux/portage-stable commit ID db2ed1e74a89944b1500dba1471072e8da3dddc9).~

Furthermore, the baselayout commit ID is bumped to include a
change from https://github.com/flatcar-linux/baselayout - to
1d32bea2c0e4335d4a8f7e0ccd6a7b41da15e4a7 - which includes
this group in the baselayout used by initramfs.

Fixes https://github.com/flatcar-linux/Flatcar/issues/169 .

# How to use

1. Fetch the Flatcar SDK for Alpha-2605.0.0: `cork create --manifest-branch refs/tags/v2605.0.0`
2. In the Flatcar SDK, check out commit `870c692cf0b4933a949efddfb6a85e313685a7fe` in `src/third_party/coreos-overlay`, and commit `db2ed1e74a89944b1500dba1471072e8da3dddc9` in `src/third_party/portage-stable`.
3. Fetch and install pre-built binary packages to the image build root:
   ```
   $ cork enter
   $ ./setup_board --board=amd64-usr
   $ ./build_packages --board=amd64-usr
   ```
4. rebuild the kernel package to update the initrd: `$ emerge-amd64-usr coreos-kernel`
5. Build the OS image, build a qemu image, and start the image: 
    ```
    $ ./build_image --board=amd64-usr
    $ ./image_to_vm.sh --from=../build/images/amd64-usr/developer-latest --board=amd64-usr
    $ ../build/images/amd64-usr/developer-latest/flatcar_production_qemu.sh
   ```
6. ssh into the image, check journal for `/usr/lib64/udev/rules.d/50-udev-default.rules:39 Unknown group 'render', ignoring`:
   ```
   ssh core@localhost -p2222
   core@localhost $ journalctl -b
   ```

# Testing done

Ran the above steps, did not see the udev error anymore.


sys-apps/baselayout/baselayout-9999.ebuild
```diff
@@ -9,7 +9,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="2b8e4de42771033072324713ba8919248585d263"  # flatcar-master
+	CROS_WORKON_COMMIT="1d32bea2c0e4335d4a8f7e0ccd6a7b41da15e4a7"  # add group "render" for udev
```
dongsupark, line 12: > add group "render" for udev

The comment looks strange, as it is not a branch name any more.
Are we going to then write human-readable comments that represent the latest commit in the repo, changing it every time when the repo got updated? Sounds like overkill.
I would rather just keep it as `main`.

t-lo: Makes sense - I'll update the comment when / if https://github.com/flatcar-linux/baselayout/pull/5 gets merged.

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -85,6 +85,7 @@ COMMON_DEPEND=">=sys-apps/util-linux-2.30:0=[${MULTILIB_USEDEP}]
 	xkb? ( >=x11-libs/libxkbcommon-0.4.1:0= )"
 
 RDEPEND="${COMMON_DEPEND}
+	acct-group/render
```
sayanchowdhury, line 88: We already create the `render` group in the post pkg installation step. We did not add the acct-* packages from Gentoo upstream in the systemd ebuild was that would be more added maintenance for us.

sys-apps/baselayout/baselayout-9999.ebuild
sayanchowdhury: Just to add, I guess baselayout is still `flatcar-master` and not `main`.

sys-apps/systemd/systemd-9999.ebuild
t-lo: So we only need to add the baselayout change for initramfs? Could you please point to the sources where `render` gets added?

sayanchowdhury: Here is the link to the sources: https://github.com/flatcar-linux/coreos-overlay/blob/main/sys-apps/systemd/systemd-9999.ebuild#L536

I checked the acct-group, and it seems they run the commands in pre installation: https://github.com/gentoo/gentoo/blob/master/eclass/acct-group.eclass#L122-L127

t-lo: Thank you for the reference. I'll close my [PR for `acct-group/render`](https://github.com/flatcar-linux/portage-stable/pull/98) and update this PR. I'll also use the opportunity to define a GID for `render` and update my [baselayout PR](https://github.com/flatcar-linux/baselayout/pull/5) accordingly.

t-lo:
* Closed my PR for acct-group/render in portage-stable
* updated this PR, removed the dependency to `acct-group`, and explicitly set the GID of `render` to 30.
* updated the `baselayout` PR to use GID 30 in initramfs: https://github.com/flatcar-linux/baselayout/pull/5

t-lo:
I'll update the comment for `CROS_WORKON_COMMIT` after we've merged https://github.com/flatcar-linux/baselayout/pull/5 (and it actually is in flatcar-master).

t-lo:
https://github.com/flatcar-linux/baselayout/pull/5 was merged.

sys-apps/baselayout/baselayout-9999.ebuild
```diff
@@ -9,7 +9,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="2b8e4de42771033072324713ba8919248585d263"  # flatcar-master
+	CROS_WORKON_COMMIT="1d32bea2c0e4335d4a8f7e0ccd6a7b41da15e4a7"  # flatcar-master
```
sayanchowdhury, line 12: I've started to point this merge commit after Kai suggestion. In this case, that would be https://github.com/flatcar-linux/baselayout/commit/b67c163348c69b7c02dac1bb087b1a15cf21afa7

t-lo: Good catch, will update.

sayanchowdhury (approved):

dongsupark:
Nit: s/sytemd/systemd/ in both the PR title and the commit message. :-)

t-lo:
@dongsupark Good catch! Reworded.

t-lo:
Merged into main and cherry-picked to branch flatcar-2605.

PR #546, opened 2020-08-26
Fix udev rules in mdadm and lvm2

krnowak:
This PR imports the mdadm 4.1 into coreos-overlay to apply some patches for recent kernel (this follows what had been done in edge). The mdadm update fixes the udev rule at the same time. The rule in lvm2 is fixed with an extra patch.

It should be merged together with flatcar-linux/portage-stable#99

# How to use

Check if warnings listed in https://github.com/flatcar-linux/Flatcar/issues/164 do not show up any more.

Fixes flatcar-linux/Flatcar#164

# Testing done

Downloaded the image from os/manifest job 885 and checked for the warnings.


dongsupark:
In general it looks good. Great! :+1: 

Side note 1: lvm2 2.02.145 is really ancient, but we can update it to a recent version later. ;-)

Side note 2: MD RAID0 part looks perfect.
This reminds of a related pending [PR](https://github.com/flatcar-linux/mantle/pull/36) (still no review). So far mantle/kola has had test cases only for MD RAID1 for some reason. To make kola support MD RAID0 as well, I needed to introduce Go templates to be used for both RAID levels. It would be really great if you could have a look into the mantle/kola PR, and test it with the current mdadm changes. Of course only if you have some time. ;-)

krnowak:
> In general it looks good. Great! +1

Thanks!

> 
> Side note 1: lvm2 2.02.145 is really ancient, but we can update it to a recent version later. ;-)

We have also lvm2 2.02.166 (which is only slightly less really ancient) in coreos-overlay, but it's not picked up by build system for some reason. But yeah, updating lvm2 (and reducing the amount of ebuild files from 17 to 1) is certainly in plans.

> 
> Side note 2: MD RAID0 part looks perfect.
> This reminds of a related pending [PR](https://github.com/flatcar-linux/mantle/pull/36) (still no review). So far mantle/kola has had test cases only for MD RAID1 for some reason. To make kola support MD RAID0 as well, I needed to introduce Go templates to be used for both RAID levels. It would be really great if you could have a look into the mantle/kola PR, and test it with the current mdadm changes. Of course only if you have some time. ;-)

Will have a look and I'll try to test it. Thanks for letting me know about it.

t-lo (approved):
LGTM, thanks! Please also cherry-pick to `flatcar-2605`.

krnowak:
Cherry-pick done.

PR #554, opened 2020-08-28
sys-kernel: allow fips mode to be enabled

vbatts:
With this kernel config, users can boot with fips=1 set in
`/usr/share/oem/grub.cfg`:
```
set linux_append="fips=1"
```

Which triggers various behaviors, for FIPS 200 certification.

with this config compiled in, and that boot parameter, users can can
that fips is enabled with:
```
flatcar ~ # cat /proc/sys/crypto/fips_enabled
1
```

Signed-off-by: Vincent Batts <vbatts@kinvolk.io>


t-lo (approved):
LGTM, will merge and cherry-pick for `flatcar-2605`.

PR #556, opened 2020-08-28
dev-lang/go: Update the default Go version to 1.15

dongsupark:
Now that Go [1.15](Go://go.googlesource.com/go/+/refs/tags/go1.15) has been released, we should update the default Go version to 1.15.

Since the [EINTR issue](https://github.com/golang/go/issues/38033) was fixed in 1.15, we can simply move from 1.13 to 1.15.

On the other hand, we should not add 1.14, as the [EINTR bug fix](https://go-review.googlesource.com/c/go/+/232862/) was not backported to 1.14.

## How to use

```
./build_packages
./build_image
./image_to_vm.sh --from=../build/images/amd64-usr/developer-2592.0.0 --board=amd64-usr
```

## Testing done

`kola run` on a qemu image.


dongsupark:
A test looks good. Ready for review.

pothos (commented):
Looks good, let's wait for the test builds to finish

dongsupark:
Test builds passed.

pothos (approved):

PR #561, opened 2020-09-02
coreos-base: Allow installation to multipath target

dongsupark:
Allow `flatcar-install` to install Flatcar into a multipath target.

Pulls in https://github.com/flatcar-linux/init/pull/24

## How to use

```
emerge-amd64-usr coreos-base/coreos-init
```

/cc @simonello


marga-kinvolk (approved):
LGTM

PR #565, opened 2020-09-04
sys-kernel: Include patch for overflow in tpacket_rcv

pothos:
A memory corruption vulnerability in AF_PACKET causes the kernel to
    panic or enter undefined behavior, tracked as CVE-2020-14386.
    While the proposed patch is not included in an upstream release,
    include it as downstream patch.
    Further information and PoC:
    https://www.openwall.com/lists/oss-security/2020/09/03/3

**Note:** Will be picked for _main_.

# How to use

A PoC was published on oss-security.

# Testing done

The PoC does not trigger a panic.


marga-kinvolk (approved):
LGTM

PR #567, opened 2020-09-04
Update kernel to 5.8.6 and firmware to 20200817

krnowak:
In more detail, this PR (besides the stuff in subject):

- cleans up the config files a bit (remove duplicates, sort the files)
  - there is also a `sort_config.sh` script that does the grunt work, maybe will be useful later
- adds some more stuff to config files, because some new dependencies have shown up
- drops the wireguard patch, it's already a part of upstream
- fakes the existence of a firmware file, so coreos-firmware build doesn't fail (see the comment)

# How to use

# Testing done

I have tested the update to 5.8.5, not 5.8.6. Currently 5.8.6 is being built. Tested by building the image, booting, running `uname -r` to make sure that the kernel is indeed 5.8. Also did `sudo ip link add dev wg0 type wireguard` to make sure you can create a wireguard network interface.


t-lo (approved):
LGTM!
Testing done:
- did a local build against current Beta-2605.2.0 release with this branch; build succeeds
- started qemu image, booted fine, new kernel is running

Please merge so it's in the next nightly.

dongsupark:
Looks good in general.
And after you merged, please don't forget to update [KV_MAIN](https://github.com/flatcar-linux/coreos-overlay/blob/a9a360ae8360c2227d08dea1e4fb0013bf0f76a4/.github/workflows/kernel-releases-main.yml#L16) in Github actions for Kernel.
We can also update it in a separate PR.

krnowak:
I'll push the github action update too.

PR #571, opened 2020-09-07
.github: fix regexp to get correct semver from Go ebuilds

dongsupark:
Github Action for Go has had a bug when parsing the current Go version from `dev-lang/go/Manifest`, only when the current ebuild file has only major + minor versions, without patchlevel.
For example, it could parse well `1.13.15`, but not `1.15`.
We need to make it deal with both versions, `x.y.z` and `x.y`.

With this PR, for example, when `VERSION_SHORT` is `1.15` and the Manifest includes a tarball `go1.15.src.tar.gz`, we can confirm the new regexp works well like below:

```
$ sed -n "s/^DIST go\(1.15.*[0-9]*\)\.src.*/\1/p" dev-lang/go/Manifest
1.15
```

### DIFF
--- a/.github/workflows/go-apply-patch.sh
+++ b/.github/workflows/go-apply-patch.sh
@@ -15,7 +15,10 @@ fi
 
 pushd "${SDK_OUTER_SRCDIR}/third_party/coreos-overlay" >/dev/null || exit
 
-VERSION_OLD=$(sed -n "s/^DIST go\(${VERSION_SHORT}.[0-9]*\).*/\1/p" dev-lang/go/Manifest | sort -ruV | head -n1)
+# Parse the Manifest file for already present source files and keep the latest version in the current series
+# DIST go1.15.src.tar.gz ... => 1.15
+# DIST go1.15.3.src.tar.gz ... => 1.15.3
+VERSION_OLD=$(sed -n "s/^DIST go\(${VERSION_SHORT}\.*[0-9]*\)\.src.*/\1/p" dev-lang/go/Manifest | sort -ruV | head -n1)
 if [[ "${VERSION_NEW}" = "${VERSION_OLD}" ]]; then
   echo "already the latest Go, nothing to do"
   UPDATE_NEEDED=0


.github/workflows/go-apply-patch.sh
```diff
@@ -15,7 +15,7 @@ fi
 
 pushd "${SDK_OUTER_SRCDIR}/third_party/coreos-overlay" >/dev/null || exit
 
-VERSION_OLD=$(sed -n "s/^DIST go\(${VERSION_SHORT}.[0-9]*\).*/\1/p" dev-lang/go/Manifest | sort -ruV | head -n1)
+VERSION_OLD=$(sed -n "s/^DIST go\(${VERSION_SHORT}.*[0-9]*\).src.*/\1/p" dev-lang/go/Manifest | sort -ruV | head -n1)
```
marga-kinvolk, line 18: Can you add a comment specifying the patterns that this regex matches and what it does with them? If I understand this correctly, this would be it, right?

```
# Parse the Manifest file for already present source files and keep the latest version in the current series
DIST go1.15.src.tar.gz ... => 1.15
DIST go1.15.3.src.tar.gz ... => 1.15.3
```

There's a couple of issues with this regex. I guess it doesn't matter a lot, as it matches, but given that it's complex and it already had bugs, it might be good to make it more accurate:
 * The first dot should be `\.`, otherwise it's any char.
 * Similarly `.src` should be `\.src`.
 * The .* at the end is not needed, so I would drop that, it just adds noise.

dongsupark: Added comments as suggested.

> The first dot should be \., otherwise it's any char.
> Similarly .src should be \.src.

Those are done.

> The .* at the end is not needed, so I would drop that, it just adds noise.

I am afraid we cannot omit the trailing `.*`. See the example below:

```
$ sed -n "s/^DIST go\(1.15\.*[0-9]*\)\.src.*/\1/p" dev-lang/go/Manifest
1.15
$ sed -n "s/^DIST go\(1.15\.*[0-9]*\)\.src/\1/p" dev-lang/go/Manifest
1.15.tar.gz 23002901 BLAKE2B a04bfb5d8bb6377e92d5d8c96f11d39b9bb878f4f9a3221baa5f42bcc822b562a209fd5d2d2a643f3b72df2db494899b2f7a446f1a9a8e25bf0df56ef87f3b7f SHA512 7d85382bcc6a0625dfa3d07196ab364860846367ed67697a7b1516b0af551a72bc4873882141fc3c7a60d39f2e27b33f6693e8b18b608de76fc9a55b5eac55ea
```

The `/p` command of sed prints out the pattern space, and the output string of the pattern space is again replaced with the grouped `\1`, i.e. `1.15`. If the sed expression does not have the trailing `.*`, the remaining characters, starting from `.tar.gz 23002901...`, will simply stay out of the pattern space. That's why we would see such a bogus output in the example above.

I know, it is not exactly easy to read and understand.
I just did not have a better idea but such ugly sed expressions.

marga-kinvolk: Right, sorry for the wrong suggestion. I thought I had tested it, but clearly I made a mistake while testing. Thanks for explaining.

marga-kinvolk (approved):
LGTM

PR #574, opened 2020-09-08
dev-lang: Upgrade Go 1.15 to 1.15.2

dongsupark:
Upgrade Go to 1.15.2, to address issues like potential cross-site scripting
vulnerability, as described in
https://groups.google.com/g/golang-announce/c/8wqlSbkLdPs .

## How to use

```
emerge dev-lang/go
```

*Note: Since Github actions for Go will not run again until the next Monday, I am simply creating this PR manually.*


krnowak:
go 1.15.2 was released today, would it make sense to package that instead?

krnowak:
Nevermind, there seems to be an issue with go1.15 anyway, which go1.15.2 does not fix, so we will most likely end up reverting the update to go 1.15.

marga-kinvolk:
The issue with go1.15 gets solved by updating the mantle ebuild (#577). You can go ahead and merge go1.15.1 or .2.

dongsupark:
> go 1.15.2 was released today, would it make sense to package that instead?

Thanks.
Updated Go version to 1.15.2.

krnowak (approved):
Looks good.

PR #575, opened 2020-09-09
.github: improve body text of PR for dev-lang/rust

dongsupark:
Improve body text of each PR for `dev-lang/rust`, by mentioning that it should be merged together with its paired PR in portage-stable.

Explicitly name `dev-lang/rust` instead of `Rust`, because now there are `dev-lang/rust` as well as `virtual/rust`.
Rename the dispatched event-type name to `rust-pull-request-main`, as `cargo` has already disappeared.

Make the repository-dispatch action send additional client-payload with a field `coreos-overlay-pull-request-number`, which will be later used by the corresponding PR in portage-stable for adding a link back to the PR in coreos-overlay.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/103.

### DIFF
--- a/.github/workflows/rust-apply-patch.sh
+++ b/.github/workflows/rust-apply-patch.sh
@@ -29,7 +29,7 @@ popd >/dev/null || exit
 
 popd >/dev/null || exit
 
-generate_patches dev-lang rust Rust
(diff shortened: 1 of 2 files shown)


marga-kinvolk (approved):
LGTM, thanks!

PR #576, opened 2020-09-09
Revert "dev-lang/go: Update the default Go version to 1.15"

marga-kinvolk:
We've been seeing a bunch of issues related to signature validations, when using Go version 1.15.  Initial investigation seems to point to this change https://go-review.googlesource.com/c/go/+/226203/2/src/crypto/rsa/pkcs1v15.go, which added an additional check to the signature validation related to the length of the signatures.

The failures we've seen were related to signatures that were 593 bytes long, while 594 byte signatures worked correctly.

This reverts flatcar-linux/coreos-overlay#556


pothos (approved):
Any preference on whether a merge commit is reverted or the individual commits? I tend towards reverting the commits directly but am fine with both up to the points where reverted merge commits are reverted and it gets confusing for me but maybe that can't be avoided :D

marga-kinvolk:
It turns out the problem was in our mantle ebuild. We don't need to revert after all.

PR #582, opened 2020-09-10
sys-apps/systemd: Bump to v246.6

krnowak:
This bumps systemd to 246.6 stable release. It's a bit more involved, because of a breaking change in systemd-tmpfiles (it now requires a user and group data in the rootfs passed with the `--root` flag; it means it does not read the user and group info anymore from the host rootfs).

# How to use

? Well, boot it?

# Testing done

Built the image, it boots. Since the update logic is modified, I need to test updating too.

Test build: http://localhost:9091/job/os/job/manifest/1138/

Needs flatcar-linux/bootengine#18 and flatcar-linux/baselayout#6. flatcar-linux/portage-stable#112 is optional, but useful.


dongsupark (approved):
Looks good, if CI passed.

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -177,10 +179,6 @@ src_unpack() {
 
 src_prepare() {
 	# Flatcar: We don't have separate patches, so no patching code here.
-	#
-	# Flatcar: Use the resolv.conf managed by systemd-resolved.
-	sed -i -e 's,/run/systemd/resolve/stub-resolv.conf,/run/systemd/resolve/resolv.conf,' tmpfiles.d/etc.conf.m4 || die
```
pothos, line 182: I think this will break all cases where 127.0.0.53 is not reachable because it's in a different namespace. I'll file a PR to revert this.

sys-apps/baselayout/baselayout-9999.ebuild
```diff
@@ -113,6 +113,17 @@ src_install() {
 	for sym in "${!LIB_SYMS[@]}" ; do
 		dosym "${LIB_SYMS[$sym]}" "${sym}"
 	done
+
+	if use cros_host; then
+		# We assume that SDK already has the user and group database.
```
pothos, line 118: The folder `/home/core/` is gone from the SDK which could be related to this change. `cork` fails to use the SDK because `useradd: cannot create directory /home/build` but I don't know if this is also related.

pothos: Ah, no, this is just moved from below, then my search continues.

pothos: This was the problem because of the changed behavior. I wonder if we could use UIDs for systemd-tmpfiles. Anyway, the PR with a workaround is here https://github.com/flatcar-linux/coreos-overlay/pull/639

PR #583, opened 2020-09-11
coreos-base/coreos-init: Update to include Calico tunl0 networkd units

pothos:
This pulls in
https://github.com/flatcar-linux/init/pull/26

Note: Pick for 2605


t-lo (approved):
LGTM, thanks!

PR #495, opened 2020-07-24
app-emulation/qemu: Upgrade qemu to 5.0.0

sayanchowdhury:
## app-emulation/qemu: Upgrade qemu to 5.0.0
Note: This is not a security fix, as we don't use QEMU anywhere else other than the SDK

Upgrade QEMU to 5.0.0 to fix the following security issues in SDK:

- [CVE-2020-11102](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11102)
- [CVE-2020-1711](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-1711)
- [CVE-2020-7039](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-7039)

Signed-off-by: Sayan Chowdhury <sayan@kinvolk.io>

# How to use

`emerge-amd64-usr app-emulation/qemu`

# Testing done

I've done a local build for the package

PR #591, opened 2020-09-17
app-emulation/qemu: Fix build errors

marga-kinvolk:
The qemu update done in #495 caused several errors:

* We currently don't have Python 3.8 available in the SDK, so adding it in the PYTHON_COMPAT field causes a build failure.
* The manifest needed to be updated
* A patch file was missing

This commit fixes these errors and makes the package build.


dongsupark (approved):
Looks good. Thanks!

PR #599, opened 2020-09-18
sys-kernel/coreos-modules: remove deprecated config

marga-kinvolk:
We were setting `CONFIG_VGACON_SOFT_SCROLLBACK=y`, but this config option was deleted with [20782abbbdfe922496a28f9cc0c3c0030f7dfb8f](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-5.8.y&id=20782abbbdfe922496a28f9cc0c3c0030f7dfb8f), due to security issues.

I've tested that the kernel builds correctly now.


t-lo (approved):
LGTM, thanks for the quick reaction!

PR #600, opened 2020-09-18
dev-db/sqlite: update to 3.32.3 to keep non-full archive

dongsupark:
Update `dev-db/sqlite` to 3.32.3, syncing with Gentoo upstream.
It is to mainly address a security issue [CVE-2020-11656](https://nvd.nist.gov/vuln/detail/CVE-2020-11656).

To do that, bring back builds for non-full archive.

Since sqlite 3.32.0, Gentoo ebuild does not deal with non-full archive, but fetches only full archive. On top of that, the upstream sqlite's full archive requires `tclsh` to be installed on the host system. Since Flatcar SDK does not include `dev-lang/tcl`, it is not possible to build sqlite from the full-archive.

It means that we need to either make the Flatcar SDK include `dev-lang/tcl`, (which takes time) or bring back the non-full archive mechanism just like ebuilds from sqlite 3.31.x.

So adapt the full-archive patches on top of the non-full archive. Make the ebuild fetch the non-full archive.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/107.

## How to use

```
emerge-amd64-usr dev-db/sqlite
```

## Testing done

CI passed


dev-db/sqlite/README.md
```diff
@@ -0,0 +1,11 @@
+This is a fork of gentoo's `dev-db/sqlite` package. The main
+reasons for having our fork are:
+
+1. Even in case of sqlite >= 3.32.0, we still need to keep conditions to
+   distinguish full archive from non-full archive. Main reason for that is
+   because it is not possible to build sqlite from its full archive (i.e.
+   "-src"), which depends `dev-lang/tcl`. Either we need to make
+   `dev-lang/tcl` available in the Flatcar SDK, or we need to bring back
+   the ability of dealing with its non-full archive. (i.e. "-autoconf")
+   Obviously the latter is much easier, so we bring back the non-full
```
marga-kinvolk, line 10: Why is it easier to maintain a fork than to include dev-lang/tcl in the SDK? This seems counter-intuitive to me. I'd have expected the other one to be easier.

dongsupark: You are right. Let me rephrase:
```
While the former is ideal way to go in the future, it will require additional updates in the SDK-only release.
The latter is easier for us to resolve the on-going security issues in the short term.
```

marga-kinvolk: Sure, yes, this makes a lot more sense :)

dongsupark: Updated the README.

dongsupark:
CI passed.

marga-kinvolk (approved):
LGTM, thanks for the clarification!

PR #603, opened 2020-09-23
Switch to our new mirrors

marga-kinvolk:
We've created flatcar mirrors to stop depending on coreos mirrors. This change switches to those mirrors instead of the deprecated coreos ones.

I've verified that the SDK job picked up this change when building from this branch.

This will need to be cherry picked to all active branches.

Needed for: kinvolk/PROJECT-flatcar-linux#333


pothos (approved):

marga-kinvolk:
Cherry-picked to flatcar-2605 and flatcar-2632

PR #608, opened 2020-09-24
coreos-base/coreos: Build and install oslogin for amd64 only

pothos:
The build for arm64 currently fails because it tries to build the
oslogin package but the package is marked as amd64-only.
Exclude the oslogin package from arm64 images.

# How to use

Build an arm64 image.

# Testing done

None yet


marga-kinvolk (approved):
LGTM

PR #610, opened 2020-09-25
Improve overall behavior on GCE

marga-kinvolk:
# Improve overall behavior on GCE

These commits make our GCE images behave correctly when using oslogin, as well as provide an alias for the `python` and `python3` commands running in containers.

Note: this PR incorporates three commits that have already been applied to main, the intention is to cherry-pick them to both flatcar-2632 and flatcar-2605, so that our beta and stable channels also have this behavior, which is needed for our official GCE images.

# How to use / Testing done

Already built and tested in the main branch.


pothos (approved):

marga-kinvolk:
Cherry picked to flatcar-2605

PR #612, opened 2020-09-25
Enable remote-fs

krnowak:
Since v242 remote-fs.target was not enabled by default, so mounting NFS shares did not happen after the update from 2512 (which has systemd 241). The ebuild was enabling remote-fs though, by putting a symlink in `/etc/systemd/system/…`. But this change in `/etc/ is apparently propagated during update, so remote-fs.target ended up being not a dependent target during boot. So the PR tries harder at not enabling services through /etc, but through /lib.

# How to use

# Testing done

The image is building for me at the moment, but:

```
find /build/amd64-usr/usr/lib/systemd/ | grep 'remote-fs'
/build/amd64-usr/usr/lib/systemd/system/remote-fs.target
/build/amd64-usr/usr/lib/systemd/system/remote-fs-pre.target
/build/amd64-usr/usr/lib/systemd/system/multi-user.target.wants/remote-fs.target # that used to be missing
/build/amd64-usr/usr/lib/systemd/system/remote-fs.target.wants
/build/amd64-usr/usr/lib/systemd/system/remote-fs.target.wants/var-lib-machines.mount
```

(in comparison with current stable)

```
Flatcar Container Linux by Kinvolk stable (2605.5.0)
core@localhost ~ $ find /usr/lib/systemd/ | grep remote-fs
/usr/lib/systemd/system/remote-fs.target.wants
/usr/lib/systemd/system/remote-fs.target.wants/var-lib-machines.mount
/usr/lib/systemd/system/remote-fs-pre.target
/usr/lib/systemd/system/remote-fs.target
```

Fixes flatcar-linux/Flatcar#191

To be propagated to flatcar-2605 and flatcar-2632.

### DIFF
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -433,6 +433,12 @@ multilib_src_install_all() {
 	# Flatcar: getty@.service is enabled manually below.
 	systemd_enable_service sysinit.target systemd-timesyncd.service
 	systemd_enable_service multi-user.target systemd-networkd.service
+	# For systemd-networkd.service, it has it in Also, which also
+	# needs to be enabled
+	systemd_enable_service sockets.target systemd-networkd.socket
+	# For systemd-networkd.service, it has it in Also, which also
+	# needs to be enabled
+	systemd_enable_service network-online.target systemd-networkd-wait-online.service
 	systemd_enable_service multi-user.target systemd-resolved.service
 	# Flatcar: not enabling reboot.target - it has no WantedBy
 	# entry.
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -507,7 +507,9 @@ save_enabled_units() {
 }
 
 pkg_preinst() {
-	save_enabled_units {machines,remote-{cryptsetup,fs}}.target getty@tty1.service
+	# Flatcar: We enable getty and remote-fs unconditionally, so
+	# no need to find out whether those are enabled or not here.
+	save_enabled_units {machines,remote-cryptsetup}.target
 
 	if ! use split-usr; then
 		local dir
@@ -559,11 +561,19 @@ pkg_postinst() {
 	# Flatcar: Reenabling systemd-timesyncd service too.
 	systemd_reenable systemd-networkd.service systemd-resolved.service systemd-timesyncd.service
 
+	# Flatcar: TODO: Possibly replace `systemctl enable` with
+	# `systemd_enable_service`, so it potentially could enable the
+	# services by making symlinks in /lib, instead of /etc?
+	# Currently does nothing, because ENABLED_UNITS end up being
+	# empty, but might not be true in future.
 	if [[ ${ENABLED_UNITS[@]} ]]; then
 		systemctl --root="${ROOT:-/}" enable "${ENABLED_UNITS[@]}"
 	fi
 
-	if [[ -z ${REPLACING_VERSIONS} ]]; then
+	# Flatcar: We enable getty and remote-fs targets ourselves
+	# above. The code below would modify /etc, which we don't
+	# want.
+	if false && [[ -z ${REPLACING_VERSIONS} ]]; then
 		if type systemctl &>/dev/null; then
 			systemctl --root="${ROOT:-/}" enable getty@.service remote-fs.target || FAIL=1
 		fi
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -426,14 +426,15 @@ multilib_src_install_all() {
 	# Flatcar: Don't enable services in /etc, move to /usr.
 	systemd_enable_service multi-user.target systemd-networkd.service
 	systemd_enable_service multi-user.target systemd-resolved.service
+	systemd_enable_service multi-user.target remote-fs.target
 	systemd_enable_service sysinit.target systemd-timesyncd.service
 
 	# Flatcar: Enable getty manually.
 	mkdir --parents "${ED}/usr/lib/systemd/system/getty.target.wants"
 	dosym ../getty@.service "/usr/lib/systemd/system/getty.target.wants/getty@tty1.service"
 
-	# Flatcar: Do not enable random services if /etc was detected
-	# as empty!!!
+	# Flatcar: Use an empty preset file, because systemctl
+	# preset-all puts symlinks in /etc, not in /usr.
 	rm "${ED}$(usex split-usr '' /usr)/lib/systemd/system-preset/90-systemd.preset" || die
 	insinto $(usex split-usr '' /usr)/lib/systemd/system-preset
 	doins "${FILESDIR}"/99-default.preset
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -505,20 +505,14 @@ migrate_locale() {
 	fi
 }
 
-save_enabled_units() {
-	ENABLED_UNITS=()
-	type systemctl &>/dev/null || return
-	for x; do
-		if systemctl --quiet --root="${ROOT:-/}" is-enabled "${x}"; then
-			ENABLED_UNITS+=( "${x}" )
-		fi
-	done
-}
+# Flatcar: save_enabled_units function is dropped, because it's
+# unused. When building releases, we assume that there was no systemd
+# previously, so there are no units to remember.
 
 pkg_preinst() {
-	# Flatcar: We enable getty and remote-fs unconditionally, so
-	# no need to find out whether those are enabled or not here.
-	save_enabled_units {machines,remote-cryptsetup}.target
+	# Flatcar: When building releases, we assume that there was no
+	# systemd previously, so there are no units to remember, so
+	# there is no point in calling save_enabled_units.
 
 	if ! use split-usr; then
 		local dir
@@ -567,28 +561,16 @@ pkg_postinst() {
 	# between OpenRC & systemd
 	migrate_locale
 
-	# Flatcar: Reenabling systemd-timesyncd service too.
-	systemd_reenable systemd-networkd.service systemd-resolved.service systemd-timesyncd.service
+	# Flatcar: Dropping the reenabling, since there earlier there
+	# was no systemd (we are building the release from scratch
+	# here). The function checks if the unit is enabled before
+	# running reenable, which in our case results in no action at
+	# all (because no service is enabled).
 
-	# Flatcar: TODO: Possibly replace `systemctl enable` with
-	# `systemd_enable_service`, so it potentially could enable the
-	# services by making symlinks in /lib, instead of /etc?
-	# Currently does nothing, because ENABLED_UNITS end up being
-	# empty, but might not be true in future.
-	if [[ ${ENABLED_UNITS[@]} ]]; then
-		systemctl --root="${ROOT:-/}" enable "${ENABLED_UNITS[@]}"
-	fi
+	# Flatcar: Dropping handling of ENABLED_UNITS.
 
-	# Flatcar: We enable getty and remote-fs targets ourselves
-	# above. The code below would modify /etc, which we don't
-	# want.
-	if false && [[ -z ${REPLACING_VERSIONS} ]]; then
-		if type systemctl &>/dev/null; then
-			systemctl --root="${ROOT:-/}" enable getty@.service remote-fs.target || FAIL=1
-		fi
-		elog "To enable a useful set of services, run the following:"
-		elog "  systemctl preset-all --preset-mode=enable-only"
-	fi
+	# Flatcar: We enable getty and remote-fs targets in /usr
+	# ourselves above.
 
 	if [[ -L ${EROOT}/var/lib/systemd/timesync ]]; then
 		rm "${EROOT}/var/lib/systemd/timesync"
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -423,11 +423,20 @@ multilib_src_install_all() {
 	# Flatcar: Don't set any extra environment variables by default.
 	rm "${ED}/usr/lib/environment.d/99-environment.conf" || die
 
-	# Flatcar: Don't enable services in /etc, move to /usr.
+	# Flatcar: These lines more or less follow the systemd's
+	# preset file (90-systemd.preset). We do it that way, to avoid
+	# putting symlink in /etc. Please keep the lines in the same
+	# order as the "enable" lines appear in the preset file.
+	systemd_enable_service multi-user.target remote-fs.target
+	systemd_enable_service multi-user.target remote-cryptsetup.target
+	systemd_enable_service multi-user.target machines.target
+	# Flatcar: getty@.service is enabled manually below.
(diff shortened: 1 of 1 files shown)


sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -558,11 +561,19 @@ pkg_postinst() {
...
+	# empty, but might not be true in future.
 	if [[ ${ENABLED_UNITS[@]} ]]; then
 		systemctl --root="${ROOT:-/}" enable "${ENABLED_UNITS[@]}"
 	fi
 
-	if [[ -z ${REPLACING_VERSIONS} ]]; then
+	# Flatcar: We enable getty and remote-fs targets ourselves
+	# above. The code below would modify /etc, which we don't
+	# want.
+	if false && [[ -z ${REPLACING_VERSIONS} ]]; then
```
pothos, line 576: Can we also remove this whole `if` block?

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -426,14 +426,15 @@ multilib_src_install_all() {
...
 	systemd_enable_service sysinit.target systemd-timesyncd.service
 
 	# Flatcar: Enable getty manually.
 	mkdir --parents "${ED}/usr/lib/systemd/system/getty.target.wants"
 	dosym ../getty@.service "/usr/lib/systemd/system/getty.target.wants/getty@tty1.service"
 
-	# Flatcar: Do not enable random services if /etc was detected
-	# as empty!!!
+	# Flatcar: Use an empty preset file, because systemctl
+	# preset-all puts symlinks in /etc, not in /usr.
```
pothos, line 437: Good idea to make the reason more clear. Maybe we can also add that we don't use `/etc` because it's not auto-updated.

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -558,11 +561,19 @@ pkg_postinst() {
 	# Flatcar: Reenabling systemd-timesyncd service too.
 	systemd_reenable systemd-networkd.service systemd-resolved.service systemd-timesyncd.service
```
pothos, line 562: It's a bit longer ago that I touched this, what is the difference between `systemd_reenable` and `save_enabled_units`?

pothos: Maybe add `getty@tty1.service` here?

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -426,14 +426,15 @@ multilib_src_install_all() {
 	# Flatcar: Don't enable services in /etc, move to /usr.
 	systemd_enable_service multi-user.target systemd-networkd.service
 	systemd_enable_service multi-user.target systemd-resolved.service
+	systemd_enable_service multi-user.target remote-fs.target
```
pothos, line 429: Some other files are also missing, I guess we need something like this (unless upstream deprecated it):
```
multi-user.target machines.target
multi-user.target remote-cryptsetup.target
network-online.target systemd-networkd-wait-online.service
sockets.target systemd-networkd.socket
```

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -558,11 +561,19 @@ pkg_postinst() {
 	# Flatcar: Reenabling systemd-timesyncd service too.
 	systemd_reenable systemd-networkd.service systemd-resolved.service systemd-timesyncd.service
 
+	# Flatcar: TODO: Possibly replace `systemctl enable` with
```
pothos, line 564: Looking at `portage-stable/eclass/systemd.eclass` where `systemd_reenable` is defined I guess we can remove the `ENABLED_UNITS` usage and the `save_enabled_units` call for it when we solely use `systemd_reenable`.

sys-apps/systemd/systemd-9999.ebuild
krnowak: I can. My change effectively does that, without removing it. :)

I was just wondering what to aim at: at the minimal divergence from upstream ebuild (change one line) or for the lowest cognitive overload here (drop superfluous stuff). But since we already dropped some lines, I can drop this one too then.

sys-apps/systemd/systemd-9999.ebuild
krnowak: Good idea, will update it.

sys-apps/systemd/systemd-9999.ebuild
krnowak: "reenable" is "disable" + "enable". `save_enabled_units` is supposed to remember which of the listed units were remembered in preinst step, so in postinst, we enable them again. But the enabling means putting stuff to `/etc`, since it's done with `systemctl enable|reenable`. We don't want that. So I guess we should rewrite those, or make sure that they are not actually called?

sys-apps/systemd/systemd-9999.ebuild
krnowak: I was wondering whether to basically translate all the preset stuff from systemd into `system_enable_service`. For now I did that only for remote-fs. Can update it here too.

sys-apps/systemd/systemd-9999.ebuild
krnowak: So, I think we shouldn't be using this at all, since they all put stuff into `/etc`.

sys-apps/systemd/systemd-9999.ebuild
pothos: Reenable is used for `systemd-networkd.service systemd-resolved.service systemd-timesyncd.service` and all are correctly placed under `/usr`.
I don't know if it has any unwanted side effect if we also used if for `machines.target` or `remote-*.target`. It seems both `save_enabled_units` and `systemd_reenable` only work if the units were enabled – I guess as long as we have the result we want it's ok to use both as needed.

sys-apps/systemd/systemd-9999.ebuild
pothos: See comment above, it's correctly working.

pothos: Or do you mean not using `ENABLED_UNITS`?

sys-apps/systemd/systemd-9999.ebuild
marga-kinvolk: Deleted stuff is quite hard to handle when diffing new versions. Either comment it out or do what you did here, but just deleting lines can be very confusing, because then you don't know whether those lines are new or they should stay deleted.

marga-kinvolk (approved):
This looks good to me as is. We can re-iterate to make it better, but I'd like for this to land in Monday's patch release.

sys-apps/systemd/systemd-9999.ebuild
krnowak: Dropped.

sys-apps/systemd/systemd-9999.ebuild
krnowak: Comment updated.

sys-apps/systemd/systemd-9999.ebuild
krnowak: This line got dropped, since it was basically a noop.

sys-apps/systemd/systemd-9999.ebuild
krnowak: I added `multi-user.target machines.target` and `multi-user.target remote-cryptsetup.target`. I also added `remount-fs.target systemd-pstore.service`. I don't know where did you get the other two - they are not listed in upstream's preset file.

sys-apps/systemd/systemd-9999.ebuild
krnowak: This got dropped.

sys-apps/systemd/systemd-9999.ebuild
krnowak: @marga-kinvolk: I did not see your comment when I wrote it, sorry about that. Frankly speaking, commenting the lines out somehow didn't come to my mind…

sys-apps/systemd/systemd-9999.ebuild
pothos: They come from the 2512.5.0 Stable release. This commit here c07b410caf58921a4d39f6b406c4ea8a3136514e has a list of CoreOS services.

krnowak: So `network-online.target systemd-networkd-wait-online.service` is not necessary I think, because there is a comment in `systemd-networkd.service` which we enable:

```
# We want to enable systemd-networkd-wait-online.service whenever this service
# is enabled. systemd-networkd-wait-online.service has
# WantedBy=network-online.target, so enabling it only has an effect if
# network-online.target itself is enabled or pulled in by some other unit.
Also=systemd-networkd-wait-online.service
```

Not sure about `sockets.target systemd-networkd.socket` - `systemd-networkd.socket` is pulled in by `systemd-networkd.service` also with the `Also=` option. So I guess it's not necessary anymore too.

pothos: I checked and the `Also=` is not enough: after `sudo systemctl status network-online.target` I saw that `systemd-networkd-wait-online.service` is inactive while on 2512.2.0 it is active afterwards.

pothos: Same situation for `systemd-networkd.socket` which gets activated on 2512.2.0 but not on 2605.5.0

krnowak:
Tested the PR with following ignition config:
```
{
  "ignition": {
    "config": {},
    "security": {
      "tls": {}
    },
    "timeouts": {},
    "version": "2.3.0"
  },
  "networkd": {},
  "passwd": {},
  "storage": {},
  "systemd": {
    "units": [
      {
        "contents": "[Unit]\nBefore=remote-fs.target\n[Service]\nType=oneshot\nRemainAfterExit=yes\nExecStart=/bin/echo ajwaj!\n[Install]\nWantedBy=remote-fs.target\n",
        "enable": true,
        "name": "my-testing-unit.service"
      }
    ]
  }
}
```

status of remote-fs.target:

```
core@localhost ~ $ systemctl status remote-fs.target
● remote-fs.target - Remote File Systems
     Loaded: loaded (/usr/lib/systemd/system/remote-fs.target; disabled; vendor preset: disabled)
     Active: active since Sat 2020-09-26 13:57:10 UTC; 3min 48s ago
       Docs: man:systemd.special(7)

Sep 26 13:57:10 localhost systemd[1]: Reached target Remote File Systems.
```

status of my testing unit:

```
core@localhost ~ $ systemctl status my-testing-unit.service
● my-testing-unit.service
     Loaded: loaded (/etc/systemd/system/my-testing-unit.service; enabled; vendor preset: enabled)
     Active: active (exited) since Sat 2020-09-26 13:57:10 UTC; 17s ago
    Process: 805 ExecStart=/bin/echo ajwaj! (code=exited, status=0/SUCCESS)
   Main PID: 805 (code=exited, status=0/SUCCESS)

Sep 26 13:57:10 localhost systemd[1]: Starting my-testing-unit.service...
Sep 26 13:57:10 localhost echo[805]: ajwaj!
Sep 26 13:57:10 localhost systemd[1]: Finished my-testing-unit.service.
```

contents of /etc/systemd/system:

```
core@localhost ~ $ ls -laR /etc/systemd/system
/etc/systemd/system:
total 28
drwxr-xr-x. 3 root root 4096 Sep 26 13:57 .
drwxr-xr-x. 6 root root 4096 Sep 26 13:57 ..
-rw-r--r--. 1 root root  137 Sep 26 13:57 my-testing-unit.service
drwxr-xr-x. 2 root root 4096 Sep 26 13:57 remote-fs.target.wants

/etc/systemd/system/remote-fs.target.wants:
total 20
drwxr-xr-x. 2 root root 4096 Sep 26 13:57 .
drwxr-xr-x. 3 root root 4096 Sep 26 13:57 ..
lrwxrwxrwx. 1 root root   43 Sep 26 13:57 my-testing-unit.service -> /etc/systemd/system/my-testing-unit.service
```

remote-fs in /usr/lib/systemd/system:

```
core@localhost ~ $ find /usr/lib/systemd/system | grep remote-fs
/usr/lib/systemd/system/remote-fs-pre.target
/usr/lib/systemd/system/multi-user.target.wants/remote-fs.target
/usr/lib/systemd/system/remote-fs.target.wants
/usr/lib/systemd/system/remote-fs.target.wants/var-lib-machines.mount
/usr/lib/systemd/system/remote-fs.target
```

Will merge and cherry-pick to other branches.

sys-apps/systemd/systemd-9999.ebuild
krnowak: Added those, but I don't see any change, maybe it's something to investigate later:

systemd-networkd-wait-online.service:

```
core@localhost ~ $ systemctl status systemd-networkd-wait-online.service
● systemd-networkd-wait-online.service - Wait for Network to be Configured
     Loaded: loaded (/usr/lib/systemd/system/systemd-networkd-wait-online.service; disabled; vendor preset: disabled)
     Active: inactive (dead)
       Docs: man:systemd-networkd-wait-online.service(8)
```
systemd-networkd.socket:

```
core@localhost ~ $ systemctl status systemd-networkd.socket
● systemd-networkd.socket - Network Service Netlink Socket
     Loaded: loaded (/usr/lib/systemd/system/systemd-networkd.socket; disabled; vendor preset: disabled)
     Active: active (running) since Sat 2020-09-26 13:57:07 UTC; 7min ago
   Triggers: ● systemd-networkd.service
       Docs: man:systemd-networkd.service(8)
             man:rtnetlink(7)
     Listen: route 1361 (Netlink)
     CGroup: /system.slice/systemd-networkd.socket
```

systemd-networkd.service:

```
core@localhost ~ $ systemctl status systemd-networkd.service
● systemd-networkd.service - Network Service
     Loaded: loaded (/usr/lib/systemd/system/systemd-networkd.service; disabled; vendor preset: disabled)
     Active: active (running) since Sat 2020-09-26 13:57:09 UTC; 7min ago
TriggeredBy: ● systemd-networkd.socket
       Docs: man:systemd-networkd.service(8)
   Main PID: 729 (systemd-network)
     Status: "Processing requests..."
      Tasks: 1 (limit: 1119)
     Memory: 2.4M
     CGroup: /system.slice/systemd-networkd.service
             └─729 /usr/lib/systemd/systemd-networkd

Sep 26 13:57:09 localhost systemd[1]: Starting Network Service...
Sep 26 13:57:09 localhost systemd-networkd[729]: Enumeration completed
Sep 26 13:57:09 localhost systemd[1]: Started Network Service.
Sep 26 13:57:09 localhost systemd-networkd[729]: eth0: IPv6 successfully enabled
Sep 26 13:57:09 localhost systemd-networkd[729]: eth0: Link UP
Sep 26 13:57:09 localhost systemd-networkd[729]: eth0: Gained carrier
Sep 26 13:57:09 localhost systemd-networkd[729]: eth0: DHCPv4 address 10.0.2.15/24 via 10.0.2.2
Sep 26 13:57:09 localhost systemd-networkd[729]: eth0: Unable to shorten overlong DHCP hostname 'flatcar_production_qemu-2632-0-0', ignoring: Numerical argum>
Sep 26 13:57:11 localhost systemd-networkd[729]: eth0: Gained IPv6LL
```

pothos: Have you tried pulling the target in before with `start network-online.target` or `sockets.target`?

pothos: Anyway thanks, also for making the situation much more clear now than it was before.

PR #613, opened 2020-09-25
sys-kernel/coreos-firmware: Disable savedconfig flag and file creation

pothos:
The savedconfig feature reads and, if not set, generates a file under
    /etc/portage/savedconfig/ to source a build configuration. We probably
    don't want this and specially not on the final image, therefore,
    disable reading and also don't write the file to the final image.

**Note:** Pick for 2605, 2632, and 2643

# How to use

Build an image and check if `/etc/portage` exists which it shouldn't.

# Testing done

Built an image and saw that the file is gone.

### DIFF
--- a/sys-kernel/coreos-firmware/coreos-firmware-99999999.ebuild
+++ b/sys-kernel/coreos-firmware/coreos-firmware-99999999.ebuild
@@ -166,9 +166,10 @@ src_prepare() {
 }
 
 src_install() {
-	if use !savedconfig; then
-		save_config ${PN}.conf
-	fi
+	# Flatcar: Don't save the firmware config to /etc/portage/savedconfig/
+	# if use !savedconfig; then
+	# 	save_config ${PN}.conf
+	# fi
 	rm ${PN}.conf || die
 	insinto /lib/firmware/
 	doins -r *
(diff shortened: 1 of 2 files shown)


sys-kernel/coreos-firmware/coreos-firmware-99999999.ebuild
```diff
@@ -166,9 +166,10 @@ src_prepare() {
 }
 
 src_install() {
-	if use !savedconfig; then
-		save_config ${PN}.conf
-	fi
+	# Flatcar: Don't save the firmware config to /etc/portage/savedconfig/
```
marga-kinvolk, line 169: Why do we need this divergence in the ebuild file? Shouldn't the USE flag be enough?

pothos: The use flag determines whether a saved config is used for the build. However, saving is done in the other case to have a base config to start editing from.

pothos: Corrected the other comment and updated the commit message

marga-kinvolk: Thanks, I understand now.

marga-kinvolk (approved):
LGTM

PR #614, opened 2020-09-25
Update srctree path to correctly populate Makefile

marga-kinvolk:
Update srctree path to correctly populate the Makefile for sandbox
environments. The patch is to adjusted for 5.x kernels

# How to use / Testing done

Sayan ran a Jenkins job with this change and after running that, the image generated has this file:
```
core@localhost ~ $ cat /usr/lib64/modules/5.8.10-flatcar/build/Makefile 
# Automatically generated by ../source/scripts/mkmakefile: don't edit
include ../source/Makefile
```

Which is correct as it no longer includes the sandbox environment.

This change needs to be cherry-picked to flatcar-2605 and flatcar-2632

### DIFF
--- a/sys-kernel/coreos-sources/files/5.8/z0001-kbuild-derive-relative-path-for-srctree-from-CURDIR.patch
+++ b/sys-kernel/coreos-sources/files/5.8/z0001-kbuild-derive-relative-path-for-srctree-from-CURDIR.patch
@@ -15,15 +15,17 @@ diff --git a/Makefile b/Makefile
 index 36eab48d1d4a..0514c0fa114d 100644
 --- a/Makefile
 +++ b/Makefile
-@@ -240,7 +240,7 @@ else
+@@ -230,9 +230,7 @@ else
+ 	building_out_of_srctree := 1
  endif
- 
- ifneq ($(KBUILD_ABS_SRCTREE),)
+
+-ifneq ($(KBUILD_ABS_SRCTREE),)
 -srctree := $(abs_srctree)
+-endif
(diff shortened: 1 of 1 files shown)


marga-kinvolk:
Cherry picked to flatcar-2632 and flatcar-2605

marga-kinvolk:
We're having troubles with the developer container because of this change.

This is the error that we get:
```
realpath: '': No such file or directory
```

This comes from this line:
```
srctree := $(shell realpath --relative-to=$(KBUILD_OUTPUT) $(abs_srctree))
```

Unfortunately we don't know which of the parameters is empty:
```
marga@calliope:/tmp$ realpath --relative-to=. ''
realpath: '': No such file or directory
marga@calliope:/tmp$ realpath --relative-to='' .
realpath: '': No such file or directory
marga@calliope:/tmp$ realpath --relative-to='' ''
realpath: '': No such file or directory
marga@calliope:/tmp$ realpath --relative-to=. .
.
marga@calliope:/tmp$ realpath .
/tmp
```

We need to somehow ensure that both variables have content.

marga-kinvolk:
Ok, it looks like it's `KBUILD_OUTPUT` that's empty, the other one should actually have content... I guess the patch should actually have a conditional... Something like this, maybe?

```
ifneq ($(KBUILD_OUTPUT),)
        srctree := $(shell realpath --relative-to=$(KBUILD_OUTPUT) $(abs_srctree))
else
        srctree := $(abs_srctree)
endif
```

@sayanchowdhury what do you think?

PR #620, opened 2020-09-28
Adjust the `srctree` based on the presence of KBUILD_OUTPUT

sayanchowdhury:

# Adjust the `srctree` based on the presence of KBUILD_OUTPUT

# How to use
``` bash
./build_image container
```

# Testing done

@marga-kinvolk tested it, and I did a local compile, need to test on Jenkins

Related #614


marga-kinvolk (approved):
LGTM.

The Jenkins build is still running, but it's already past the part where it builds the kernel, and given how late it is, I'd like to merge this and trigger new builds for the releases.

PR #622, opened 2020-09-30
eclass: exclude unnecessary architectures from DTC in Kernel

dongsupark:
DTC (Device Tree Compiler) source tree in Flatcar Kernel modules unnecessarily takes too much space, especially the `include-prefixes` directory.

```
$ sudo du -a /usr/lib64/modules/$(uname -r)/source/ | sort -n -r | head -n5
130100  /usr/lib64/modules/5.8.11-flatcar/source/
69180   /usr/lib64/modules/5.8.11-flatcar/source/include
56324   /usr/lib64/modules/5.8.11-flatcar/source/scripts
51384   /usr/lib64/modules/5.8.11-flatcar/source/scripts/dtc
50728   /usr/lib64/modules/5.8.11-flatcar/source/scripts/dtc/include-prefixes
$ sudo ls /usr/lib64/modules/$(uname -r)/source/scripts/dtc/include-prefixes/
arc  arm  arm64  c6x  dt-bindings  h8300  microblaze  mips  nios2 openrisc  powerpc  sh  xtensa
```

Most of them are for architectures that are not supported by Flatcar, so we can remove them from the production image.

OTOH, as `dt-bindings` looks more like an architecture-independent one, for now we keep it.

Before:

```
$ du -s /usr/lib64/modules/$(uname -r)/source/scripts/dtc/
51384   /usr/lib64/modules/5.8.11-flatcar/source/scripts/dtc/
$ du -s /usr/lib64/modules/
250308  /usr/lib64/modules/
$ df /usr
Filesystem      1K-blocks   Used Available Use% Mounted on
/dev/mapper/usr   1007760 934152     21592  98% /usr
```

After:

```
$ du -s /usr/lib64/modules/$(uname -r)/source/scripts/dtc/
6632    /usr/lib64/modules/5.8.11-flatcar/source/scripts/dtc/
$ du -s /usr/lib64/modules/
205144  /usr/lib64/modules/
$ df /usr
Filesystem      1K-blocks   Used Available Use% Mounted on
/dev/mapper/usr   1007760 907628     48116  95% /usr
```

## How to use

```
emerge-amd64-usr coreos-modules
```

## Testing done

Tested on both amd64 and arm64.

### DIFF
--- a/eclass/coreos-kernel.eclass
+++ b/eclass/coreos-kernel.eclass
@@ -187,16 +187,55 @@ shred_keys() {
 install_build_source() {
 	local kernel_arch=$(tc-arch-kernel)
 
+	# NOTE: We have to get ${archabspaths} before removing symlinks under
+	# /usr/lib/modules. However, do not exclude "dt-bindings" for now,
+	# as it looks architecture-independent.
+	local archabspaths=($(ls -1d ${D}/usr/lib/modules/${KV_FULL}/source/scripts/dtc/include-prefixes/* \
+		| grep -v dt-bindings ))
+
 	# remove the broken symlinks referencing $ROOT
 	rm "${D}/usr/lib/modules/${KV_FULL}"/{build,source} || die
 
+	# Compose list of architectures to be excluded from the kernel modules
+	# tree in the final image. It is an array to be used as a pattern for
+	# grep command below at the end of "find source/scripts" command for
+	# fetching kernel modules list, e.g.:
+	#   find source/scripts -follow -print \
+	#   | grep -E -v -w "include-prefixes/arc|include-prefixes/xtensa"
+	declare -a excarchlist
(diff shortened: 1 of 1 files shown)


pothos:
Are there any symlinks from arm64 to arm which won't work now?

I think it would be good to run a CI test on all platforms (clearing the format lists in the manifest job).

dongsupark:
> Are there any symlinks from arm64 to arm which won't work now?

AFAIK, Kernel always distinguishes arm64 from arm.
From a quick look at the Kernel tree, I think we don't have to worry about that issue.

> I think it would be good to run a CI test on all platforms (clearing the format lists in the manifest job).

Today I ran the CI tests again on all platforms and all providers.
No regression, no unexpected issue found.

pothos (approved):

PR #637, opened 2020-10-09
Revert "sys-apps/systemd: Drop the resolv.conf workaround"

pothos:
This reverts commit c414b38c7c56dafb05a86040443c634763527f05.
The real DNS server IP addresses should be in /etc/resolve.conf and not
just 127.0.0.53 because all cases that bind-mount /etc/resolve.conf
into a new network namespace can't reach the loopback interface that
resolved is listening on.

# How to use

# Testing done

None


krnowak:
I dropped this workaround, because the commit that added it pointed to the systemd issue https://github.com/systemd/systemd/issues/3826, which is supposedly fixed upstream already.

pothos:
Yes, but I think there are many cases where the `resolve.conf` is expected to not contain a loopback IP address because the file is bind-mounted to another network namespace.

sayanchowdhury (approved):

PR #640, opened 2020-10-13
dev-lang: Upgrade Rust 1.46.0 to 1.47.0

dongsupark:
Update `dev-lang/rust` version to 1.47.0.
Remove unnecessary patches, add new patches for rust 1.47.0.

Fix an issue of wrong stage0 tarball names. Rust stage0 tarballs should not be based on a patchlevel release like `1.45.1`. It might work in case of the previous version 1.45.1, which already exists. However, it will not work in case of x.y.1 is missing.
So the build rust 1.47.0, should pull tarballs for rust 1.46.0, instead of 1.46.1, which does not exist.

It should be merged together with https://github.com/flatcar-linux/portage-stable/pull/114.

## Testing done

CI passed


dongsupark:
This PR is a blocker for every further Github Actions for Rust, because `dev-lang/rust` in the current main branch still has a bug in its semver patchlevel.
It was fixed via https://github.com/kinvolk/coreos-overlay/pull/640/commits/6c393d3ba90421d2402e1010a23a08ad8a14f5f1 , which was not merged yet.
That's why Github Actions failed like [that](https://github.com/kinvolk/coreos-overlay/runs/1446368441?check_suite_focus=true), even if there was a new Rust release [1.48.0](https://github.com/rust-lang/rust/releases/tag/1.48.0).

I will soon merge this PR.

PR #649, opened 2020-10-15
sys-kernel: remove blank kernel module nf-conntrack-ipv4

dongsupark:
So far Flatcar has kept a third-party patch to add a blank kernel module `nf-conntrack-ipv4.ko` to avoid regression around Kubernetes.
The issue was that kube-proxy with ipvs started using `nf-conntrack.ko`, which does not exist in Kernel < 4.19.
The patch was originally added by https://github.com/flatcar-linux/coreos-overlay/commit/a24dbb6cb639d7e87c38502558336bdfe16f55b0.

However, Kubernetes 1.13 or newer already deals with the issue.
It automatically loads a different Kernel module according to Kernel versions:
`nf-conntrack-ipv4` for Kernel < 4.19, and `nf-conntrack` for Kernel >= 4.19.
See https://github.com/kubernetes/kubernetes/commit/4b90559369261f1892f81ecdd3cd808e2fe39d22 .

We can simply remove the Kernel module, as since then all production systems have updated Kubernetes to the newer versions than 1.13.

## How to use

```
./build_packages
```

## Testing done

CI passed.


pothos:
Since the kola Kubernetes tests run only on GCE I [rebuild the vm-matrix job](http://localhost:9091/job/os/job/board/job/vm-matrix/1245/console) with that image format.

sys-kernel/coreos-kernel/coreos-kernel-5.8.14-r1.ebuild
```diff
@@ -2,7 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=5
-COREOS_SOURCE_REVISION=""
+COREOS_SOURCE_REVISION="-r1"
```
pothos, line 5: I think it's ok to leave the package version at the same revision that it doesn't conflict with https://github.com/flatcar-linux/coreos-overlay/pull/647

pothos (approved):

sys-kernel/coreos-kernel/coreos-kernel-5.8.14-r1.ebuild
dongsupark: Removed the revision changes.

pothos: It anyway conflicts, no idea why, but at least we tried :smile:

dongsupark: Interesting. Manual rebase work just well.
I simply pushed a new rebase to https://github.com/flatcar-linux/coreos-overlay/pull/647

PR #655, opened 2020-10-20
dev-lang/go: update Go to 1.12.17, remove Go 1.10

dongsupark:
Make Go 1.12.x runtime up-to-date with its latest release 1.12.17, to fix numerous bugs.
See also https://golang.org/doc/devel/release.html#go1.12.minor .

Since Go 1.10 has not been used by packages, we can completely remove Go 1.10.

Remove `dev-lang/go:1.10` from the SDK dependencies list.
Instead add `dev-lang/go:1.15` to the list.

## How to use

```
./build_packages
```

## Testing done

CI passed


vbatts:
does this address https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-24553?

dongsupark:
> does this address https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-24553?

I doubt that.
The [original advisory](https://www.redteam-pentesting.de/en/advisories/rt-sa-2020-004/-inconsistent-behavior-of-gos-cgi-and-fastcgi-transport-may-lead-to-cross-site-scripting) simply says `Affected Versions: <= 1.14.7, 1.15`, so it means they only tested with 1.14.x and 1.15.x. Not known if 1.12 is affected as well.

Individual milestone lists of Go 1.12.x like https://github.com/golang/go/issues?q=milestone%3AGo1.12.17+label%3ACherryPickApproved also do not have security issues.

krnowak (approved):
Looks good.

Question though: are there projects that can't be built with go 1.15? We seem to go back with go as far as 1.6…

dongsupark:
> Question though: are there projects that can't be built with go 1.15? We seem to go back with go as far as 1.6…

`app-emulation/runc` 1.0.0_rc2 (!) needed by Docker 1.12 still uses Go 1.6.
`app-emulation/docker-runc` 1.0.0_rc2, `app-emulation/docker-proxy` 0.8.0_p20161019, `app-emulation/docker` 1.12.6-r8, `app-emulation/docker` 17.03.2, `app-emulation/containerd` 0.2.x still use Go 1.7.
`app-emulation/docker-proxy` 0.8.0_p20170917 still uses Go 1.8.
Those are blockers that prevent us from cleaning up Go 1.[6-8].

We would try to simply build every package above with Go 1.15. It might work.
Though it could also uncover hidden issues. ;-)

Best path forwards would be first to make Docker 1.12 deprecated, so we could remove Go 1.6, etc.

vbatts:
yes, remove docker-1.12 and yes to trying to build everything with a compiler that upstream is still making releases for (even go1.12 would be a decent goal)

PR #657, opened 2020-10-22
profiles: gdbus-codegen only for python 3.6

dongsupark:
To be able to update `dev-util/gdbus-codegen` to 2.64.5, we need to
specify a single target python3.6 for gdbus-codegen.

Without it, emerge gdbus-codegen 2.64.5 is not possible, because
it recognizes there are multiple python single targets.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/115 .

## Testing done

TBD


dongsupark:
See https://github.com/flatcar-linux/coreos-overlay/pull/658 instead.

PR #658, opened 2020-10-22
profiles: gdbus-codegen only for python 3.6

dongsupark:
To be able to update `dev-util/gdbus-codegen` to 2.64.5, we need to specify a single target python3.6 for gdbus-codegen.

Without it, it is not possible to emerge gdbus-codegen, because it thinks there are multiple python single targets for the package.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/118.

## Testing done

CI passed


dongsupark:
CI passed

vbatts:
LGTM

PR #659, opened 2020-10-26
Update systemd (sync, use gentoo workflow, add missing patches)

krnowak:
This PR does three things:

- It synces with the latest version of the ebuild file from Gentoo. This adds an enabled-by-default USE flag, which pulls in a dep on `app-arch/zstd`, which I imported in portage-stable (PR: https://github.com/flatcar-linux/portage-stable/pull/120). So the PR in portage-stable needs to be merged first.
- It replaces the CROS_WORKON workflow with the Gentoo workflow. Please see the commit messages for rationale. This means that the patches are now part of the coreos-overlay repo, instead of the systemd repo in our org. This also means that the diff between our ebuild and Gentoo ebuild is slightly lower. It also means that the ebuild file needs to be renamed if we want to use a patch release.
- It adds the patches I missed when updating systemd from v245 to v246. This follows what I did in https://github.com/flatcar-linux/systemd/pull/13, so if this PR is merged, then https://github.com/flatcar-linux/systemd/pull/13 can be closed.

Closes flatcar-linux/systemd#13

Test build: http://localhost:9091/job/os/job/manifest/1300/


krnowak:
Test build passed, ready for review.

krnowak:
The required PR in portage-stable is now merged.

t-lo:
PR looks good, thanks! Let's wait for @sayanchowdhury to have a look, for visibility.

pothos (approved):

PR #682, opened 2020-11-13
sys-block/open-iscsi: Add the version 2.1.2

sayanchowdhury:

# sys-block/open-iscsi: Add the version 2.1.2

# How to use

```bash
emerge-amd64-usr sys-block/open-iscsi
```

# Testing done
Done local testing and successful Jenkins build

To be merged with kinvolk/portage-stable#123
Fixes kinvolk/flatcar#219


dongsupark (approved):
Build works well.
I thought we needed to apply third-party patches again on top of 2.1, but it was not the case.
It looks actually pretty straightforwards to update to 2.1.
In the future we should probably move the open-iscsi ebuilds to portage-stable, but we can do it later. ;-)
Thanks!

PR #685, opened 2020-11-13
.github: fix mantle URL for downloading cork

dongsupark:
The repo `github.com/flatcar-linux/mantle` has been moved to `github.com/kinvolk/mantle`.
However, Github Actions still fetch cork binaries from the original URL, by running `curl` without `-L`.
So the request does not get redirected to the new URL.
As a result, `CORK_VERSION` becomes null.

Fix it by replacing `flatcar-linux` with `kinvolk`, as well as adding `-L` to the curl command, just in case.

### DIFF
--- a/.github/workflows/setup-flatcar-sdk.sh
+++ b/.github/workflows/setup-flatcar-sdk.sh
@@ -2,9 +2,9 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


vbatts (approved):
LGTM

PR #686, opened 2020-11-16
eclass: remove also other optimization flags with -Wl,-O*

dongsupark:
We need to filter not only `-Wl,-O1`, but also other flags like `-Wl,-O2`, `-Wl,-Og`, `-Wl,-Os`, etc.
Otherwise, SDK build would fail, for example, as its default `$LDFLAGS` includes `-Wl,-O2`.

We need to manually strip only the optimization element of comma-separated flags, e.g. from `-Wl,-O1,-s` to `-Wl,-s`.
To support multiple characters that can follow `-O`, e.g. `-Ofast`, we should use regexp like `[[:alnum:]]*`.

## How to use

```
sudo emerge dev-lang/go:1.15
sudo FORCE_STAGES="stage4" ./bootstrap_sdk
```

## Testing done

SDK build passed in CI.

### DIFF
--- a/eclass/coreos-go-utils.eclass
+++ b/eclass/coreos-go-utils.eclass
@@ -87,7 +87,14 @@ go_export() {
 	# Remove certain flags from $LDFLAGS to fix validation errors in
 	# Go >= 1.15.5 like:
 	#   go build runtime/cgo: invalid flag in go:cgo_ldflag: -Wl,-O1
-	filter-ldflags "-Wl,-O1"
+	#
+	# Note, we need to manually strip only the optimization element of
+	# comma-separated flags, e.g. from `-Wl,-O1,-s` to `-Wl,-s`.
+	# To support multiple characters that can follow `-O`, e.g. `-Ofast`,
+	# we should use regexp like `[[:alnum:]]*`.
+	# Work-around for https://github.com/golang/go/pull/42631.
+	export LDFLAGS="$(echo "$LDFLAGS" | sed 's/,-O[[:alnum:]]*//g')"
+	filter-ldflags "-Wl"
 
 	export CC=$(tc-getCC)
 	export CXX=$(tc-getCXX)


pothos:
This cuts away everything in `LDFLAGS` after the `-Wl,O` unless a space is used to separate it. Is this intended?

pothos:
Ah, looks like `-Wl,O` isn't valid anyway as it should be `-Wl,-O`.

pothos:
For something like `gcc -Wl,-lm,-S input.c` GCC preserves all linker flags. That means the glob would cut away too much.

eclass/coreos-go-utils.eclass
```diff
@@ -87,7 +87,7 @@ go_export() {
 	# Remove certain flags from $LDFLAGS to fix validation errors in
 	# Go >= 1.15.5 like:
 	#   go build runtime/cgo: invalid flag in go:cgo_ldflag: -Wl,-O1
-	filter-ldflags "-Wl,-O1"
+	filter-ldflags "-Wl,-O*"
```
pothos, line 90: Maybe something like this could work to filter all `-O?` away while still preserving any other parameters.
```suggestion
	export LDFLAGS="$(echo "$LDFLAGS" | sed 's/,-O.//g')"
	filter-ldflags "-Wl"
```

pothos: Actually the Go compiler shouldn't reject `-Wl,-O1` but does so because of a typo https://github.com/golang/go/blob/master/src/cmd/go/internal/work/security.go#L182 (capital `L`). We should fix this upstream to be able to remove this filtering.

dongsupark: Updated as your suggestion, but had to slightly change it like that:
```
export LDFLAGS="$(echo "$LDFLAGS" | sed 's/,-O[[:alnum:]]*//g')"
filter-ldflags "-Wl"
```
Will soon run another SDK build.

pothos: Filed https://github.com/golang/go/pull/42631 but we can work around it for now

eclass/coreos-go-utils.eclass
```diff
@@ -87,7 +87,13 @@ go_export() {
 	# Remove certain flags from $LDFLAGS to fix validation errors in
 	# Go >= 1.15.5 like:
 	#   go build runtime/cgo: invalid flag in go:cgo_ldflag: -Wl,-O1
-	filter-ldflags "-Wl,-O1"
+	#
+	# Note, we need to manually strip only the optimization element of
+	# comma-separated flags, e.g. from `-Wl,-O1,-s` to `-Wl,-s`.
+	# To support multiple characters that can follow `-O`, e.g. `-Ofast`,
+	# we should use regexp like `[[:alnum:]]*`.
```
pothos, line 94: Maybe add a comment that this is a work-around for https://github.com/golang/go/pull/42631

pothos (approved):

eclass/coreos-go-utils.eclass
dongsupark: Added the comment

eclass/coreos-go-utils.eclass
dongsupark: SDK build passed

dongsupark:
cherry-picked to flatcar-2643

PR #690, opened 2020-11-17
profiles: make SDK include dev-util/dwarves

dongsupark:
To build Kernel with `CONFIG_DEBUG_INFO_BTF`, we need to make `pahole` in dwarves included in the Flatcar SDK.

To do that, we need to make it accept `~amd64` keywords for dwarves and binutils.
Also enable USE flag `python_single_target_python3_6` for dwarves.

See also https://github.com/kinvolk/Flatcar/issues/225 .

This PR should be merged together with https://github.com/kinvolk/portage-stable/pull/126.

## How to use

```
sudo emerge dev-util/dwarves
```

## Testing done

CI passed http://localhost:9091/job/os/job/manifest/1484/cldsv/


t-lo (commented):
LGTM, thank you!

t-lo (approved):

PR #691, opened 2020-11-17
sys-kernel: enable BPF_JIT, POWER_SUPPLY, overlayfs options

dongsupark:
Enable the following kernel configs.

`CONFIG_BPF_JIT_ALWAYS_ON` enables BPF JIT and removes BPF interpreter to avoid speculative execution of BPF instructions by the interpreter.

`CONFIG_POWER_SUPPLY` enables power supply class used to represent battery, UPS, AC or DC power supply properties to user-space.
It defines core set of attributes, which should be applicable to most power supplies out there.

`CONFIG_OVERLAY_FS_METACOPY`, metadata only copy up feature in overlayfs. When turned on, overlayfs will only copy up metadata when a metadata specific operation like chown/chmod is performed.
Full file will be copied up later when file is opened for WRITE operation. More or less like delayed data copy up operation.

`CONFIG_OVERLAY_FS_REDIRECT_DIR`, which is equivalent to "redirect_dir=on" in the kernel command-line.
When turned on, overlayfs will copy up directory first, before the actual contents.

Fixes https://github.com/kinvolk/Flatcar/issues/185
Fixes https://github.com/kinvolk/Flatcar/issues/215
Fixes https://github.com/kinvolk/Flatcar/issues/170

## How to use

```
./build_packages
```

## Testing done

CI passed http://localhost:9091/job/os/job/manifest/1485/cldsv/


dongsupark:
CI passed

sayanchowdhury (approved):
LGTM

PR #693, opened 2020-11-18
dev-util/profiles/sys-devel/sys-libs: update to gcc-9.3.0

t-lo:
This PR includes the necessary changes to upgrade the SDK compiler to gcc-9.3.0.

The PR depends on a [PR in the scripts repo](https://github.com/kinvolk/flatcar-scripts/pull/102) which fixes a symlink issue with liblto, as well as the respective update-to-gcc-9.3.0 [PR in the portage-stable](https://github.com/kinvolk/portage-stable/pull/127) repo.

The changes include:

update dev-util/strace to strace-5.9.ebuild
update sys-devel/gdb to gdb-9.2.ebuild

sys-devel/gdb/gdb-9.2.ebuild: use EAPI6 to work around BDEPEND emerge bug
update sys-libs/nss-usrfiles to nss-usrfiles-2.30.ebuild to support glibc > 2.29
update sys-kernel/README.md to call out need for updating kernel-headers, perf
add sys-libs/glibc/README.md outlining our changes to the glibc recipe
update profiles/coreos/base/package.accept_keywords to include new toolchain

The change also adds a README to
sys-libs/glibc/README.md
and it improves on a README in
sys-kernel/README -> sys-kernel/README.md


pothos (changes requested):
Please split the commit up into the original file import from Gentoo and the downstream changes in a follow-up commit. Otherwise the next person may miss what needs to be ported when updating to a newer version.

t-lo:
@pothos I've split up the PR into two commits.

sys-libs/nss-usrfiles/nss-usrfiles-9999.ebuild
```diff
@@ -9,7 +9,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="42bec47544ad80d3e39342b11ea33da05ff9133d" # v2.23.0
+	CROS_WORKON_COMMIT="9078c543ba7d2bc5011737675b3dddb882673ce7" # flatcar-master
```
pothos, line 12: This commit ID is from `master`, not `flatcar-master`. Both branches are the same, though…

pothos (approved):

sys-libs/nss-usrfiles/nss-usrfiles-9999.ebuild
t-lo: I'll fix this and use the one from flatcar-master.

t-lo:
Adjusted the CROS_WORKON_COMMIT in `sys-libs/nss-usrfiles/nss-usrfiles-9999.ebuild` to point to flatcar-master.

PR #702, opened 2020-11-24
Move cryptsetup back to portage-stable

pothos:
The LTS branch contains modifications from flatcar-2605 and needs to be diffed easily, thus keep the package in portage-stable even if we make a change.


vbatts:
Did the `--enable-fips` flag get preserved with this move back

pothos:
Yes, here https://github.com/kinvolk/portage-stable/pull/128/files#diff-994222f1c1827d10553338b1b6400497c3d912569f66e3df1c911dfa4ac05a62R88

PR #705, opened 2020-11-25
kernel: Enable HOTPLUG_PCI_ACPI for arm64

lht:
# Enable kernel feature HOTPLUG_PCI_ACPI for arm64

Attaching additional volumes to a running EC2 arm64 based instance doesn't work. The disks won't be detected. To reproduce the issue,

* Launch a EC2 arm64 instance with Flatcar-alpha-2697.0.0
* run `lsblk` and note down the block device name
* attach an additional volume to the instance.
* run `lsblk` again. The newly attached volume is not shown.

# Testing done
With AMI built with the patch, newly attached volumes are shown by `lsblk`.


pothos (commented):
Thanks, great, I will make a test, too.

PR #711, opened 2020-11-26
x11-drivers/nvidia-drivers: Fix the dev container download URL

sayanchowdhury:
# x11-drivers/nvidia-drivers: Fix the dev container download URL

The kola tests fails to download during the release because the
artifacts of the release has not been pushed to the website yet.
This adds the logic to check if the URL is 200, then only download
or else fallback to the GCS bucket url.

This commit also changes a bug with the check to see if nvidia
is installed or required.

# How to use

```
./build_packages
./build_image
./image_to_vm --format=azure_pro
```
Boot and check the output of the command of `nvidia-smi`

# Testing done

Tested the script on an Azure box both with GPU and without.
Image matrix rebuild: http://localhost:9091/job/os/job/board/job/image-matrix/1641/ 

Note: this needs to be picked to `flatcar-2643` branch

### DIFF
--- a/x11-drivers/nvidia-drivers/files/bin/setup-nvidia
+++ b/x11-drivers/nvidia-drivers/files/bin/setup-nvidia
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -euo pipefail
+
 . /usr/share/flatcar/release
 . /usr/share/flatcar/update.conf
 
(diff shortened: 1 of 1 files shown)


x11-drivers/nvidia-drivers/files/bin/setup-nvidia
```diff
@@ -25,8 +26,12 @@ function download_flatcar_developer_container() {
...
       FLATCAR_DEVELOPER_CONTAINER_URL="https://storage.googleapis.com/flatcar-jenkins/developer/developer/boards/${FLATCAR_RELEASE_BOARD}/${FLATCAR_RELEASE_VERSION}/${FLATCAR_DEVELOPER_CONTAINER}.bz2"
-    else
-      FLATCAR_DEVELOPER_CONTAINER_URL="https://${GROUP}.release.flatcar-linux.net/${FLATCAR_RELEASE_BOARD}/${FLATCAR_RELEASE_VERSION}/${FLATCAR_DEVELOPER_CONTAINER}.bz2"
+    else if [[ "$(curl -I -s -o /dev/null -w "%{http_code}" "https://${GROUP}.release.flatcar-linux.net/${FLATCAR_RELEASE_BOARD}/${FLATCAR_RELEASE_VERSION}/${FLATCAR_DEVELOPER_CONTAINER}.bz2")" == 200 ]]
+      then
+        FLATCAR_DEVELOPER_CONTAINER_URL="https://${GROUP}.release.flatcar-linux.net/${FLATCAR_RELEASE_BOARD}/${FLATCAR_RELEASE_VERSION}/${FLATCAR_DEVELOPER_CONTAINER}.bz2"
+      else
+        FLATCAR_DEVELOPER_CONTAINER_URL="https://storage.googleapis.com/flatcar-jenkins/${GROUP}/boards/${FLATCAR_RELEASE_BOARD}/${FLATCAR_RELEASE_VERSION}/${FLATCAR_DEVELOPER_CONTAINER}.bz2"
+      fi
     fi
```
pothos, line 35: When at it it would be good to add `--fail` to curl in the lines below so that the error doesn't bubble up from bzip failing.

pothos (approved):

x11-drivers/nvidia-drivers/files/bin/setup-nvidia
sayanchowdhury: Sure.

PR #712, opened 2020-11-26
Enable BCM7XXX_PHY as a module on arm64_defconfig-5.9

shift:
# Enables Raspberry Pi 4 PHY

The following 1 line change enables the kernel module to be build enabling the Raspberry Pi 4 PHY enabling the on-board NIC.

# How to use

Build it and boot it :)

# Testing done

Validated the config change against known working 5.8.y kernels on the Pi4.


pothos (changes requested):
Thank for for the Pull Request, can you please base your branch on `main` and then point the PR to `main`?
It would be good to take the current PR title for the commit, too.

shift:
I've updated the PR, it can be found here https://github.com/kinvolk/coreos-overlay/pull/716.

PR #719, opened 2020-12-01
.github: use create-pull-request@v3 to fix set-env issues

dongsupark:
Github Actions for Rust started failing with following errors:

```
Error: Unable to process command '::set-env name=PULL_REQUEST_NUMBER::718' successfully.
Error: The `set-env` command is disabled. Please upgrade to using
Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`.
For more information see:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

It happens because we have used `peter-evans/create-pull-request@v2`, which did not have a bug fix for the set-env issue.
The bug was fixed in create-pull-request [v3.4.1](https://github.com/peter-evans/create-pull-request/releases/tag/v3.4.1).
So we just need to update the version to `v3`, which already includes v3.4.1.

### DIFF
--- a/.github/workflows/containerd-releases-main.yml
+++ b/.github/workflows/containerd-releases-main.yml
@@ -33,7 +33,7 @@ jobs:
           PATH: ${{ steps.setup-flatcar-sdk.outputs.path }}
         run: .github/workflows/containerd-apply-patch.sh
       - name: Create pull request for main
-        uses: peter-evans/create-pull-request@v2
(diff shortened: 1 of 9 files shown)


marga-kinvolk (approved):
LGTM

PR #724, opened 2020-12-03
coreos-base/[oem-]update-vendortools: vendor tools update service

t-lo:
This PR requires https://github.com/kinvolk/flatcar-scripts/pull/105.

Packages added:
- `coreos-base/update-vendortools` - the main service + script (`/usr/bin/check-update-vendortools.sh`)
- `coreos-base/oem-update-vendortools` - "activation" package to be added as a dependency to all oem-* packages which want to make use of the update mechanism.

`coreos-base/update-vendortools` introduces a simple `check-update-vendortools` systemd service to update the vendor tools shipped on the OEM partition. A dependency to this service has been added to coreos-base/coreos/coreos-0.0.1.ebuild. By default, the service does nothing.

In order for `coreos-base/oem-*` packages to make use of the update mechanism, all the oem packages need to do is to `DEPEND` / `RDEPEND` on `coreos-base/oem-update-vendortools`. This will trigger a placeholder file `/usr/share/oem/flatcar-oem-release` to be added to the VM image at the `image_to_vm.sh` OS image build step.

Presence of `/usr/share/oem/flatcar-oem-release` will:
- cause `image_to_vm.sh` to populate the placeholder file with the actual version being built as well as the vendor-specific name of the update tarball.
- cause image_to_vm.sh to generate a vendor tools partition update tarball, and add the update tarbal to the release files. This will cause the update tarball to be published.
- make the check-update-vendortools service download the update tarball via a update_engine hook, and install it on next boot. A rollback tarball is created which is applied by the same service should the system come up with the old flatcar version in a succeeding boot.

# Testing done
- Added `DEPEND`+`REDPEND` `coreos-base/oem-update-vendortools`  to oem-qemu for testing, and built an image
- verified that `flatcar_production_qemu_oem-update.tz` is generated and contains the correct release version (*2705.0.0* in my case)
- booted the image; checked journal for errors - no errors.
- manually edited `flatcar_production_qemu_oem-update.tz` and changed `FLATCAR_OEM_VERSION` in `flatcar-oem-release` to *2706.0.0*
- started a local web server (`python -m http.server` FTW!)
- `/usr/bin/check-update-vendortools.sh -u http://<host-ip>/ -v 2706.0.0` in the qemu image succeeded in downloading and staging the tarball
- changed `/etc/lsb-release` `DISTRIB_RELEASE` to *2706.0.0* and rebooted
- OEM partition contents were successfully replaced by tarball contents, roll-back OEM tarball was generated
- changed `/etc/lsb-release` `DISTRIB_RELEASE` back to *2706.0.0* and rebooted
- OEM partition was successfully rolled back to *2705.0.0*

# TODO
- update server to support distribution mechanism / URLs for receiving vendor updates
- hook `/usr/bin/check-update-vendortools.sh` into update_engine to download / stage OEM update tarballs

### DIFF
--- a/coreos-base/coreos/coreos-0.0.1.ebuild
+++ b/coreos-base/coreos/coreos-0.0.1.ebuild
@@ -112,6 +112,7 @@ RDEPEND="${RDEPEND}
 	coreos-base/coreos-init
 	coreos-base/update-ssh-keys
 	coreos-base/update_engine
+	coreos-base/update-vendortools
 	dev-db/etcdctl
 	dev-util/strace
 	dev-vcs/git
--- a/coreos-base/oem-update-vendortools/files/flatcar-oem-release
+++ b/coreos-base/oem-update-vendortools/files/flatcar-oem-release
@@ -1,0 +1,1 @@
+# placeholder, replaced by image_to_vm.sh
--- a/coreos-base/oem-update-vendortools/oem-update-vendortools-0.1.ebuild
+++ b/coreos-base/oem-update-vendortools/oem-update-vendortools-0.1.ebuild
@@ -1,0 +1,25 @@
+# Copyright (c) 2020 Kinvolk GmbH. All rights reserved.
+# Distributed under the terms of the GNU General Public License v2
+
+# Vendor tools depending on this package will have vendor tools updates
+#  activated. If the file created below is present, image_to_vm.sh will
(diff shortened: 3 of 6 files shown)


t-lo:
Closing this proof of concept draft as we'll be going forward with [sysext for OEM updates](https://github.com/flatcar-linux/Flatcar/issues/60).

PR #727, opened 2020-12-04
app-editors/vim: do not create symlink vimdiff for minimal

dongsupark:
A symlink `vimdiff` should not be created, if the USE flag `minimal` is enabled.
Otherwise running `vimdiff` results in failure like that:

```
$ vimdiff aaa bbb
This Vim was not compiled with the diff feature.
```

See also https://github.com/gentoo/gentoo/pull/18496.
Fixes https://github.com/kinvolk/Flatcar/issues/221.

## How to use

```
emerge-amd64-usr app-editors/vim
```


marga-kinvolk (approved):
LGTM

sayanchowdhury (approved):

PR #728, opened 2020-12-04
dev-util/bsdiff: sync with Gentoo for integer signedness error

dongsupark:
Before applying Flatcar patches to bsdiff, sync with upstream Gentoo, so the ebuilds could make use of EAPI=7.
Also drop third-party patches, to be able to start from scratch.

Doing that we can fix [CVE-2014-9862](https://nvd.nist.gov/vuln/detail/CVE-2014-9862), integer signedness error in bspatch.c.
With the vulnerability, remote attackers to execute arbitrary code or cause a denial of service (heap-based buffer overflow) via a crafted patch file.
Since Gentoo already has the third-party patch, we can simply make use of it.

See also https://bugs.gentoo.org/701848 , https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=f4d7646f1d69 .

Apply existing Flatcar changes on top of vanilla Gentoo ebuilds.
Basically add arm64 keyword, and apply the sais patch.
It is based on the following commits:
* [4ee6aa895a02](https://github.com/kinvolk/coreos-overlay/commit/4ee6aa895a02) ("Add arm64 keywords")
* [60d47e7359d1](https://github.com/kinvolk/coreos-overlay/commit/60d47e7359d1) ("Change suffix sort to sais-lite")

## How to use

```
emerge-amd64-usr dev-util/bsdiff
```


marga-kinvolk (approved):
LGTM

PR #729, opened 2020-12-04
profiles: disable jpeg for qemu

dongsupark:
Qemu has enabled `jpeg` USE flag since the beginning, without any reason specified.
As a result, qemu pulls in unnecessary packages, `virtual/jpeg` as well as `media-libs/libjpeg-turbo`.
However, Flatcar runs qemu always with `-display none` option.
So the `jpeg` flag is not needed at all.

Simply remove `jpeg` USE flag from qemu.

This PR should be merged together with https://github.com/kinvolk/portage-stable/pull/133.

## How to use

```
sudo FORCE_STAGES="stage4" bootstrap_sdk
```

## Testing done

SDK build passed. http://localhost:9091/job/os/job/manifest/1594/cldsv/

### DIFF
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -15,7 +15,7 @@ dev-vcs/git		webdav curl bash-completion
 net-misc/curl		kerberos threads
 net-misc/iputils	arping tracepath traceroute
 sys-devel/gettext	-git
-app-emulation/qemu	aio caps curl jpeg ncurses png python threads uuid vhost-net virtfs vnc -xkb qemu_softmmu_targets_x86_64 qemu_softmmu_targets_aarch64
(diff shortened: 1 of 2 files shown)


marga-kinvolk:
Can this affect what happens when a user runs qemu (on their host) passing the Flatcar qemu image?

dongsupark:
This change has nothing to do with qemu used by end-users on their host.
Also Flatcar production images do not include qemu binaries at all.

Qemu binaries we deal with here are only included in Flatcar SDK.
Main use case is of course mantle/kola. 
You can see qemu command-line options used by kola [here](https://github.com/kinvolk/mantle/blob/fdee6416d26dd9994842450aad89c8de06f93958/platform/qemu.go#L277).

dongsupark:
SDK build in CI passed.

marga-kinvolk (approved):
LGTM

PR #732, opened 2020-12-08
sys-libs/pam: update to 1.5.1, fix auth issues

dongsupark:
We need to update `sys-libs/pam` to 1.5.1, mainly to address [CVE-2020-27780](https://nvd.nist.gov/vuln/detail/CVE-2020-27780): a flaw in the way it handles empty passwords for non-existing users. When the user doesn't exist PAM try to authenticate with root and in the case of an empty password it successfully authenticate.

This PR depends on https://github.com/kinvolk/baselayout/pull/12. Without the baselayout changes, the system simply does not allow users to authenticate and log in at all.

This PR is based on previous work by @krnowak , which was actually pam 1.4.0. So I updated the version to 1.5.1, removed unnecessary patches, updated comments, etc.

## How to use

```
./build_packages
```

## Testing done

CI passed http://localhost:9091/job/os/job/manifest/1607/cldsv/ .


pothos (commented):
The `sys-libs/pam` folder content wasn't deleted when importing the new version. This leads to the ported commit https://github.com/kinvolk/coreos-overlay/pull/732/commits/d397b7e1e7b69fd057ac5588d1634b36abd1104e that doesn't create `sys-libs/pam/files/tmpfiles.d/pam.conf` as the original commit https://github.com/kinvolk/coreos-overlay/commit/8d8002043bd218703b1dcbef1eeacd0fad0a2371 did – the next person updating the package may run into this and start archeology again.

dongsupark:
All right.
In this PR so far I have simply tried to keep Krzesimir's commits as much as possible.
Tomorrow I will try to sort out commits again as you said.

krnowak:
I won't be mad if my authorship of commits disappears, so don't worry about it.

pothos:
Thanks for the work so far. I'll volunteer with `git -i` to update the first commit, modify the /usr commit and drop the drop commit. If it works well, I will force-push here.

pothos:
Done

pothos (approved):
(Once the merge commit is used for baselayout)

pothos (commented):
I would not rename the baselayout-3.6.8-r1.ebuild symlink but it doesn't hurt either (just sometimes when cherry-picking…).

dongsupark:
Removed also unnecessary sentences in README.

PR #733, opened 2020-12-09
.github: add Github Actions for auto-updating linux-firmware

dongsupark:
Add Github Actions for coreos-firmware, just like other Kernel packages, basically to detect new releases from the upstream [linux-firmware](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git) repo.

Fixes https://github.com/kinvolk/Flatcar/issues/289.


marga-kinvolk (approved):
LGTM

PR #734, opened 2020-12-09
flatcar 2643 azure pro use temp beta pro update channel

t-lo:


t-lo:
Closed as non-working. Was worth a try :smile:

PR #740, opened 2020-12-11
dev-vcs/repo: update to 2.8 for python3

dongsupark:
The [repo v2.10](https://groups.google.com/g/repo-discuss/c/rpSfMCl83Sk) was released dropping python2 support. As a result, every `repo init` failed to run.

To unblock CI builds, we released mantle [v0.15.2](https://github.com/kinvolk/mantle/releases/tag/v0.15.2), including a workaround to set the target branch to [`maint`](https://gerrit.googlesource.com/git-repo/+/refs/heads/maint), which still supports python2.
Now with cork v0.15.2, `cork create` or `cork update` will work well for now.

However, the current state is quite fragile. It will get broken again when the upstream `maint` branch changes.
We should update `dev-vcs/repo` in coreos-overlay to 2.x with python3, and get it included in Flatcar SDK, so we could later set the target branch in mantle back to `stable`.

At the moment, none of the source repos has the tarball for repo 2.10, neither GCS nor Gentoo distfiles.
So for now we update it to 2.8. It will be linked to python 3.6 in Flatcar SDK.

Also note that we do not have to keep `files/repo-1.25` script in the coreos-overlay repo, because the script is simply identical to the upstream `repo` script.
I am not sure why the third-party script was there in the first place. So simply remove the script.

## How to use

```
sudo emerge dev-vcs/repo ## in the SDK
```

## Testing done

SDK builds: http://localhost:9091/job/os/job/manifest/1622/cldsv/

Locally test with the result SDK tarball, like:
```
# install cork v0.15.2

# create SDK env with the new SDK
cork create --sdk-url-path=flatcar-jenkins/developer/sdk --sdk-version=2020.12.11+dev-flatcar-master-1622 --repo-branch=stable
```


dongsupark:
CI passed.

was able to confirm it is working as expected, with cork 0.15.2, like:
```
cork create --sdk-url-path=flatcar-jenkins/developer/sdk --sdk-version=2020.12.11+dev-flatcar-master-1622 --repo-branch=stable
```

marga-kinvolk (approved):
LGTM

PR #745, opened 2020-12-14
app-emulation/containerd: Explicitly disable TCP

pothos:
Even though the containerd cri plugin should not listen on TCP by
default it still seems to do so.
Explicitly disable it to make sure it won't be on.

# How to use

Run the kola tests

# Testing done

[Ongoing](http://localhost:9091/job/os/job/manifest/1634/cldsv/)


pothos:
Not needed, the port does not serve CNI but just the streaming API.

PR #746, opened 2020-12-14
sys-apps/systemd: Drop sysctl rp_filter patches

pothos:
The patches were not taking effect because they did not set
net.ipv4.conf.default.rp_filter for new interfaces. Also, they got
overwritten by the baselayout configuration which takes precedence
and is the place for Flatcar-specific sysctl settings.
The desired configuration was enforced there:
https://github.com/kinvolk/baselayout/pull/13

*Note:* Pick for 2605 and 2705

# How to use

This PR alone makes no difference because it just removes settings that weren't effective.

# Testing done

None yet


t-lo (commented):
LGTM, thanks.

PR #753, opened 2020-12-18
sys-kernel: enable CONFIG_DEBUG_INFO_BTF

dongsupark:
Enable Kernel config `CONFIG_DEBUG_INFO_BTF` to make Kernel have debug info for BTF (BPF Type Format) and CO-RE (BPF compile once, run everywhere).

See also https://github.com/kinvolk/Flatcar/issues/225.

It is now possible to add the kernel config, as Alpha 2723.0.0 is going to include `dev-util/dwarves` 1.19. Without 1.19, arm64 builds would fail.

## How to use

```
./build_packages --board=amd64-usr
./build_packages --board=arm64-usr
...
```

## Testing done

CI passed: http://localhost:9091/job/os/job/manifest/1659/cldsv/ (based on SDK 2723.0.0)


pothos:
Ah, we could have included it in this Alpha since it gets built with the new SDK already

dongsupark:
CI passed for both amd64 and arm64.

pothos (approved):

PR #760, opened 2020-12-30
Move main to kernel 5.10.4

marga-kinvolk:
# Move to linux 5.10.4 

With this change, we start tracking linux 5.10. Only a couple of config changes were necessary:

1. Explicitly include `CONFIG_IP6_NF_IPTABLES`, as it's no longer implicitly included.
   https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=144b0a0e608690d46e9a77819249bdd8d23bdcb6

2. Move `CONFIG_EFI_VARS` to amd64 only, as it's no longer available on non Intel platforms. It's been replaced by `CONFIG_EFIVARS_FS` which is already enabled on the common config.
   https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=963fabf37f6a94214a823df0a785e653cb8ad6ea

# Testing done

Built the local image. Ran Jenkins with this branch (manifest 1711), packages built successfully and tests ran. All amd64 tests passed. For arm64, on top of the old `setenforce: command not found ` failures, there were these failures:

```
cl.filesystem/blacklist:
files.go:210: Blacklisted files or directories found:     /run/udev/data/+platform:Fixed MDIO bus.0

cl.basic/DbusPerms:
cluster.go:78: kolet: Process exited with status 1
```

Also, I had to re-run a number of tests due to flaky kernel panics. Looking at the nightly results, all of this seems to be the current state of ARM :-/. So, while not great, at least there's no regression.


sys-kernel/coreos-modules/files/commonconfig-5.10
```diff
@@ -272,6 +271,7 @@ CONFIG_INPUT_MOUSEDEV=m
 # CONFIG_INPUT_MOUSEDEV_PSAUX is not set
 CONFIG_IOSCHED_BFQ=y
 CONFIG_IO_STRICT_DEVMEM=y
+CONFIG_IP6_NF_IPTABLES=m
```
krnowak, line 274: How did you find that out? I'm curious.

krnowak (approved):
Looks good.

pothos:
We also have to update `KV_MAIN` to be `5.10` in `.github/workflows/kernel-releases-main.yml`

sys-kernel/coreos-modules/files/commonconfig-5.10
marga-kinvolk: I was facing weird build issues and @dongsupark pointed me to this missing entry, I don't know how he knew.  The commit message I pointed to triggers this because this code is no longer implicitly included.

krnowak:
Oh, yeah, forgot about the github workflow update.

marga-kinvolk:
Ok, I've updated the workflow. Thanks for pointing it out.

sys-kernel/coreos-modules/files/commonconfig-5.10
krnowak: Thanks.

dongsupark: > I was facing weird build issues and @dongsupark pointed me to this missing entry, I don't know how he knew. The commit message I pointed to triggers this because this code is no longer implicitly included.

TL;DR: No magic. `git log` and `make menuconfig` followed by reading help messages.

Long version:

It is one of the trickiest kernel configuration issues, because of nested dependencies of configs.

First time when we try to build Kernel 5.10 with an old kernel config, we see error like that:

```
 * ERROR: sys-kernel/coreos-modules-5.10.0::coreos failed (configure phase):
 *   Requested options not enabled in build:
 *     CONFIG_IP6_NF_FILTER
 *     CONFIG_IP6_NF_MANGLE
...
```

However, we already know that those kernel configs are already enabled in commonconfig-5.10 as `m`.
Then we need to dig into each config, to figure out what happened with that in Kernel 5.10.
For simplicity, let's take `CONFIG_IP6_NF_FILTER` as an example.

The first possible case would be that `CONFIG_IP6_NF_FILTER` disappeared in Kernel 5.10.
Hard to believe, but we need to check.
Go to the local Linux source repo, run `make menuconfig`, press `/` for help messages, and search for `IP6_NF_FILTER`. (without the prefix `CONFIG_`!)

Then you can see the help message like below:

```
Symbol: IP6_NF_FILTER [=m]
 Type  : tristate 
  Defined at net/ipv6/netfilter/Kconfig:191   
   Prompt: Packet filtering
    Depends on: NET [=y] && INET [=y] && IPV6 [=y] && NETFILTER [=y] && IP6_NF_IPTABLES [=m]
```

So Kernel 5.10 still has `CONFIG_IP6_NF_FILTER`, which did not disappear.
It means the issue cannot be resolved by removing deprecated options, no low hanging-fruit.

The next possible case would be some dependency issues.
As you see above, the kernel config depends on `NET`, `INET`, `IPV6`, `NETFILTER`, and `IP6_NF_IPTABLES`. It is quite hard to believe that the first 4 configs about subsystems have suddenly disappeared.

However, what about the last one, `IP6_NF_IPTABLES`? Is it really enabled by our kernel config?
No, neither `commonconfig-5.10` nor `amd64_defconfig-5.10` have that.
Geez, how has it worked at all so far?

Run `git log v5.9..v5.10 ./net/netfilter` in the Kernel tree, and try searching for `IP6_NF_IPTABLES`.
Nothing. Fantastic. Commit messages are unfortunately not perfect for such purposes.

What about `git diff v5.9..v5.10 ./net/netfilter`?
Found. `select IP6_NF_IPTABLES` was removed from `net/netfilter/ipvs/Kconfig`.
`git log net/netfilter/ipvs/Kconfig` shows that commit 144b0a0e608690d46e9a77819249bdd8d23bdcb6 ("pvs: remove dependency on ip6_tables") changed the file to remove dependency on `IP6_NF_IPTABLES`.

Until 5.9, it worked well because `CONFIG_IP_VS_IPV6` automatically enabled `CONFIG_IP6_NF_IPTABLES`.
Starting from 5.10, it does not work because the implicit dependency was removed.
To solve it, we need to explicitly enable `CONFIG_IP6_NF_FILTER` in commonconfig-5.10.

If it works, do the similar processes for other kernel config errors.

marga-kinvolk: Wow, thanks a lot for the detailed explanation!

PR #762, opened 2021-01-04
x11-drivers/nvidia-drivers: Handle NVIDIA Version upgrades

sayanchowdhury:
# x11-drivers/nvidia-drivers: Handle NVIDIA Version upgrades

Move the NVIDIA variables into a seperate file inorder to handle them
when Flatcar upgrade happens

# How to use
```bash
emerge-amd64-usr x11-drivers/nvidia-drivers
```

# Testing done

No testing done yet.


x11-drivers/nvidia-drivers/nvidia-drivers-450.80.02.ebuild
```diff
@@ -18,9 +18,11 @@ S="${WORKDIR}"
...
 
 src_install() {
-	insinto "/usr/share/oem"
-	doins -r "${FILESDIR}/units"
-	exeinto "/usr/share/oem/bin"
-	doexe "${FILESDIR}/bin/install-nvidia"
-	doexe "${FILESDIR}/bin/setup-nvidia"
+  insinto "/usr/share/oem"
+  doins -r "${FILESDIR}/units"
+  exeinto "/usr/share/nvidia"
```
pothos, line 23: It could also be `insinto` and `doins`, or?

pothos (changes requested):
The file needs to be in a separate package that is installed in the generic image

x11-drivers/nvidia-drivers/files/bin/setup-nvidia
```diff
@@ -3,9 +3,8 @@ set -euo pipefail
 
 . /usr/share/flatcar/release
 . /usr/share/flatcar/update.conf
+. /usr/share/nvidia/release
```
marga-kinvolk, line 6: I'm on the fence here, but I think it would be better if this file was also in /usr/share/flatcar, so that it's clear to the user that it's a flatcar specific file...

x11-drivers/nvidia-drivers/files/bin/setup-nvidia
```diff
@@ -3,9 +3,8 @@ set -euo pipefail
...
 
-NVIDIA_DRIVER_VERSION=450.80.02
-NVIDIA_PRODUCT_TYPE=tesla
```
marga-kinvolk, line 8: Can the product type change with the release version? I thought it was associated with the card in the machine and thus it would be hardware dependent, not release dependent, no?

sayanchowdhury: > Can the product type change with the release version? I thought it was associated with the card in the machine and thus it would be hardware dependent, not release dependent, no?

No, the product type cannot change with the release version. `tesla` is the name behind the data center GPUs that NVIDIA manufactures. The product type is an umbrella for a couple of hardware in that category. Though I read somewhere NVIDIA is planning to drop the name `tesla` to avoid confusion.

x11-drivers/nvidia-drivers/files/bin/setup-nvidia
sayanchowdhury: Makes sense. we can rename it to `/usr/share/flatcar/nvidia-metadata`?

marga-kinvolk: Sounds good to me.

sayanchowdhury:
Picked to `flatcar-2605`, `flatcar-2705`

marga-kinvolk (approved):
LGTM

PR #763, opened 2021-01-04
Track latest init commit

marga-kinvolk:
Needed for kinvolk/Flatcar#278


pothos (approved):

marga-kinvolk:
I've tested the nightly image and verified that it fixes the issue, so I've cherry picked this change to all channels.

PR #766, opened 2021-01-06
app-emulation/containerd: Disable shim debug logs

pothos:
Debug output clutters the logs which with K8s liveness/readiness probes
quickly becomes a problem.

Fixes https://github.com/kinvolk/Flatcar/issues/313

# How to use

Run a cluster with liveness/readiness probes and see that the log output is reduced because it does not contain `level=debug` entries anymore.

# Testing done

None


marga-kinvolk (approved):
LGTM

PR #768, opened 2021-01-07
coreos-base/afterburn: Restart on failure and keep unit active

pothos:
When the metadata server is unavailable for some time the service did
not retry. Also, the service was triggered possibly multiple times
each time another service pulled it in which can cause problems if,
e.g., the service experiences a failure and corrupts the existing file
which could have been kept because rerunning wasn't needed.

Fixes https://github.com/kinvolk/Flatcar/issues/311

# How to use

Build and verify that all tests pass

# Testing done

None


sayanchowdhury (commented):
LGTM

sayanchowdhury (approved):

PR #730, opened 2020-12-04
Update systemd-9999.ebuild to use systemd-resolved's stub resolver

f0o:
# Re-Enable the stub-resolver to fully utilize systemd-resolved...
...for applications that do not do nsswitch (https://github.com/kinvolk/baselayout/pull/10)

REQUIRES https://github.com/kinvolk/baselayout/pull/11

Goes in tandem with https://github.com/kinvolk/baselayout/pull/10

# How to use

Set up Split-Horizon DNS as outlined in https://github.com/kinvolk/baselayout/pull/10 

# Testing done

Same tests as https://github.com/kinvolk/baselayout/pull/10 but with `dig` instead of ping to bypass nsswitch

Related: https://github.com/kinvolk/Flatcar/issues/285

PR #769, opened 2021-01-08
sys-apps/systemd: Switch back to using a merged /etc/resolv.conf

pothos:
Using only 127.0.0.53 for /etc/resolv.conf causes problems for
Kubernetes which is not systemd-resolved aware yet (the kubelet passes
on /etc/resolv.conf contents to containers).
Switch back for now to merging all DNS servers into /etc/resolv.conf
which breaks split DNS and we need to document how to make split DNS
work for those that want it.

(Reverts the change of https://github.com/flatcar-linux/coreos-overlay/pull/730)

# How to use

Needs a docs PR and a kubelet upstream issue would also be nice

# Testing done

### DIFF
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -194,6 +194,14 @@ src_prepare() {
 	# from Gentoo. Thus we dropped the `if ! use vanilla` code
 	# here.
 
+	# Flatcar: The Kubelet takes /etc/resolv.conf for, e.g., CoreDNS which has dnsPolicy "default", but unless
+	# the kubelet --resolv-conf flag is set to point to /run/systemd/resolve/resolv.conf this won't work with
+	# /etc/resolv.conf pointing to /run/systemd/resolve/stub-resolv.conf which configures 127.0.0.53.
+	# See https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#known-issues
+	# This means that users who need split DNS to work should point /etc/resolv.conf back to /run/systemd/resolve/stub-resolv.conf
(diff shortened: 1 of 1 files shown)


sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -194,6 +194,14 @@ src_prepare() {
 	# from Gentoo. Thus we dropped the `if ! use vanilla` code
 	# here.
 
+	# Flatcar: The Kubelet takes /etc/resolve.conf for, e.g., CoreDNS which has dnsPolicy "default", but unless
+	# the kubelet --resolv-conf flag is set to point to /run/systemd/resolve/resolv.conf this won't work with
+	# /etc/resolve.conf pointing to /run/systemd/resolve/stub-resolv.conf which configures 127.0.0.53.
+	# See https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#known-issues
+	# This means that users who need split DNS to work should have to /etc/resolve.conf back to /run/systemd/resolve/stub-resolv.conf
```
marga-kinvolk, line 201:
```suggestion
	# Flatcar: The Kubelet takes /etc/resolv.conf for, e.g., CoreDNS which has dnsPolicy "default", but unless
	# the kubelet --resolv-conf flag is set to point to /run/systemd/resolve/resolv.conf this won't work with
	# /etc/resolv.conf pointing to /run/systemd/resolve/stub-resolv.conf which configures 127.0.0.53.
	# See https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#known-issues
	# This means that users who need split DNS to work should point /etc/resolv.conf back to /run/systemd/resolve/stub-resolv.conf
```

marga-kinvolk (approved):
LGTM, some fixes for typos.

PR #771, opened 2021-01-11
app-emulation/containerd: Switch to default socket location

pothos:
The upstream socket is under /run/containerd/containerd.sock which many
tools like crictl will use by default and diverging causes users to
always have to configure a non-default location.
Switch to the upstream default while still keeping a symlink so that
users are not forced to update their configurations they had to do for
the non-default location. This also keeps Docker using the old socket
location as an assertion that the symlink works. The state directory
is also switch to the default location.

# How to use

Build an image and run the tests

# Testing done

Manually created a drop-in config with the new contents.

### DIFF
--- a/app-emulation/containerd/files/config.toml
+++ b/app-emulation/containerd/files/config.toml
@@ -1,7 +1,7 @@
 # persistent data location
 root = "/var/lib/containerd"
 # runtime state information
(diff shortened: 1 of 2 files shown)


marga-kinvolk (approved):
LGTM

pothos:
Picked for flatcar-2605 and flatcar-2705

PR #772, opened 2021-01-12
.github: fix env vars and sed expressions

dongsupark:
1. pass env variables explicitly as string

Since Kernel 5.10, Github Actions simply stopped working.
What happens is that `KV_MAIN` gets passed as environmental variable to the inline script, but not as string but float, because it contains `.`.
Apparently the last digit of the misinterpreted float number is afterwards simply dropped by YAML parsing library used by GA.
As a result, `KV_MAIN` becomes `5.1` instead of `5.10`, `versionMain` becomes simply `5.10`, not `5.10.6`.
Then in the next steps, both `VERSION_NEW` and `VERSION_OLD` become `5.10`, and the script thinks it is already the latest version, so simply does not create a new pull request.

Simply set `KV_MAIN` or others explicitly as strings, by adding quotes, to avoid such issues.

2. Sscape dot correctly in sed expressions

So far all sed expressions have used correct regular expressions around semantic versions, around `.`.
As a result, they matched strings even without correct dots in place.
We need to escape the dot correctly.

### DIFF
--- a/.github/workflows/containerd-apply-patch.sh
+++ b/.github/workflows/containerd-apply-patch.sh
@@ -13,14 +13,14 @@ fi
 
 pushd "${SDK_OUTER_SRCDIR}/third_party/coreos-overlay" >/dev/null || exit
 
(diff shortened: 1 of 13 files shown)


marga-kinvolk (approved):
LGTM

PR #778, opened 2021-01-13
dev-lang/rust: Move TargetResult to Target in our local patch

marga-kinvolk:
# Refactor the gentoo patch

The target has undergone signficant refactoring. It's no longer a
TargetResult, it's just a Target. And also the vendor is now part of the
options.

# Testing done

Did a toolchains+packages run on Jenkins, it passed.

### DIFF
--- a/dev-lang/rust/files/1.49.0-gentoo-musl-target-specs.patch
+++ b/dev-lang/rust/files/1.49.0-gentoo-musl-target-specs.patch
@@ -27,16 +27,16 @@ index 0000000..420fe7c
 --- /dev/null
 +++ b/compiler/rustc_target/src/spec/aarch64_gentoo_linux_musl.rs
 @@ -0,0 +1,11 @@
-+use crate::spec::TargetResult;
++use crate::spec::Target;
 +
-+pub fn target() -> TargetResult {
-+    let mut base = super::aarch64_unknown_linux_musl::target()?;
++pub fn target() -> Target {
++    let mut base = super::aarch64_unknown_linux_musl::target();
 +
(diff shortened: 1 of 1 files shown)


dev-lang/rust/files/1.48.0-gentoo-musl-target-specs.patch
```diff
@@ -27,50 +27,50 @@ index 0000000..420fe7c
 --- /dev/null
```
dongsupark, line 27: Nit: the patch file name should be '1.49.0-gentoo...`.

dongsupark (approved):
Looks good. Thanks!
This change is exactly what I thought about yesterday.

I confirm that this PR fixes the build issue with Rust 1.49.
It would be great if we could send the PR also to upstream Gentoo, as there is no 1.49 there yet.
But no need to hurry. ;-)

dev-lang/rust/files/1.48.0-gentoo-musl-target-specs.patch
marga-kinvolk: Sure, will rename. Thanks.

marga-kinvolk:
CI passed, so I'm submitting this. Thanks for the review.

How do I send this to Gentoo?

dongsupark:
> How do I send this to Gentoo?

I used to simply create Github PRs on https://github.com/gentoo/gentoo .
Though it is required to explicitly add `Signed-off-by` to commit messages.

In the case of Rust, you might need to also create a new file `rust-1.49.0.ebuild`, because Gentoo usually keeps old ebuilds for a while.

PR #780, opened 2021-01-15
sys-boot/syslinux: Update the SRC_URI to use kernel.org directly

krnowak:
It seems that there is no "kernel" mirror specified in third party
mirrors files in profiles any more. And gentoo seems to have switched
to direct kernel.org URLs anyway, probably because kernel.org is using
also some mirroring system, so we don't have to. Also, this syslinux
version is quite old, so if its tarball ever was on distfiles mirror,
it's gone by now.

This came up when building SDK with new portage, and this fix moved the build further.


sayanchowdhury (approved):
LGTM
(not tested locally)

krnowak:
Started a build to be doubly-sure it doesn't break with old portage (http://localhost:9091/job/os/job/manifest/1797/).

krnowak:
Build succeeded, merging.

PR #781, opened 2021-01-15
dev-libs/openssl: Drop bindist from RESTRICT variable

krnowak:
It's really a hindrance during bootstrap, and we would be looking into
ways of making an exception for openssl anyway. Using
package.accept_restrict file does not do the trick, apparently because
of catalyst using its own portage config.

This came up during building SDK stage1 with new portage. The error was like:

```
!!! All ebuilds that could satisfy "dev-libs/openssl:0=" for /tmp/stage1root/ have been masked.
- dev-libs/openssl-1.1.1g::coreos (masked by: bindist in RESTRICT)
```

It either stopped being racy with new portage or was simply quite much more likely to be triggered. With restriction removed, the build went further.

On a second thought, this restrict didn't make sense to me, because the logic behind this restrict is "if you disable bindist USE flag, then add bindist to RESTRICT", while I think it should be more like "if you enable bindist USE flag then add bindist to RESTRICT too". My reasoning is that bindist USE flag causes openssl to contain some possibly patent-encumbered algorithms, so they obviously make distribution of openssl unclear, thus it should be restricted. And to acknowledge the restriction, one should add bindist to ACCEPT_RESTRICT somewhere. So either I understood things wrong, or it's a bug in openssl packaging. I could file a bug in gentoo tracker about this.

### DIFF
--- a/dev-libs/openssl/openssl-1.1.1g.ebuild
+++ b/dev-libs/openssl/openssl-1.1.1g.ebuild
@@ -32,7 +32,7 @@ SLOT="0/1.1" # .so version of libssl/libcrypto
 [[ "${PV}" = *_pre* ]] || \
 KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv s390 sparc x86 ~x86-linux"
 IUSE="+asm bindist elibc_musl rfc3779 sctp cpu_flags_x86_sse2 sslv3 static-libs test tls-heartbeat vanilla zlib"
-RESTRICT="!bindist? ( bindist )
+RESTRICT="
 	!test? ( test )"
 
 RDEPEND=">=app-misc/c_rehash-1.7-r1


krnowak:
Started a build to be doubly-sure it doesn't break with old portage (http://localhost:9091/job/os/job/manifest/1796/)

pothos:
If you retrigger the build with setting `SDK_VERSION: sdk-new` you can test if the image builds with the compiled SDK

dongsupark (approved):

krnowak:
> If you retrigger the build with setting `SDK_VERSION: sdk-new` you can test if the image builds with the compiled SDK

Argh, yes. I did that for my portage update builds, but forgot about it for this build… Retriggered, build: http://localhost:9091/job/os/job/manifest/1798/

krnowak:
Build succeeded, merging.

krnowak:
I haven't filed any issue on gentoo, found some other two:

https://bugs.gentoo.org/531540
https://bugs.gentoo.org/762850

So it looks like some patents expired (apparently last of them in 2020), so gentoo will maybe decide to drop bindist use flag from openssl and openssh.

krnowak:
And with bindist use flag dropped, the bindist restriction would be dropped too, I believe.

PR #782, opened 2021-01-15
profiles: disable cxx for net-libs/nghttp2

dongsupark:
When building `net-libs/nghttp2` needed by curl 7.74, build fails when checking for prerequisites of boost libs.

```
configure:20402: checking whether the Boost::ASIO library is available
configure:20433: x86_64-cros-linux-gnu-g++ -std=c++14 -c -O2 -pipe -mtune=generic -g   conftest.cpp >&5
configure:20433: $? = 0
configure:20447: result: yes
configure:20540: error: Could not find a version of the library!
```

To avoid such issues, we should disable the `cxx` USE flag for `net-libs/nghttp2`.

This PR should be merged together with https://github.com/kinvolk/portage-stable/pull/137.

## How to use

```
emerge-amd64-usr net-misc/curl
```

## Testing done

CI passed http://localhost:9091/job/os/job/manifest/1816/cldsv/

### DIFF
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -116,3 +116,7 @@ sys-kernel/coreos-firmware -savedconfig
 
 # Make kmod support kernel modules compressed via lzma(xz)
 sys-apps/kmod lzma
+
+# net-libs/nghttp2 should be built with -cxx to avoid issues with boost 1.65.
+# configure script is not able to check if Boost:ASIO library exists.
+net-libs/nghttp2 -cxx


profiles/coreos/base/package.use
```diff
@@ -116,3 +116,6 @@ sys-kernel/coreos-firmware -savedconfig
 
 # Make kmod support kernel modules compressed via lzma(xz)
 sys-apps/kmod lzma
+
+# net-libs/nghttp2 should be built with -cxx to avoid build issues with boost
```
marga-kinvolk, line 120: Can you add to this comment why the build against boost fails? Is it because we have a very old boost? Because we don't have it at all? Thanks

dongsupark: Added comments and commit messages.

```
configure:20402: checking whether the Boost::ASIO library is available
configure:20433: x86_64-cros-linux-gnu-g++ -std=c++14 -c -O2 -pipe -mtune=generic -g   conftest.cpp >&5
configure:20433: $? = 0
configure:20447: result: yes
configure:20540: error: Could not find a version of the library!
```

Flatcar's boost version is 1.65.
Configure script fails only when checking for ASIO, not during generic checks for boost itself.
Not sure if it would be better with newer versions of boost. Could not figure out.

marga-kinvolk: Ok, thanks!

marga-kinvolk (approved):

profiles/coreos/base/package.use
krnowak: You would need to check the `config.log` file for details. The checks for ASIO seem to find header files, but it failed to find `libboost_system` library for some reason (some nonstandard location of those?). Fixing that probably would require passing `--with-boost=<path>` or `--with-boost-libdir=<path>`.

Boost checks in configure are here: https://github.com/nghttp2/nghttp2/blob/master/configure.ac#L459-L473
This `for` loop found no `libboost_system` lib: https://github.com/nghttp2/nghttp2/blob/master/m4/ax_boost_asio.m4#L87
This file (used with `AX_BOOST_BASE` macro) is responsible for finding CPPFLAGS and LDFLAGS for boost: https://github.com/nghttp2/nghttp2/blob/master/m4/ax_boost_base.m4

Man, looking at that configure/m4 gobbledygook is such a blast from the past…

dongsupark: > Fixing that probably would require passing --with-boost=<path> or --with-boost-libdir=<path>.

Right, thanks.

As you said, we can fix the nghttp2 build issue by adding the following lines to the ebuild:
```
diff --git a/net-libs/nghttp2/nghttp2-1.41.0.ebuild b/net-libs/nghttp2/nghttp2-1.41.0.ebuild
index 675e2b5f..39d8699a 100644
--- a/net-libs/nghttp2/nghttp2-1.41.0.ebuild
+++ b/net-libs/nghttp2/nghttp2-1.41.0.ebuild
@@ -59,6 +59,8 @@ multilib_src_configure() {
                --disable-werror
                --without-cython
                $(use_enable cxx asio-lib)
+               --with-boost="${EROOT}"/usr/lib64
+               --with-boost-libdir="${EROOT}"/usr/lib64
                $(use_enable debug)
                $(multilib_native_use_enable hpack-tools)
                $(use_enable static-libs static)
```

It has nothing to do with boost version.
It is just an issue of nghttp2 not being able to find correct boost libs.
Strangely enough, I cannot reproduce the issue when running `configure` directly in the source tree. It also does not happen when I run all these on a clean Gentoo system.

Also the boost issue is not the only issue of nghttp2.
If we enable `cxx` USE flag in nghttp2, it starts to pull in `openssl[-bindist]`.
Then it turns into the same issue we fixed last week in openssl ebuilds.

So would we need to move nghttp2 from portage-stable to coreos-overlay, add the boost configure options to the nghttp2 ebuilds, and finally we need to remove the `openssl[-bindist]` requirement from nghttp2.
I would say, in theory it is possible to do so, but it is becoming quite a snowball. ;-)
Let's create another issue to keep track of the issue to address it later, and let's merge this PR as it is.

dongsupark: Filed an issue https://github.com/kinvolk/Flatcar/issues/323.

krnowak: Oh good grief, the bindist story… Thanks for filing the issue.

krnowak (approved):

PR #786, opened 2021-01-19
app-emulation/open-vmdk: Update

krnowak:
Replace the use of deprecated git eclass with git-r3 and bump the
commit version to latest version. This version dropped a dependency on
jq.

It is a breaking change for users of mkova.sh, since it has changed
the order of parameters to allow passing multiple vmdk files to it.

I have made this commit for the portage update and it has pushed the build further.

I haven't found any use of the `mkova.sh` script in jenkins or build scripts. I haven't found any use of `vmdk-converter` anywhere either, so not sure why this package needs to be a part of SDK. But this is something to investigate later.

Test build: http://localhost:9091/job/os/job/manifest/1831/cldsv/

### DIFF
--- a/app-emulation/open-vmdk/open-vmdk-1.0.ebuild
+++ b/app-emulation/open-vmdk/open-vmdk-1.0.ebuild
@@ -4,7 +4,7 @@
 
 EAPI=5
 
-inherit git-2
+inherit git-r3
 
 DESCRIPTION="Tool to convert vmdk to an ova file"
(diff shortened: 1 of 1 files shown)


krnowak:
Build failed, probably picked old portage-stable, so curl license was missing.

New test build: http://localhost:9091/job/os/job/manifest/1834

krnowak:
Build succeeded.

dongsupark (approved):

PR #790, opened 2021-01-20
sys-apps/systemd: Update to systemd v247

sayanchowdhury:

# sys-apps/systemd: Update to systemd v247

The patches are adjusted, as they were failing to apply. 

# How to use
```
emerge-amd64-usr systemd
./build_image
```

# Testing done

Scheduled a Jenkins [build](http://localhost:9091/job/os/job/manifest/1842/cldsv/)


sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -401,16 +401,16 @@ multilib_src_install_all() {
...
-		rm -r "${ED}${rootprefix}"/lib/udev/hwdb.d || die
-	fi
-
 	# Flatcar: Upstream uses keepdir commands to keep some empty
 	# directories.
 	#
 	# Flatcar: TODO: Consider using that instead of
 	# systemd_dotmpfilesd "${FILESDIR}"/systemd-flatcar.conf below.
 
+	if use hwdb; then
```
krnowak, line 410: Any reason for this change?

krnowak (approved):
LGTM. But had one question.

sys-apps/systemd/systemd-9999.ebuild
sayanchowdhury: I was using `vimdiff` for comparing the diff with upstream, and the `# Flatcar` comment seemed out of place.

krnowak: Alright.

I think this comment was about the stuff that exists in upstream ebuild, but we have dropped. So this comment is probably bound to seem out of place.

PR #791, opened 2021-01-21
sys-block/open-iscsi: Update iscsi to 2.1.3

sayanchowdhury:
# sys-block/open-iscsi: Update iscsi to 2.1.3

This commits updates the open-iscsi to 2.1.3. open-iscsi == 2.0.873 [upstream](https://github.com/open-iscsi/open-iscsi) did not ship systemd service file, so this package shipped it own service files. With the 2.1.x upgrade, the upstream now packages the systemd service files as well. This commit tries to go close to the Gentoo upstream, and replaces the local service files with the upstream ones. 

This change:
- Replaces the initiator name prefix from `iqn.2016-04.com.flatcar.iscsi` to iscsi-iname default `iqn.2005-03.org.open-iscsi`
- Service for iscsi daemon from `iscsid` to `iscsi`

Needs to be merge with kinvolk/flatcar-docs#152

# How to use

```bash
emerge-amd64-usr sys-block/open-iscsi
./build_image
```

# Testing done

[Jenkins](http://localhost:9091/job/os/job/manifest/1844/cldsv/) build passed


dongsupark (approved):
Looks good, Thanks!
Yesterday I have tested it, and it worked well.

marga-kinvolk:
We should include a prominent note in the release notes that the service has been renamed.

pothos:
Maybe an alias could be set up for the previous name but I haven't tried it and don't know if there are limitations: https://www.freedesktop.org/software/systemd/man/systemd.unit.html

sayanchowdhury:
The upstream ships iscsid and iscsi, but the primary one is the `iscsi`. https://github.com/open-iscsi/open-iscsi/tree/master/etc/systemd

krnowak:
Is there a reason to have a bunch of ebuilds instead of just one?

pothos:
Ah, so it's not a rename but both have different purposes. `iscsi.service` depends on `iscsid.d` through `iscsid.socket`.

sayanchowdhury:
Picked to `flatcar-2605` & `flatcar-2705`

sys-block/open-iscsi/files/iscsi-init.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=One time configuration for iscsi.service
+After=systemd-tmpfiles-setup.service
+ConditionPathExists=!/etc/iscsi/initiatorname.iscsi
+
+[Service]
+Type=oneshot
+RemainAfterExit=no
+ExecStart=/usr/bin/sh -c 'echo "InitiatorName=`/usr/sbin/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
```
pothos, line 9: minor issue: using
```
`cmd`
```
requires bash, otherwise it should be `$(cmd)`

marga-kinvolk:
Reviewing this change for the release notes, I found that a bunch of files are gone from /etc, is this expected?
```
-drwxr-xr-x  3 root    root       4096 ./etc/iscsi
-drwxr-xr-x  2 root    root       4096 ./etc/iscsi/ifaces
--rw-r--r--  1 root    root       5984 ./etc/iscsi/ifaces/iface.example
--rw-r--r--  1 root    root         15 ./etc/iscsi/initiatorname.iscsi
--rw-r--r--  1 root    root       1167 ./etc/iscsi/initiatorname.iscsi.example
--rw-------  1 root    root      13338 ./etc/iscsi/iscsid.conf
```

marga-kinvolk:
And these are the unit differences:
```
--rw-r--r--  1 root    root        368 ./usr/lib64/systemd/system/iscsid.service
+-rw-r--r--  1 root    root        364 ./usr/lib64/systemd/system/iscsid.service
 -rw-r--r--  1 root    root        175 ./usr/lib64/systemd/system/iscsid.socket
--rw-r--r--  1 root    root        525 ./usr/lib64/systemd/system/iscsi.service
--rw-r--r--  1 root    root        407 ./usr/lib64/systemd/system/iscsiuio.service
+-rw-r--r--  1 root    root        297 ./usr/lib64/systemd/system/iscsi-init.service
+-rw-r--r--  1 root    root        513 ./usr/lib64/systemd/system/iscsi.service
+-rw-r--r--  1 root    root        403 ./usr/lib64/systemd/system/iscsiuio.service
 -rw-r--r--  1 root    root        165 ./usr/lib64/systemd/system/iscsiuio.socket
```

I had understood that we were moving from iscsid to iscsi, but it seems both units existed before?

sayanchowdhury:
- the changes from /etc to /usr looks good and is expected. they moved to around during the 2.1.2 release
- 2.1.x series brings the systemd units so yes, they were brought in the last open-iscsi upgrade but there were couple issues(or rather applying CoreOS/Flatcar specific patches) that were missed during the upgrade. They were fixed in this update.

PR #794, opened 2021-01-22
flatcar-eks: facilitate provisioning EKS workers

marga-kinvolk:
# Facilitate provisioning EKS workers

This change adds a new flatcar-eks package, that ships with all scripts
needed to join a Flatcar instance to an EKS cluster.

It includes the bootstrap.sh script used on Amazon Linux, to keep
compatibility with existing provisioning tools.

The package is included from the oem-ec2-compat package, when the board
is aws_pro, and it's part of board-packages, so that it's built by the
os/board/packages job.

## How to use / Testing done

Using and testing this change takes time and effort.

1. Build the flatcar-eks package, then build the an ami_vmdk_pro image, with use flag aws_pro. This will include the flatcar-eks package in the image, as well as the ignition config. This can be done in the CI by using the marga-kinvolk/aws-pro branch from the flatcar-scripts repo.
2. Upload the AMI to Amazon
3. Create an EKS cluster and provision worker nodes with that AMI, by creating a Launch template and then adding a Node Group with that Launch template.
4. The nodes should become healthy after they've been provisioned and it should be possible to interact with them with `kubectl` provided the right config (still working on coming up with clear instructions for this last part).

## Node Configuration 
The workers need to be assigned a role with these four policies:
 - arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy
 - arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy
 - arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly
 - arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess

And the userdata needs to be:
```
#!/bin/bash
/usr/share/oem/eks/bootstrap.sh <cluster-name>
```

## Note for reviewers
The package includes some files taken from an Amazon repo. Some are taken verbatim, while others are modified to fit our needs. I suggest starting the review by looking at the ebuild file, which will make clear where things come from.


coreos-base/flatcar-eks/files/download-kubelet.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+set -euo pipefail
+
+# Get the cluster name from /etc/eks/cluster.conf
+. /etc/eks/cluster.conf
+if [[ -x ${CLUSTER_NAME} ]]; then
```
krnowak, line 6: Cluster name is supposed to be an executable or did you mean using `-z`?

coreos-base/flatcar-eks/files/download-kubelet.sh
```diff
@@ -0,0 +1,52 @@
...
+  echo "No cluster name found. Aborting"
+  exit 1
+fi
+
+# Query the Kubernetes version of the cluster
+mkdir -p /opt/eks
+shopt -s expand_aliases
+alias aws="docker run --rm --network host -v /opt/eks:/eks amazon/aws-cli"
+CLUSTER_VERSION=$(aws eks describe-cluster --name ${CLUSTER_NAME} --output text --query cluster.version)
```
krnowak, line 15: I'd quote `${CLUSTER_NAME}`.

krnowak (commented):
Some initial things I found by reading.

coreos-base/flatcar-eks/files/download-kubelet.sh
marga-kinvolk: Yeah, it was supposed to be -z, thanks for catching this. I don't know how it didn't fail in my tests.

coreos-base/flatcar-eks/files/download-kubelet.sh
marga-kinvolk: Sure, will do, thanks.

coreos-base/flatcar-eks/files/bootstrap.sh
```diff
@@ -0,0 +1,376 @@
...
+    else
+        echo "No entry for $INSTANCE_TYPE in $MAX_PODS_FILE. Not setting max pods for kubelet"
+    fi
+fi
+
+mkdir -p /etc/systemd/system/kubelet.service.d
+
+cat <<EOF > /etc/systemd/system/kubelet.service.d/10-kubelet-args.conf
+[Service]
+Environment='KUBELET_ARGS=--node-ip=$INTERNAL_IP --pod-infra-container-image=$PAUSE_CONTAINER'
```
krnowak, line 323: In `kubelet.service` we specify `--node-ip ${COREOS_EC2_IPV4_LOCAL}` and then apply `$KUBELET_ARGS`. And here the script adds the `--node-ip` option to `KUBELET_ARGS`. Isn't it going to clobber the `--node-ip` option we set in `kubelet.service`? Should this be patched out?

marga-kinvolk: It's actually the same IP, so I've just removed it from the service file and kept this one.  Thanks for noticing!

krnowak (approved):

marga-kinvolk:
Cherry picked to flatcar-2605 and flatcar-2705

PR #800, opened 2021-01-25
coreos-base,profiles: delete nss

dongsupark:
As `dev-libs/nss` is not used anywhere, let's simply remove nss.
The only ebuild that pulls in is `net-misc/curl`, but only if the USE flag `nss` is enabled.
As the `nss` flag is disabled for curl, we do not need to keep `dev-libs/nss` at all.

According to the change, delete `dev-libs/nss` from profiles and sdk-depends.

This PR should be merged together with https://github.com/kinvolk/portage-stable/pull/141 .

## How to use

```
./bootstrap_sdk
./build_packages
...
```

## Testing done

CI passed http://localhost:9091/job/os/job/manifest/1877/cldsv/


marga-kinvolk (approved):
LGTM

PR #802, opened 2021-01-26
coreos-base/hard-host-depends: bring back dependency on dev-libs/nspr

dongsupark:
Now that `dev-libs/nss` is removed from the depencencies list of hard-host-depends, SDK does not include `dev-libs/nspr` any more.
As a result, `dev-lang/spidermonkey` fails to build, because it requires `dev-libs/nspr` in the SDK.
It is not sufficient to have nspr under `/build/amd64-usr`.

Add `dev-libs/nspr` back to the dependencies of `hard-host-depends`, to make it included in the SDK.

## How to use

```
emerge -pv --tree --emptytree coreos-base/hard-host-depends
```

and check its output includes `dev-libs/nspr`.

# Testing done

CI passed: http://localhost:9091/job/os/job/manifest/1901/cldsv/


marga-kinvolk (approved):
LGTM

krnowak (approved):
Looks good, but I don't understand why nspr emerged and installed during package-matrix job is not enough for spidermonkey?

krnowak:
SDK job finished successfully, and so now the toolchains job is running. But I can't find "nspr" in SDK job's console output, which confuses me even more.

dongsupark:
> Looks good, but I don't understand why nspr emerged and installed during package-matrix job is not enough for spidermonkey?

I am not completely sure why.
However I have seen plenty of ebuilds that simply look up libraries under `/usr/lib64`, not `/build/amd64-usr/usr/lib64`.
I suppose it is such a corner case as well.
Anyway we should look into spidermonkey's configure script to see exactly why.
However, the spidermonkey package is one of the big can of worms. ;-)

> SDK job finished successfully, and so now the toolchains job is running. But I can't find "nspr" in SDK job's console output, which confuses me even more.

It is not sufficient to look into the SDK job's console output.
Try downloading the result [SDK tarball](https://storage.googleapis.com/flatcar-jenkins/developer/sdk/amd64/2021.01.26+dev-flatcar-master-1901/flatcar-sdk-amd64-2021.01.26+dev-flatcar-master-1901.tar.bz2), and see if it contains nspr.
I am sure it does.

dongsupark:
spidermonkey during packages-matrix looks good.
Merging.

PR #816, opened 2021-02-02
app-emulation/cri-tools: add a new package

dongsupark:
[cri-tools](https://github.com/kubernetes-sigs/cri-tools) consists of `crictl`, a standard CLI tool for CRI, and `critest`, a validation testsuite for CRI.
The latter is not needed by Flatcar, but the former tool is quite often used nowadays.

In the past, only the Edge channel included `cri-tools`.
But it was never a part in Stable-Alpha.
Nowadays, Stable-Alpha already started to [include](https://github.com/kinvolk/Flatcar/issues/201) other missing parts like `containerd-shim-v[12]`. 
However, `cri-tools` has not been a part of Flatcar.

We should get cri-tools included in the Flatcar image, so tools like kubeadm could easily use it for provisioning.

Fixes https://github.com/kinvolk/Flatcar/issues/325.

## How to use

```
emerge-amd64-usr app-emulation/cri-tools
```

## Testing done

CI passed: http://localhost:9091/job/os/job/manifest/1942/cldsv/


app-emulation/cri-tools/cri-tools-1.18.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+SLOT="0"
+KEYWORDS="amd64 arm64"
+
+S=${WORKDIR}/cri-tools-${PV}
+
+DEPEND=""
+RDEPEND="${DEPEND}"
+
+src_compile() {
+	GO_LDFLAGS="-s -w -extldflags=-Wl,-z,now,-z,relro,-z,defs"
```
krnowak, line 28: Upstream's Makefile also specifies `-X $(PROJECT)/pkg/version.Version=$(VERSION)` in LDFLAGS, should we also do it? Also, it specifies `-tags selinux`, which we do not. Is it something we should use too?

In general, why not using upstream's Makefile, but using only the `crictl` target? Maybe with patching the upstream's Makefile to add those linker flags for binary size reduction?

app-emulation/cri-tools/cri-tools-1.18.0.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit coreos-go bash-completion-r1
```
krnowak, line 6: `bash-completion-r1` seems to be unnecessary, since we don't install any bash completions.

app-emulation/cri-tools/Manifest
```diff
@@ -0,0 +1 @@
+DIST cri-tools-1.18.0.tar.gz 5295512 BLAKE2B 9dda4a401b858e98aa60521d3268d8f6ff6a2c1bd2c219a0e5e9e5360fba58d41e97258c636da26ee57bbfc3286798c74107c7de92d3e3398bdf53b7d9078e94 SHA512 849b496e6d19eec29e0e56f0b37c094649311b80514f319b0c68c6519e3ddd48e1973c3d288144db998826676416644ec705b7ea0916c21898b40928ea800af1
```
krnowak, line 1: Any reason for picking up 1.18 instead of 1.20? The package seems to be unmaintained on Gentoo side, which maybe is why this package is lagging.

app-emulation/cri-tools/cri-tools-1.18.0.ebuild
dongsupark: > Upstream's Makefile also specifies -X $(PROJECT)/pkg/version.Version=$(VERSION) in LDFLAGS, should we also do it?

Yes, let's do that.

> it specifies -tags selinux, which we do not. Is it something we should use too?

The `selinux` tag is anyway enabled by default in upstream cri-tools.
In theory, we can allow it to be disabled via the USE flag, though that needs a lot of work in eclass.

> In general, why not using upstream's Makefile, but using only the crictl target?

ok, let me see if it is possible.

> Maybe with patching the upstream's Makefile to add those linker flags for binary size reduction?

In theory yes, possible.
Will do that in the future, as it needs a lot of time until the upstream PR could be merged.

app-emulation/cri-tools/cri-tools-1.18.0.ebuild
dongsupark: > bash-completion-r1 seems to be unnecessary, since we don't install any bash completions.

ok, will do.

app-emulation/cri-tools/Manifest
dongsupark: > Any reason for picking up 1.18 instead of 1.20? The package seems to be unmaintained on Gentoo side, which maybe is why this package is lagging.

I just happened to copy Gentoo ebuild, which was 1.18. That's all.

Ideally, I would like to update the version to 1.20.
However, when I changed the version to 1.20, then the binary size of `crictl` gets much bigger, from 19M to 30M.
Not sure I like it. Let me think how we can avoid.

app-emulation/cri-tools/cri-tools-1.18.0.ebuild
dongsupark: > In general, why not using upstream's Makefile, but using only the crictl target?

No, it is not possible to use upstream cri-tools' Makefile, because the Makefile does not allow custom variables like `BUILDTAGS` or `GO_LDFLAGS` to be configured.
Unfortunately it is the only way to run `go_build` like that.
Added comments about that.

app-emulation/cri-tools/cri-tools-1.18.0.ebuild
dongsupark: Done.

app-emulation/cri-tools/Manifest
dongsupark: Fortunately crictl 1.19 is only 21M. So I simply updated to 1.19.
Added commit messages about that.

Another option to resolve the size issue is to get crictl included in torcx tarballs.
However, it is not so a clean way, because cri-tools is not an existing package of the "torcx world". cri-tools is not containerd.
We would need to touch all over the place in torcx, it does not look good in the long term.
I would not like to add another dimension of complexity to torcx.

krnowak (approved):

PR #817, opened 2021-02-02
flatcar-eks: add missing mkdir and update to latest versions

marga-kinvolk:
# flatcar-eks: add missing mkdir and update to latest versions

The bootstrapping script relies on `/etc/docker` existing, but this
directory doesn't exist on vanilla Flatcar. Add the missing call to
`mkdir -p /etc/docker` before the directory gets used.

Also, update the upstream files to their latest version.


krnowak (approved):

marga-kinvolk:
Cherry picked to flatcar-2605, flatcar-2705 and flatcar-2765

PR #820, opened 2021-02-03
coreos-base/coreos-init: Show OEM in motd

pothos:
- coreos-base/coreos-init: Show OEM in motd
  This pulls in https://github.com/kinvolk/init/pull/34
- coreos-base/oem-*: add Pro to OEM name where applicable 
  Using the change in https://github.com/kinvolk/init/pull/34
  we can show the OEM on the motd, and by including "Pro" in the OEM
  name we can also show whether it is a Pro image or not. Later this
  may be revisited if the /usr/../os-release file is the place for it.

# How to use

Build an OEM image and see the motd after login

# Testing done

None


marga-kinvolk (approved):
LGTM

PR #826, opened 2021-02-04
app-emulation,torcx: delete docker 1.12

dongsupark:
One of the torcx profiles in Flatcar is for docker 1.12, which is outdated since a long time.
It takes ~27 MB of space in production images almost for no reason.
We should delete docker 1.12.

After deletion:
```
$ df -h /usr
Filesystem       Size  Used Avail Use% Mounted on
/dev/mapper/usr  985M  843M   91M  91% /usr
```

Note, let's keep the remaining file name as before, `docker-1.12-no.json`, to be consistent with naming scheme of the torcx repo itself of Flatcar.

Now that docker 1.12 is gone, we can also delete containerd 0.2.5 and `app-emulation/runc` 1.0.0_rc2, which had dependencies on docker 1.12.
Note, please do not get confused, we do not delete `app-emulation/docker-runc` 1.0.0_rc2, because that one is still needed by Docker 17.03.

Fixes https://github.com/kinvolk/Flatcar/issues/339.

This PR should be merged together with https://github.com/kinvolk/flatcar-scripts/pull/116.

## How to use

```
./build_packages
```

## Testing done

CI passed: http://localhost:9091/job/os/job/manifest/1960/cldsv/
Note, the CI build was running including the deletion of Go 1.6, and it worked well.
Though the Go 1.6 part is not included in this PR, because if we delete Go 1.6 at once, the nightly build will be broken.
We should merge deletion of Go 1.6 one day after we merged this PR.


krnowak (approved):
This will need some extra release note. Also, are there some docs that mention docker 1.12 to be removed?

dongsupark:
> This will need some extra release note. Also, are there some docs that mention docker 1.12 to be removed?

Sure, we will explicitly say about that in the next Alpha release note.
don't know about the "docs", I would leave other folks do the releases decide the details about that.

pothos:
I thought that in torcx there is also something to remove, but if we document that even though the mechanics are still in place, they are not supported, it's also fine.

PR #828, opened 2021-02-04
coreos-base/gmerge: Stop installing gmerge script

krnowak:
It is not used anywhere during the build process, thus drop
it. Dropping it makes it easier to port this ebuild to python3, since
there will be only one script to port to python3. The
`emerge-gitclone` script will need porting anyway, because it imports
portage code, which will become python3 after the update.

Most likely the package should be then renamed to
`coreos-base/emerge-gitclone`, but this can be done later.

Test build passed just fine without the tool. It's being pulled into dev container through coreos-base/dev-util package.

Test build: http://localhost:9091/job/os/job/toolchains/1294/


dongsupark (approved):
Yeah, I cannot figure out why gmerge was installed.
It was just like that since the beginning.

`coreos-base/cros-devutils` still calls gmerge, but only if `! cros_host`.
So it looks ok to remove gmerge.

krnowak:
Yeah, and funnily enough, gmerge is gone at upstream - https://chromium.googlesource.com/chromiumos/platform/dev-util/+/refs/heads/main

PR #834, opened 2021-02-09
Contributions/update to samba 4.12.9

t-lo:
# Update Samba to 4.12.9

# How to use

```
$ emerge-amd64-usr samba
```

# Testing done

CI succeeded (Manifest build #1992) on both ARM64 and AMD64.

# Dependencies

Requires portage-stable PR https://github.com/kinvolk/portage-stable/pull/144 for prerequisites.


profiles/coreos/targets/sdk/package.accept_keywords
```diff
@@ -11,3 +11,6 @@ virtual/rust ~amd64
 # Accept go utilities
 dev-go/glide ~amd64
 dev-go/godep ~amd64
+
+# Accept perl libs required to build >=samba-4.12
+dev-perl/Parse-Yapp ~amd64 ~arm64
```
krnowak, line 16: Is this needed? The `dev-perl/Parse-Yapp` package has keywords like `KEYWORDS="~alpha amd64 arm arm64 hppa …`, so shouldn't the build be fine without this entry?

coreos-devel/sdk-depends/sdk-depends-0.0.1.ebuild
```diff
@@ -47,6 +47,7 @@ DEPEND="
 	sys-firmware/edk2-ovmf
 	sys-fs/btrfs-progs
 	sys-fs/cryptsetup
+	dev-perl/Parse-Yapp
```
krnowak, line 50: I still don't get it why we need to add some deps here and not let the ebuild pull those deps.

profiles/coreos/targets/sdk/package.accept_keywords
t-lo: Yes, it's needed, or else the SDK build breaks. Core issue here is that Samba is a package required by the OS image while Parse-Yapp needs to be part of the SDK in order to build Samba. Gentoo is not able to map this dependency.

coreos-devel/sdk-depends/sdk-depends-0.0.1.ebuild
t-lo: Same as above, we need that perl lib in the SDK to build the Samba package. "SDK" is a different target than "base-image" and Gentoo has no concept we could use to map this.

profiles/coreos/targets/sdk/package.accept_keywords
t-lo: (We were initially missing `~arm64` there which is why the ARM SDK build broke)

krnowak (approved):
I don't understand the changes in `package.accept_keywords`, but seems like the build process was ignoring keywords in ebuild (?), for which I'd say that there is something fishy going on.

Let's get it in anyway.

PR #840, opened 2021-02-12
portage update: update portage and related packages to newer versions

krnowak:
Please see the commit messages for details.

Needs to be merged together with:

https://github.com/kinvolk/flatcar-dev-util/pull/5
https://github.com/kinvolk/flatcar-scripts/pull/118
https://github.com/kinvolk/portage-stable/pull/145

After merging the flatcar-dev-util PR, I'll update this PR to point to the merge commit.


krnowak:
Test build: http://localhost:9091/job/os/job/manifest/2021/

app-portage/repoman/README.md
```diff
@@ -0,0 +1,3 @@
+This is a fork of app-portage/repoman package. The sole reason for
+having it here is to drop support for python 3.8 and 3.9 we haven't
+yet packaged.
```
dongsupark, line 3: Just a note for me.
In the future, we should really update eclass files to make them accept python 3.8 and 3.9.
It has been a huge pain for everyone who touches the code.

profiles/coreos/base/use.mask
```diff
@@ -1,2 +1,6 @@
 # Never enable experimental code
 kdbus
+
+# We default to python 3.6 for now
+python_targets_python3_7
+python_single_target_python3_7
```
dongsupark, line 6: Note for me.
When we introduce the python 3.7 target in the future, we need to unmask 3.7 here again.

dongsupark (approved):
Ok if CI passed.

app-portage/repoman/README.md
krnowak: That may mean custom modification of the eclass, because the updated upstream version may, for example, drop support for python2 we still package or for old EAPI we still use.

PR #846, opened 2021-02-15
sys-kernel: enable kernel config CONFIG_BPF_LSM

mauriciovasquezbernal:
Enable this option to make it possible to use LSM hooks with BPF.

Fixes https://github.com/kinvolk/Flatcar/issues/343 

# Test 

I built a local qemu image and performed the following tests.

1. Checked that CONFIG_BPF_LSM is actually enabled

```
$ cat /proc/config.gz | gunzip | grep CONFIG_BPF_LSM
CONFIG_BPF_LSM=y
```

2. Deployed a[ simple eBPF](https://github.com/kinvolk/libbpf-bootstrap/blob/mauricio/test-lsm/src/minimal.bpf.c) program that uses the LSM open_file hook. 

```
# compile example on local host (requires clang-10 and llvm-10)
$ cd libbpf-bootstrap/src
$ make
# copy to test instance
$ scp -P 2222 minimal core@localhost:/tmp
```
```
$ sudo /tmp/minimal 
libbpf: loading object 'minimal_bpf' from buffer
libbpf: elf: section(2) lsm/file_open, size 104, link 0, flags 6, type=1
libbpf: sec 'lsm/file_open': found program 'restrict_filesystems' at insn offset 0 (0 bytes), code size 13 insns (104 bytes)
libbpf: elf: section(3) license, size 4, link 0, flags 3, type=1
libbpf: license of minimal_bpf is GPL
libbpf: elf: section(4) .rodata.str1.1, size 13, link 0, flags 32, type=1
libbpf: elf: skipping unrecognized data section(4) .rodata.str1.1
libbpf: elf: section(5) .BTF, size 458, link 0, flags 0, type=1
libbpf: elf: section(6) .BTF.ext, size 144, link 0, flags 0, type=1
libbpf: elf: section(7) .symtab, size 120, link 11, flags 0, type=2
libbpf: looking for externs among 5 symbols...
libbpf: collected 0 externs total
libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0
Successfully started!
.............................................
```

In other terminal

```
$ sudo cat /sys/kernel/debug/tracing/trace_pipe
            sudo-1007    [012] d...   797.038227: bpf_trace_printk: file opened

            sudo-1007    [012] d...   797.038259: bpf_trace_printk: file opened

            sudo-1007    [012] d...   797.038539: bpf_trace_printk: file opened

            sudo-1007    [012] d...   797.038555: bpf_trace_printk: file opened
```

It shows the eBPF program is working as expected.


krnowak:
@mauriciovasquezbernal : Want to have a stab at running a test build? Or want me to start it?

t-lo:
It's the first change after branching off 2583. Looks like a low-risk change to me and was tested locally (had a chat with @mauriciovasquezbernal about it); we might even consider to just merge and wait for the nightly.

mauriciovasquezbernal:
I updated the PR description with the details of the tests I did. I don't know that well the details of how the merge / build / test processes work for Flatcar, so I let you folks choose the best way to move ahead.

krnowak (approved):
Well, so it builds (sometimes adding a config breaks a build actually), you tested that it works, so let the nightly build run the tests on the change.

dongsupark (approved):

mauriciovasquezbernal:
There is a potential [performance penalty](https://lore.kernel.org/bpf/20200820164753.3256899-1-jackmanb@chromium.org/) with this PR as the `BPF LSM` hook is enabled by default on Flatcar (we're not explicitly setting [`CONFIG_LSM`](https://github.com/torvalds/linux/blob/master/security/Kconfig#L278) and it includes `bpf` by default).

Other distros like [Ubuntu](https://lists.ubuntu.com/archives/kernel-team/2020-November/115178.html) don't enable bpf in `CONFIG_LSM` by default. I think we could follow the same approach, enable `CONFIG_BPF_LSM` but don't add bpf to the default list in `CONFIG_LSM`, users willing to use this feature could boot with the `lsm=...bpf` parameter on the kernel. 

Said that, I'm wondering how to remove `bpf` on Flatcar since we're not setting `CONFIG_LSM`. I have two ideas:
1. Define it to the default value it has now but removing bpf. `CONFIG_LSM="lockdown,yama,loadpin,safesetid,integrity,selinux,smack,tomoyo,apparmor"`. If some of the `DEFAULT_SECURITY_` flags is set it could be out of sync. 
2. Add some logic [here](https://github.com/kinvolk/coreos-overlay/blob/main/sys-kernel/coreos-modules/coreos-modules-5.10.16.ebuild#L22) to remove bpf from the list. It seems to me the better way to go, but it's not done for any other flag so I prefer to ask before. 

Any suggestions to do this? 
Thanks!

t-lo:
I think 1. is a cleaner approach than 2. The initramfs hack in the build script makes sense because it integrates the kernel / modules config with the initramfs step in the build process. Adding CONFIG_LSM there would not be related to any other build step; having it in the kernel config would be much more transparent on what we want to achieve, therefore easier to maintain.

mauriciovasquezbernal:
Makes sense to me. I updated the PR to include `CONFIG_LSM="lockdown,yama,loadpin,safesetid,integrity,selinux,smack,tomoyo,apparmor"`, this is the [default](https://github.com/torvalds/linux/blob/f40ddce88593482919761f74910f42f4b84c004b/security/Kconfig#L284) value in the kernel when there is not any `DEFAULT_SECURITY_*` flag set.

I tested again locally and it works fine, the bpf hook is disabled by default and can be enabled with the `lsm=bpf` kernel boot parameter.

t-lo:
LGPM, let's pull it in.

PR #851, opened 2021-02-18
sys-kernel/dracut: Drop the use of mirror://kernel

krnowak:
The kernel mirror was dropped from the thirdpartymirrors file in
profiles, so it's use needs to be replaced with an address to
kernel.org.

It's a quick hack to fix the issue without spending too much time on updating every single package. The change in SRC_URIs were done with `sed -i'' -e 's!mirror://kernel!https://www.kernel.org/pub!g' $(git grep 'mirror://kernel' | cut -f1 -d: | grep -e '\.\(eclass\|ebuild\)$' | sort -u)`.

Needs to be merged together with https://github.com/kinvolk/portage-stable/pull/148.

Test build: http://localhost:9091/job/os/job/manifest/2062/


krnowak:
CI passed.

dongsupark (approved):

krnowak:
Cherry-picked to flatcar-2783.

PR #853, opened 2021-02-18
.github: Minor fixes in kernel update workflow

krnowak:
Some small fixes for stuff I have noticed when checking the kernel update workflow failure. Not yet tested, but should be harmless.

### DIFF
--- a/.github/workflows/kernel-apply-patch.sh
+++ b/.github/workflows/kernel-apply-patch.sh
@@ -34,7 +34,7 @@ done
 
 popd >/dev/null || exit
 
-generate_patches sys-kernel coreos-sources Linux
+generate_patches sys-kernel coreos-sources Kernel
 
 apply_patches
 
--- a/.github/workflows/kernel-apply-patch.sh
+++ b/.github/workflows/kernel-apply-patch.sh
@@ -34,7 +34,7 @@ done
 
 popd >/dev/null || exit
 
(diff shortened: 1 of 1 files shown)


.github/workflows/kernel-apply-patch.sh
```diff
@@ -25,16 +25,16 @@ if [[ "${VERSION_NEW}" = "${VERSION_OLD}" ]]; then
...
+  pushd "sys-kernel/coreos-${pkg}" >/dev/null || exit
+  git mv "coreos-${pkg}"-*.ebuild "coreos-${pkg}-${VERSION_NEW}.ebuild"
+  sed -i -e '/^COREOS_SOURCE_REVISION=/s/=.*/=""/' "coreos-${pkg}-${VERSION_NEW}.ebuild"
+  popd >/dev/null || exit
 done
 
 popd >/dev/null || exit
 
-generate_patches sys-kernel coreos-{sources,kernel,modules} Linux
+generate_patches sys-kernel coreos-sources Linux
```
dongsupark, line 37: Good catch! :see_no_evil: 

Can you please replace the description string, "Linux" with "Linux Kernel" or simply "Kernel"?

dongsupark (approved):

.github/workflows/kernel-apply-patch.sh
krnowak: Done. Went with "Kernel".

PR #856, opened 2021-02-19
dev-util/catalyst: Use python3 in scripts explicitly

krnowak:
Catalyst was using `/usr/bin/python` likely with an assumption that `python` is `python3` by default, which is not true in our case. This has caused a script to fail during importing a portage module, because it was a python2 script importing python3 module.

Test build: http://localhost:9091/job/os/job/manifest/2074 (the test build seems to have moved past the point where it has failed before).


dongsupark (approved):

krnowak:
The build hasn't yet finished, but I'm merging it right now - the fix certainly makes the build to progress, since it currently is on stage1, whereas before it failed during stage1. Another reason for merging it right now to have a fresh nightly build ready on Monday.

krnowak:
Cherry-picked to flatcar-2783.

PR #857, opened 2021-02-22
app-admin/etcd-wrapper: use docker for the etcd service and drop etcd-wrapper

pothos:
As rkt is deprecated we need to run the Etcd and Flannel container with Docker
    or Podman. The etcd/flannel-wrapper script is based on rkt arguments and
    can't be used in a compatible way but we cannot remove it since ct
    explicitly uses it in the ExecStart directive when writing out a
    drop-in file once flannel settings are given in a Container Linux
    Config.
    A better way to run the Flannel/etcd container image is Podman because
    Flannel depends on etcd but wants to be run before Docker so that it
    can set up the Docker networking. Etcd and Flannel are part of the
    Container Linux Config specification and thus can't be removed easily.
    For now we have to resort to running these services with Docker and try
    to restart Docker for the Flannel options to take effect (but that also
    terminates the etcd and flannel containers, causing the services to
    restart).

# How to use

Run kola tests for a cloud provider, not QEMU where the tests are excluded 

A fix for the Kubernetes test is here: https://github.com/kinvolk/mantle/pull/162

Note: From now on it's not possible to specify `Requires=flanneld.service` for `docker.service` because that creates a cyclic dependency.

# Testing done

[here](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2246/cldsv/)

### DIFF
--- a/app-admin/etcd-wrapper/etcd-wrapper-3.3.25.ebuild
+++ b/app-admin/etcd-wrapper/etcd-wrapper-3.3.25.ebuild
@@ -14,7 +14,8 @@ IUSE=""
 SLOT=0
 
 DEPEND=""
-RDEPEND=">=app-emulation/rkt-1.9.1[rkt_stage1_fly]"
+
+RDEPEND=">=app-admin/sdnotify-proxy-0.1.0"
 
 S=${WORKDIR}
 
--- a/app-admin/etcd-wrapper/files/etcd-member.service
+++ b/app-admin/etcd-wrapper/files/etcd-member.service
@@ -8,21 +8,22 @@ Conflicts=etcd2.service
 
 [Service]
 Type=notify
-Restart=on-failure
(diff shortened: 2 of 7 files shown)


pothos:
Found out that the Kubernetes test failed because I had `Wants=etcd-member.service` for flannel.service but the test has etcd on a separate node, updated the PR.
Another reason it failed was that the test sets `Requires=flanneld.service` and `After=flanneld.service` for Docker and now we have a deadlock – I will update the test because Flannel restarts Docker now and settings this dependency does not work anymore.

t-lo (approved):
LGTM, thank you!

pothos:
The test should still work with rkt instead of docker, otherwise we can add `ExcludeChannels: []string{"beta", "stable", "lts"},` and remove each channel when a > 2801 release comes in it.

PR #858, opened 2021-02-22
delete dhcpcd, containerd-stress

dongsupark:
Now that dhcpcd was removed from portage-stable, we can simply delete dependency on dhcpcd.

Each Flatcar production image includes a binary `containerd-stress`, as a part of torcx tarballs.
However it does not seem to be used anywhere.
It looks like a stress testing tool for containerd, so I don't see a good reason to keep it.
The binary was there since the beginning, via commit [fdd926949a10](https://github.com/kinvolk/coreos-overlay/commit/fdd926949a10c3011746a57a04fc36f451fb8009), but there is no comment or messages why it was needed.
We can simply remove `containerd-stress`.

This PR should be merged together with https://github.com/kinvolk/portage-stable/pull/149.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2075/cldsv/ (part of a combined testing branch)


marga-kinvolk (approved):
LGTM

PR #863, opened 2021-02-25
Fix polkitd on arm64

wrl:
Old/ancient versions of Spidermonkey were catastrophically broken on arm64 (at the very least because of https://bugzilla.mozilla.org/show_bug.cgi?id=1143022). This PR bumps the versions of polkit and Spidermonkey, and brings in a few new deps of polkit. It also includes a few extra Spidermonkey build fixes for arm64.

ref https://github.com/kinvolk/Flatcar/issues/156


pothos (commented):
Great work, a first comment on updating the polkit ebuild file:
There were downstream patches which need to be applied again in a separate commit after importing the Gentoo ebuild file. See https://github.com/kinvolk/coreos-overlay#updating
For example, one missing change I spotted is 5c4d184e22fd93ab926878a131150047b54d0b6c

dev-lang/spidermonkey/spidermonkey-68.12.0.ebuild
```diff
@@ -0,0 +1,264 @@
...
+SLOT="68"
+LICENSE="MPL-2.0"
+IUSE="cpu_flags_arm_neon debug +jit test"
+
+RESTRICT="!test? ( test )"
+
+BDEPEND="dev-lang/python:2.7
+	test? ( ${PYTHON_DEPS} )"
+
+DEPEND=">=dev-libs/icu-63.1:=
```
dongsupark, line 67: Oh, interesting. 
spidermonkey 68 simply pulls in `dev-libs/icu`, while spidermonkey 60 allows us to set `system-icu` USE flag.
For context, we have recently actually tried to delete `dev-libs/icu` completely, to minimize size of production image.

So it would be really great, if we could delete the dependency on icu here as well.
I see 2 options.
* Set the default version of spidermonkey to 60, and set USE flag like `dev-lang/spidermonkey -system-icu` in `profiles/coreos/base/package.use`.
* Keep spidermonkey 68, and update the ebuild to make it have `system-icu` USE flag just like 60.

dongsupark (commented):
First of all, thanks you so much for your work!

One issue I see is, its build fails like that:
```
The following USE changes are necessary to proceed:
 (see "package.use" in the portage(5) man page for more details)
>=dev-lang/python-2.7.15:2.7 sqlite
```
You can add the USE flag to `profiles/coreos/base/package.use`.

See below as well.

dev-lang/spidermonkey/spidermonkey-68.12.0.ebuild
dongsupark: I have just realized, polkit 0.116 pulls in spidermonkey 60, while polkit 0.117 needs spidermonkey 68.
polkit 0.116 is the default version both for amd and arm.
So the first option, spidermonkey 60 with an additional USE flag would be the way to go.

dongsupark:
> The following USE changes are necessary to proceed:
 (see "package.use" in the portage(5) man page for more details)
>=dev-lang/python-2.7.15:2.7 sqlite

Oh, adding the USE flag `sqlite` opens up a can of worm.
Building sqlite fails like:
```
In file included from /usr/include/features.h:489,
                 from /usr/include/bits/libc-header-start.h:33,
                 from /usr/include/stdint.h:26,
                 from sqlite3.c:13484:
/usr/include/gnu/stubs.h:7:11: fatal error: gnu/stubs-32.h: No such file or directory
    7 | # include <gnu/stubs-32.h>
      |           ^~~~~~~~~~~~~~~~
```
So far in Flatcar, sqlite was not installed for arm64.
Now that sqlite is enabled, we start seeing the build issue.

pothos:
`gnu/stubs-32.h` means that it's not using the cross compiler settings…

dongsupark:
Created a separate issue for the sqlite cross-compilation issue. https://github.com/kinvolk/Flatcar/issues/367

t-lo:
Let's make sure to ingest https://gitlab.freedesktop.org/polkit/polkit/-/commit/a04d13affe0fa53ff618e07aa8f57f4c0e3b9b81# while we're working on polkitd so we fix https://github.blog/2021-06-10-privilege-escalation-polkit-root-on-linux-with-bug/.

jepio:
I tried to test this and came across the following issues while building:
1. `portage-stable/eclass/autotools.eclass` needs this change cherry-picked gentoo/gentoo@53678d996f4fda6eff08e5eeed3a518004a8e1dc. Updating all of `eclass/autotools.eclass` is possible but drops support for EAPI 4, which we still have in use somewhere.
2. Patch the profile to specify the right use flags (maybe add jit to spidermonkey?):
```patch
diff --git a/profiles/coreos/targets/generic/package.use b/profiles/coreos/targets/generic/package.use
index f912708ed..2c8e0fee5 100644
--- a/profiles/coreos/targets/generic/package.use
+++ b/profiles/coreos/targets/generic/package.use
@@ -3,7 +3,8 @@

 app-admin/sudo         ldap sssd
 app-editors/vim                minimal
-dev-lang/python                -berkdb gdbm
+dev-lang/python                -berkdb gdbm sqlite
+dev-lang/spidermonkey   -system-icu
 dev-libs/dbus-glib     tools
 dev-libs/elfutils      -utils
 dev-libs/openssl       pkcs11
```
3. The gentoo ebuild does not work for cross-compilation as-is, we might want to upstream these changes:
```diff --git a/dev-lang/spidermonkey/spidermonkey-60.5.2_p0-r4.ebuild b/dev-lang/spidermonkey/spidermonkey-60.5.2_p0-r4.ebuild
index 52f96f099..9a6024018 100644
--- a/dev-lang/spidermonkey/spidermonkey-60.5.2_p0-r4.ebuild
+++ b/dev-lang/spidermonkey/spidermonkey-60.5.2_p0-r4.ebuild
@@ -77,8 +77,12 @@ src_configure() {

        tc-export AR RANLIB

+       # Flatcar: Mozilla has always used these flags differently then autoconf.
+       # https://bugzilla.mozilla.org/show_bug.cgi?id=475488
        ECONF_SOURCE="${S}/js/src" \
        econf \
+               --host="${CBUILD:-${CHOST}}" \
+               --target="${CHOST}" \
                --disable-jemalloc \
                --enable-readline \
                --with-system-nspr \
@@ -109,26 +113,13 @@ src_compile() {
        cd "${BUILDDIR}" || die
        if tc-is-cross-compiler; then
                tc-export_build_env BUILD_{AR,CC,CXX,RANLIB}
-               cross_make \
-                       MOZ_OPTIMIZE_FLAGS="" MOZ_DEBUG_FLAGS="" \
-                       HOST_OPTIMIZE_FLAGS="" MODULE_OPTIMIZE_FLAGS="" \
-                       MOZ_PGO_OPTIMIZE_FLAGS="" \
-                       host_jsoplengen host_jskwgen
                cross_make \
                        MOZ_OPTIMIZE_FLAGS="" MOZ_DEBUG_FLAGS="" HOST_OPTIMIZE_FLAGS="" \
                        -C config nsinstall
-               mv {,native-}host_jskwgen || die
-               mv {,native-}host_jsoplengen || die
                mv config/{,native-}nsinstall || die
-               sed -i \
-                       -e 's@./host_jskwgen@./native-host_jskwgen@' \
-                       -e 's@./host_jsoplengen@./native-host_jsoplengen@' \
-                       Makefile || die
-               sed -i -e 's@/nsinstall@/native-nsinstall@' config/config.mk || die
+               sed -i -e 's@/nsinstall@/native-nsinstall@' ../config/config.mk || die
                rm -f config/host_nsinstall.o \
-                       config/host_pathsub.o \
-                       host_jskwgen.o \
-                       host_jsoplengen.o || die
+                       config/host_pathsub.o || die
        fi

        MOZ_MAKE_FLAGS="${MAKEOPTS}" \
```
The whole `if tc-is-cross-compiler; then` seems to be copied from an older version and I think it doesn't make sense any longer.
4. The pkgconfig file for `mozjs-60` ends up with the following line: `Cflags: -include ${includedir}/mozjs-60/js/RequiredDefines.h -I${includedir}/mozjs-60`. pkg-config prepends sysroot to the `-I` parts, but not to `-include`. `polkit` doesn't build because of that.

wrl:
closed in favour of #1263

PR #864, opened 2021-02-25
net-fs/samba: EAPI=7, new USE flags, remove deps on icu

dongsupark:
Now that portage was updated to the latest version, we should update `EAPI` to 7.
It is mainly to allow ebuilds to make `BDEPEND` contain real build-time dependencies, not runtime ones.

Introduce a USE flag `spotlight`, to be able to disable the spotlight backend by default, as it is not needed by Linux.
Introduce a USE flag `rededit`, to be able to disable the rededit tool if needed.
Introduce a USE flag `glusterfs`, to be able to disable the glusterfs by default.
Introduce a USE flag `ntvfs`, to be able to disable the ntvfs-fileserver by default.
Since the `docbook-xsl-stylesheets` and `libxslt` are needed only at build time, we should move those deps to `BDEPEND`.

`dev-libs/icu` is needed only if spotlight is enabled. If not enabled, we should not pull in icu.

To minimize the size of bundled libraries of Samba, we should install only the `ldb` backend by default.

Disable unnecessary `regedit` in samba to minimize the package size.

This PR should be merged together with https://github.com/kinvolk/portage-stable/pull/150 .

See also a Gentoo PR https://github.com/gentoo/gentoo/pull/19650 , which is slightly different from this one.

Fixes https://github.com/kinvolk/Flatcar/issues/346

## Testing done

CI passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2106/

### DIFF
--- a/net-fs/samba/metadata.xml
+++ b/net-fs/samba/metadata.xml
@@ -16,12 +16,16 @@
 		<flag name="client">Enables the client part</flag>
 		<flag name="cluster">Enable support for clustering</flag>
 		<flag name="dmapi">Enable support for DMAPI. This currently works only in combination with XFS.</flag>
+		<flag name="glusterfs">Enable support for Glusterfs filesystem via <pkg>sys-cluster/glusterfs</pkg></flag>
 		<flag name="gpg">Use <pkg>app-crypt/gpgme</pkg> for AD DC</flag>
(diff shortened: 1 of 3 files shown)


dongsupark:
CI passed

marga-kinvolk (approved):
LGTM

PR #874, opened 2021-03-04
fix sdk bootstrap: use python3 in catalyst chroot-functions, set cros_host USE when bootstrapping

t-lo:
# SDK bootstrap related fixes

This change adds the USE flag cros_host to the
SDK's make.default, as part of a larger fix for the SDK bootstrap build.
The SDK bootstrap build was broken in stage 1 since package upgrades
were allowed to leak into that phase.

We now limit stage 1 to only "known good" package ebuilds, which caused
downstream breakage from missing flags in the stage 2 SDK bootstrapping.
This change fixes that breakage.

This change further contains a patch for catalyst to explicitly call python3 in 
chroot-functions.sh for portage related debug output.

# How to use

Needed in combination with
* portage-stable: https://github.com/kinvolk/portage-stable/pull/152
* flatcar-scripts: https://github.com/kinvolk/flatcar-scripts/pull/123 (the main PR)

```
$ cork enter
$ sudo ./bootstrap_sdk
```

# Testing done

SDK CI build (manifest #2158) passed.

### DIFF
--- a/profiles/coreos/amd64/sdk/make.defaults
+++ b/profiles/coreos/amd64/sdk/make.defaults
@@ -1,3 +1,7 @@
 # Enable optimizations for common x86_64 CPUs
 CFLAGS="-O2 -pipe -mtune=generic"
 CXXFLAGS="${CFLAGS}"
+
+# add cros_host to bootstrapping USE flags so SDK / toolchains bootstrapping
+# will use vim's vimrc instead of baselayouts',
+BOOTSTRAP_USE="$BOOTSTRAP_USE cros_host"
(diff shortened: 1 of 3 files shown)


profiles/coreos/amd64/sdk/make.defaults
```diff
@@ -1,3 +1,8 @@
 # Enable optimizations for common x86_64 CPUs
 CFLAGS="-O2 -pipe -mtune=generic"
 CXXFLAGS="${CFLAGS}"
+
+# Enable static-libs for bootstrapping b/c openssl needs static-libs on zlib
```
sayanchowdhury, line 5: I don't understand the comment about `static-libs` here.

t-lo: Oh, that's a left-over comment, sorry for that. I've updated the comment to describe why `cros_host` is there.

dongsupark (approved):

PR #875, opened 2021-03-04
sdnotify: avoid the github redirect

vbatts:
Signed-off-by: Vincent Batts <vbatts@kinvolk.io>

Background for this is that i've deleted our https://github.com/kinvolk/sdnotify-proxy and properly forked from coreos/sdnotify-proxy


marga-kinvolk (approved):
LGTM

marga-kinvolk:
This broke our builds because they were still pointing to the other repo.  We can fix our current live branches, but it will mean that older versions that are not patched won't build because of this.

PR #881, opened 2021-03-05
app-admin/toolbox: update to use docker instead of rkt

dongsupark:
Since rkt will be deprecated soon, we should make toolbox run docker instead of rkt.

Also delete dependency on `app-emulation/rkt`, and update hyperlinks.

It pulls in https://github.com/kinvolk/toolbox/pull/1 .

## How to use

```
emerge-amd64-usr app-admin/toolbox
```

## Testing done

local tests


marga-kinvolk (approved):
LGTM

PR #888, opened 2021-03-08
Add explicit paths to coreos-metadata and containerd executables in their unit files

marga-kinvolk:
While the execution of the unit may succeed by finding the executables
by searching the current PATH, calling `systemd-analyze verify` on the
units fails because this requires an absolute path.

Fixes: kinvolk/Flatcar#360

# Testing done

CI currently running

### DIFF
--- a/app-emulation/containerd/files/containerd-1.0.0.service
+++ b/app-emulation/containerd/files/containerd-1.0.0.service
@@ -6,8 +6,8 @@ After=network.target
 [Service]
 Delegate=yes
 Environment=CONTAINERD_CONFIG=/usr/share/containerd/config.toml
-ExecStartPre=mkdir -p /run/docker/libcontainerd
-ExecStartPre=ln -fs /run/containerd/containerd.sock /run/docker/libcontainerd/docker-containerd.sock
+ExecStartPre=/usr/bin/mkdir -p /run/docker/libcontainerd
+ExecStartPre=/usr/bin/ln -fs /run/containerd/containerd.sock /run/docker/libcontainerd/docker-containerd.sock
 ExecStartPre=-/sbin/modprobe overlay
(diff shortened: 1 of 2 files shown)


marga-kinvolk:
CI passed and I verified that the error reported doesn't happen anymore.  Without the fix:

```
Flatcar Container Linux by Kinvolk stable (2765.2.0)
core@localhost ~ $ systemd-analyze verify coreos-metadata.service
/run/systemd/system/docker.socket:8: ListenStream= references a path below legacy directory /var/run/, updating /var/run/docker.sock → /run/docker.s.
/usr/lib/systemd/system/dbus.socket:5: ListenStream= references a path below legacy directory /var/run/, updating /var/run/dbus/system_bus_socket → .
coreos-metadata.service: Command ln is not executable: No such file or directory
```

WIth the fix:

```
Flatcar Container Linux by Kinvolk developer 2021.03.08+dev-flatcar-master-2188
core@localhost ~ $ systemd-analyze verify coreos-metadata.service
/run/systemd/system/docker.socket:8: ListenStream= references a path below legacy directory /var/run/, updating /var/run/docker.sock → /run/docker.sock; please update the unit file accordingly.
/usr/lib/systemd/system/dbus.socket:5: ListenStream= references a path below legacy directory /var/run/, updating /var/run/dbus/system_bus_socket → /run/dbus/system_bus_socket; please update the unit file accordingly.
```

(the printed messages are just warnings, we should probably fix them in a separate change).

marga-kinvolk:
This caused breakage on the containerd side (I'm very confused because I saw the CI results and they were green). I'll revert the containerd part and keep only the coreos-metadata part.

pothos:
Ah, seen in https://github.com/kinvolk/coreos-overlay/pull/893

PR #893, opened 2021-03-10
Revert adding explicit path to the containerd unit

marga-kinvolk:
This unit is post-processed for torcx purposes and adding absolute paths breaks execution.

Needs to be cherry picked to flatcar-2765 and flatcar-2801


t-lo (approved):
LGTM, thank you for the quick turn-around!

dongsupark (approved):

PR #899, opened 2021-03-15
coreos-base/coreos-init: use alternative interface names for virtio

pothos:
This pulls in
https://github.com/kinvolk/init/pull/38
to set predictable network interface names as alternative interface
names for virtio devices, and also add a special hardcoded ens4v1
name for GCE because the special udev rule to rename the device
stopped working after the systemd 247 update.

Note: Backport to 2765 when merging (may need a flatcar-2765 branch in `init`)

# How to use/Testing done

See https://github.com/kinvolk/init/pull/38


marga-kinvolk (approved):
LGTM

PR #900, opened 2021-03-15
coreos-base/coreos-init: use alternative interface names for virtio (flatcar-2765)

pothos:
This pulls in
https://github.com/kinvolk/init/pull/38
to set predictable network interface names as alternative interface
names for virtio devices, and also add a special hardcoded ens4v1
name for GCE because the special udev rule to rename the device
stopped working after the systemd 247 update.
A branch flatcar-2765 was created in the init repository to backport
only this change.

See https://github.com/kinvolk/coreos-overlay/pull/899


marga-kinvolk (approved):
LGTM

PR #906, opened 2021-03-18
Update sys-libs/ncurses to 6.2-r1

blanquicet:
# Update sys-libs/ncurses

This PR updates package from version 6.1-r2 to 6.2-r1 and applies downstream patch.

Fixes kinvolk/Flatcar#364

# Testing done

After updating the package, I performed a scratch build to make sure your change is reproducible and did not break pristine builds.

**STATUS ON alpha-2801.0.0 BUILD WITHOUT MODIFICATIONS**
On build environment:
```
~/trunk/src/scripts $ equery-amd64-usr l sys-libs/ncurses 
 * Searching for ncurses in sys-libs ...
[I-O] [  ] sys-libs/ncurses-6.1-r2:0/6
```

Using the qemu image to interact with the image:
```
core@localhost ~ $ ldconfig -v 2>/dev/null | grep ncurses   
	libncursesw.so.6 -> libncursesw.so.6.1
	libncurses.so.6 -> libncurses.so.6.1
```

**AFTER UPDATE**
On build environment:
```
~/trunk/src/scripts $ equery-amd64-usr l sys-libs/ncurses
 * Searching for ncurses in sys-libs ...
[I-O] [  ] sys-libs/ncurses-6.2-r1:0/6
```

Using the qemu image to interact with the image:
```
core@localhost ~ $ ldconfig -v 2>/dev/null | grep ncurses   
	libncurses.so.6 -> libncurses.so.6.2
	libncursesw.so.6 -> libncursesw.so.6.2
```

In addition, using command `equery-amd64-usr d sys-libs/ncurses` I retrieved packages that depends on ncurses so I checked some of them have continued to work properl after the update, e.g. vim and less.


marga-kinvolk:
Thanks, I've triggered a CI run with this change.

marga-kinvolk:
CI failed for unrelated reasons, but it seems your branch is not up to date... Could you rebase against main? Thanks.

marga-kinvolk:
Nevermind, I rebased the branch used for the CI.

blanquicet:
Thanks @marga-kinvolk

marga-kinvolk:
CI passed.

sys-libs/ncurses/ncurses-6.2-r1.ebuild
```diff
@@ -270,24 +278,11 @@ multilib_src_install() {
...
-			# [u]rxvt users used to be pretty common.  Probably should drop this
-			# since upstream is dead and people are moving away from it.
-			rxvt{,-unicode}{,-256color}
-			# xterm users are common, as is terminals re-using/spoofing it.
-			xterm xterm-{,256}color
-			# screen is common (and reused by tmux).
-			screen{,-256color}
-			screen.xterm-256color
-		)
 		local x
```
dongsupark, line 284: How about adding a short comment, like `Flatcar: Add a symlink-usr USE flag for keeping a minimal set of terminfo files in /usr/share/terminfo.`?
I know, the original coreos commit did not have such comments, but recently Flatcar started writing comments like that, so other folks could easily recognize it is Flatcar-specific code.

sys-libs/ncurses/ncurses-6.2-r1.ebuild
```diff
@@ -105,6 +111,8 @@ multilib_src_configure() {
 }
 
 do_configure() {
```
dongsupark: Here as well.
How about adding a short comment, like `Flatcar: Also allow writes to /dev/ptmx, which sometimes causes the sandbox to fail Jenkins builds.`?

dongsupark (commented):
In general it looks good.

See below please:

sayanchowdhury (commented):
I agree with Dongsu suggestions. Other than that everything looks good. Thanks!

sys-libs/ncurses/ncurses-6.2-r1.ebuild
blanquicet: I agree. I will add it.

sys-libs/ncurses/ncurses-6.2-r1.ebuild
blanquicet: Same here, I will add this too.

blanquicet:
I made the changes in a new commit. I think we can collapse Flatcar patches into a single one next time we update this package. Is it OK?

dongsupark (approved):
The comments looks good. Thanks!

Usually we just squash Flatcar patches into one commit, via force-push.

blanquicet:
Understood. I will use it next time. Thanks!

PR #911, opened 2021-03-23
sys-apps/baselayout: enable pam_faillock with relaxed defaults

pothos:
This pulls in
https://github.com/kinvolk/baselayout/pull/17
to enable the pam_faillock module as replacement for pam_tally2.
The "faillock" binary can be used to see the login attempts and
account lock status which before was available with the pam_tally
command. While the tally defaults did not temporarily lock the
account on wrong password login attempts, this is done by default
with faillock. However, the default behavior was relaxed to allow
more wrong attempts and have a shorter lock time span.

# How to use

# Testing done


marga-kinvolk (approved):
LGTM

PR #915, opened 2021-03-25
dev-libs/openssl: Sync with Gentoo upstream 

sayanchowdhury:
# dev-libs/openssl: Sync with Gentoo upstream 

Bumps the openssl to 1.1.1k

# How to use

```
emerge-amd64-usr openssl
```

# Testing done

Tested locally


marga-kinvolk (commented):
I thought we were moving to a workflow of marking Flatcar changes with comments in the ebuild file, is that not the case?

dev-libs/openssl/openssl-1.1.1k.ebuild
```diff
@@ -47,12 +47,9 @@ PDEPEND="app-misc/ca-certificates"
 
 PATCHES=(
 	"${FILESDIR}"/${PN}-1.1.0j-parallel_install_fix.patch #671602
-	"${FILESDIR}"/${PN}-1.1.1k-release-changes.patch
```
pothos, line 50: The first commit should delete the whole folder and create it with new upstream contents, this should also delete these patch files which are not needed anymore.

sayanchowdhury:
> I thought we were moving to a workflow of marking Flatcar changes with comments in the ebuild file, is that not the case?

Agreed @marga-kinvolk I will add them here. Not all the ebuilds we have updated have that yet. I've usually ended up adding to the ones which are complicated. openssl patches are easy cherry-picks and has bit of information in the commits so did not add them.

dev-libs/openssl/openssl-1.1.1k.ebuild
sayanchowdhury: @pthos: Do you mean there should be another commit, which reverts the patch commit, then sync with Gentoo, and then apply the Flatcar patches?

pothos: No, it can be in one commit but before syncing, remove the existing files (first step in https://github.com/kinvolk/coreos-overlay#updating)

pothos: done

dongsupark (approved):

PR #929, opened 2021-04-02
Enable MDIO_BCM_UNIMAC for arm64

wernerb:
Fixes  https://github.com/kinvolk/Flatcar/issues/375

# How to use

This flag is missing for BCM so the MAC can get reused after already being used by other stages such as iPXE. You can run flatcar with iPXE as noted in the original issue 

# Testing done
While the flag should work I can't test it because it is not included due to another upstream bug related to dracut and linux. Looking at portage-stable flatcar is running dracut-050. At least version 052 is needed to fix this. See https://github.com/dracutdevs/dracut/pull/1020/commits/c8be886786d5a28d711ef37532b9e9f88a91f8dd


jepio (approved):
`CONFIG_BCMGENET` specifies `select MDIO_BCM_UNIMAC` so this setting was already active before, but it is a good idea to be explicit about what we want enabled. I'll look into the dracut update next.

PR #958, opened 2021-04-19
coreos-base/oem-gce: use systemd-nspawn to run the GCE daemon

pothos:
The GCE daemon container was run with rkt from an ACI tar ball.
To replace rkt with systemd-nspawn, extract the tar ball to an image
and run the daemon as systemd-nspawn container.

# How to use

For the release notes:
```
- GCE: The oem-gce.service was ported to use systemd-nspawn instead of rkt.
   A one-time action is required to fetch the new service file because the OEM partition is not updated:
    sudo curl -s -S -f -L -o /etc/systemd/system/oem-gce.service https://raw.githubusercontent.com/kinvolk/coreos-overlay/fe7b0047ef5b634ebe04c9627bbf1ce3008ee5fa/coreos-base/oem-gce/files/units/oem-gce.service && sudo systemctl daemon-reload && sudo systemctl restart oem-gce.service
```

# Testing done

Checked that service runs on a GCE instance with [this image](http://jenkins.infra.kinvolk.io:8080/job/os/job/board/job/packages-matrix/3749/console):
```
systemctl status oem-gce
● oem-gce.service - GCE Linux Agent
     Loaded: loaded (/etc/systemd/system/oem-gce.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2021-04-22 08:29:43 UTC; 1min 1s ago
```

### DIFF
--- a/coreos-base/oem-gce/files/units/oem-gce.service
+++ b/coreos-base/oem-gce/files/units/oem-gce.service
@@ -4,6 +4,7 @@ After=local-fs.target network-online.target
 
 [Service]
 Type=notify
+NotifyAccess=all
 Restart=always
 RestartSec=5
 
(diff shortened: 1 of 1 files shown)


t-lo (approved):
LGTM, CI / nightlies will show whether there are any issues with GCP. Let's get it merged!

sayanchowdhury (approved):

pothos:
This completed a CI test (unrelated k8s test failure) and a manual check that the service runs

PR #976, opened 2021-04-26
sys-apps/systemd: Upgrade to systemd v248

sayanchowdhury:
# sys-apps/systemd: Upgrade to systemd v248

# How to use

```
emerge-amd64-usr sys-apps/systemd
./build_packages
```

# Testing done

Jenkins CI [running](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2514/)


pothos (changes requested):
Since this package is based on the Gentoo package, move this folder to a temporary location, then copy over the Gentoo portage folder in one commit (can amend the existing one), and then move the saved folder back and commit this change as "Apply Flatcar patches".

pothos:
I think we should update the ebuild file not only to reduce the downstream changes like `sd_*` vs `meson_*` naming but also to have the correct ebuild dependencies modeled (e.g., the use flags for http). Also there is this `inherit` change which I'm not sure about and much more.

sayanchowdhury:
Closing this PR as we directly migrated to v249

PR #1023, opened 2021-05-21
net-misc/openssh: update to 8.6

dongsupark:
Update openssh to the latest version 8.6, mainly to support [FIDO/U2F authentication](https://www.openssh.com/releasenotes.html#8.2).
Due to recent security issues, we should avoid old openssh versions like 8.2 ~ 8.5.

Sync with Gentoo, make ebuilds accept both keywords `~amd64` and `~arm64`.

Apply Flatcar changes like:
* Drop the init.d files.
* Remove the socket unit's rate limiting.

Instead of dropping bindist, enable it with the profiles now so it doesn't need to be modified on future updates.

Imported from commit [6c0c1c8806be](https://github.com/kinvolk/coreos-overlay/commit/6c0c1c8806bedcc164e5bd3541ab50b2c21e2498).

Fixes https://github.com/kinvolk/Flatcar/issues/395 .

## How to use

```
emerge-amd64-usr net-misc/openssh
```

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2673/cldsv


pothos (approved):
Thanks, looks good

PR #1042, opened 2021-06-01
sys-apps/baselayout: filter out bullet point when parsing failed units (for flatcar-2765)

pothos:
The output of "systemctl list-units --state=failed --no-legend" still
contains the bullet point which is not expected and ends up being taken
as the unit name which was previously on the start of the line.
Filter the bullet point out to stay compatible with the old behavior in
case upstream would remove the bullet point again.

This pulls in
https://github.com/kinvolk/baselayout/pull/16/

### DIFF
--- a/sys-apps/baselayout/baselayout-9999.ebuild
+++ b/sys-apps/baselayout/baselayout-9999.ebuild
@@ -9,7 +9,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="abbf6b604659f780846a881a6c6bafbcd8e2da87" # flatcar-master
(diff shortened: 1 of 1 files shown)


sys-apps/baselayout/baselayout-9999.ebuild
```diff
@@ -9,7 +9,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="abbf6b604659f780846a881a6c6bafbcd8e2da87" # flatcar-master
+	CROS_WORKON_COMMIT="003645f150af7a331307a3a472ae1f7f1014b0ba" # flatcar-master
```
tormath1, line 12: just asking - why don't we directly pull the latest commit (`bbe3f2589c46f61a2c9628ab5dd19e31ff651477`) ?

pothos: I don't want to introduce the other changes that go through a regular Alpha→Beta→Stable transition

tormath1 (approved):

PR #1067, opened 2021-06-21
Update to glibc 2.33

t-lo:
What it says on the tin :)

Dependency PRs: 
* portage-stable: https://github.com/kinvolk/portage-stable/pull/177
* flatcar-scripts: https://github.com/kinvolk/flatcar-scripts/pull/126


pothos (changes requested):
Looks good, but the downstream commit is not self-contained and references files introduced by other commits which makes it difficult to see all downstream changes. Please remove all the directory contents first when importing a new Gentoo version and then let the downstream commit create the files again. While doing so we could also think about dropping old ebuild files – I don't have a strong opinion in general but dropping them would make the downstream commit smaller because it would only patch the actually used new ebuild file.

pothos:
Following https://github.com/kinvolk/coreos-overlay#updating I would align the whole folder with upstream in the first commit (and drop unused ebuild files) and in the second one bring in the downstream changes:

[0001-sys-libs-glibc-Sync-from-Gentoo.patch.txt](https://github.com/kinvolk/coreos-overlay/files/6726436/0001-sys-libs-glibc-Sync-from-Gentoo.patch.txt)
[0002-sys-libs-glibc-Apply-Flatcar-patches.patch.txt](https://github.com/kinvolk/coreos-overlay/files/6726437/0002-sys-libs-glibc-Apply-Flatcar-patches.patch.txt)

t-lo:
Good point, and this will also create a reference point for future glibc updates. Will update the PR.

t-lo:
I've tried my best to align this PR with the new process in https://github.com/kinvolk/coreos-overlay#updating :  
- https://github.com/kinvolk/coreos-overlay/pull/1067/commits/f85720fb019a060ab88b51b3372817a32e0dcf79 removes glibc-2.29, restores glibc-2.32 to upstream default, and addes upstream 2.33
- https://github.com/kinvolk/coreos-overlay/pull/1067/commits/f05e84f335ecad01fe9163a8738c37dee9ed4966 adds Flatcar specific changes to both 2.32 and 2.33.

pothos:
Ah, do we need both 2.32 and 2.33 at the same time? I thought only one is needed and thus dropped the old file.

dongsupark:
According to a Gentoo maintainer, we would need to have glibc [2.33-r1](https://bugs.gentoo.org/792261#c4), in order to address CVE-2021-33574.
2.33-r1 has `PATCH_VER=4`, while 2.33 has `PATCH_VER=3`. 
(So difficult to track, very easy to be confused. I know)

pothos (changes requested):
The first commit should delete the `files/nscd-conf.tmpfiles` and the file should be created by the  second commit.

t-lo:
Addressed all the above, thanks for your input!

t-lo:
I agree we should, and I will (maybe with the Alpha release after next?). I'm just cautious for now since this is only my 2nd glibc upgrade :smiley_cat:

sayanchowdhury (approved):

sys-libs/glibc/glibc-2.33-r1.ebuild
```diff
@@ -1,40 +1,43 @@
...
+# Gentoo patchset (ignored for live ebuilds)
+PATCH_VER=4
+PATCH_DEV=dilfridge
+
 if [[ ${PV} == 9999* ]]; then
-	EGIT_REPO_URI="https://sourceware.org/git/glibc.git"
 	inherit git-r3
 else
-	KEYWORDS="~alpha amd64 arm arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sh ~sparc x86"
+	KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86"
```
dongsupark, line 26: Now the glibc 2.33-r1 ebuild file has keywords `~amd64` and `~arm64`.
We need to either make it accept both `~amd64` or `~arm64` for glibc 2.33-r1, or change the line directly in the ebuild file.
Otherwise, glibc 2.33-r1 will be simply ignored, and glibc 2.32-r2 will be installed.

dongsupark (approved):
Looks good. Thanks!

PR #1079, opened 2021-06-29
net-firewall/{ip,eb,nf}tables: Sync with Gentoo upstream; iptables 1.8.7; ebtables 2.0.11; nftables 0.9.9

sayanchowdhury:
# net-firewall/iptables: Sync with Gentoo upstream; iptables 1.8.7

## How to use

```
emerge-amd64-usr net-firewall/iptables
```

## Testing done

~~CI Passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2837/cldsv/~~

## Update

~CI Passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3083/cldsv/~
CI Passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3587/cldsv/

To be merged with https://github.com/kinvolk/portage-stable/pull/188


tormath1 (approved):
LGTM, thanks ! 
Out of curiosity, do you think it would be relevant to provide kola tests for this package ? with basic / common iptables rules ?

dongsupark:
You would probably want to create another PR in portage-stable as well.
Without that, build will simply fail.

net-firewall/iptables/iptables-1.8.7.ebuild
```diff
@@ -43,8 +43,6 @@ BDEPEND="${BUILD_DEPEND}
 RDEPEND="${COMMON_DEPEND}
 	${BUILD_DEPEND}
 	nftables? ( net-misc/ethertypes )
-	!<net-firewall/ebtables-2.0.11-r1
-	!<net-firewall/arptables-0.0.5-r1
```
dongsupark: A simple comment to say why those are needed, like "not needed as nftables is disabled by default".

net-firewall/iptables/iptables-1.8.7.ebuild
```diff
@@ -123,9 +121,9 @@ src_install() {
 		rm "${ED}"/sbin/{arptables,ebtables}{,-{save,restore}} || die
 	fi
 
-	systemd_dounit "${FILESDIR}"/systemd/iptables-{re,}store.service
+	systemd_dounit "${FILESDIR}"/systemd/iptables{,-{re,}store}.service
 	if use ipv6 ; then
-		systemd_dounit "${FILESDIR}"/systemd/ip6tables-{re,}store.service
+		systemd_dounit "${FILESDIR}"/systemd/ip6tables{,-{re,}store}.service
```
dongsupark: short comments would be great

dongsupark (approved):
Looks good in general.

So the USE flag `nftables` is disabled by default.
To address https://github.com/kinvolk/Flatcar/issues/394, we would need some extra work to enable the flag in the future, right?

Anyway updating only the version to 1.8 alone is a huge path forwards. :+1:

pothos:
In the release notes we should link to https://wiki.nftables.org/wiki-nftables/index.php/Troubleshooting#Question_4._How_do_nftables_and_iptables_interact_when_used_on_the_same_system.3F because we switch the default iptables binary to use the nftables subsystem. The concurrent usage of both subsystems is common as each container may use a different subsystem but it's worth pointing out what happens in this case to ease debugging.

Users can explicitly use the legacy binaries on the host to interact with the iptables subsystem, or even set up a bind mount to point the iptables binaries to the legacy binaries, if they want to continue to deal primarily with the iptables subsystem in the mean time but of course it's best to migrate all containers to the newer subsystem if one prefers to only have a single subsystem being used.

net-firewall/iptables/iptables-1.8.7.ebuild
```diff
@@ -135,22 +136,19 @@ src_install() {
...
-			if ! eselect ${tables} show &>/dev/null; then
-				elog "Current ${tables} implementation is unset, setting to ${default_iptables}"
-				eselect ${tables} set xtables-nft-multi
-			fi
-		done
+   	if use nftables; then
+		if ! eselect ebtables show &>/dev/null; then
+			elog "Current ebtables implementation is unset, setting to ${default_iptables}"
+			eselect ${tables} set "${default_iptables}"
+		fi
```
krnowak: You probably could reduce the diff to a single line by replacing `{arp,eb}tables` to `ebtables` in the `for` loop.

net-firewall/iptables/iptables-1.8.7.ebuild
```diff
@@ -19,6 +19,7 @@ IUSE="conntrack ipv6 netlink nftables pcap static-libs"
 BUILD_DEPEND="
 	>=app-eselect/eselect-iptables-20200508
 "
+
```
krnowak, line 22: This line is not necessary.

net-firewall/iptables/iptables-1.8.7.ebuild
```diff
@@ -41,10 +42,7 @@ BDEPEND="${BUILD_DEPEND}
 	)
 "
 RDEPEND="${COMMON_DEPEND}
-	${BUILD_DEPEND}
```
krnowak: Why dropping it? I'd try to explain it either here or in a readme file in this directory.

net-firewall/iptables/iptables-1.8.7.ebuild
```diff
@@ -115,17 +113,20 @@ src_install() {
 		newconfd "${FILESDIR}"/ip6tables-r1.confd ip6tables
 	fi
 
+
```
krnowak, line 117: Unnecessary change.

net-firewall/nftables/nftables-0.9.9.ebuild
```diff
@@ -114,66 +115,3 @@ src_compile() {
...
-
-	systemd_dounit "${FILESDIR}"/systemd/${PN}-restore.service
-
-	if use python ; then
-		python_foreach_impl python_make install
-		python_foreach_impl python_optimize
-	fi
-
-	find "${ED}" -type f -name "*.la" -delete || die
-}
```
krnowak: I'd probably keep the `src_install()` function, but still reduced to running default and dropping the libtool rubbish. Also, why is `systemd_dounit "${FILESDIR}"/systemd/${PN}-restore.service` dropped too? Do we not need it?

```bash
src_install() {
	default
	systemd_dounit "${FILESDIR}"/systemd/${PN}-restore.service # maybe?
	find "${ED}" -type f -name "*.la" -delete || die
}
```

krnowak (changes requested):
Did you make any changes to `net-firewall/ebtables`? There is no "Apply Flatcar modifications" commit for it. Also, please squash the fixup commit - for this kind of PRs it's easier to review it commit by commit.

net-firewall/iptables/iptables-1.8.7.ebuild
krnowak: There is a comment about it in the fixup commit. I haven't noticed that when I reviewed this commit.

net-firewall/nftables/nftables-0.9.9.ebuild
sayanchowdhury: We had the save/restore operations getting dropped in previously as well, so I went ahead and dropped it. 
https://github.com/coreos/coreos-overlay/commit/71d1bd88b493e4e1b2383996d5aa6832aa3e843e

net-firewall/iptables/iptables-1.8.7.ebuild
sayanchowdhury: Ack

net-firewall/iptables/iptables-1.8.7.ebuild
sayanchowdhury: Hmm, I actually had that previously but dropped it to make it simpler :) I could revert to use the `for` loop. I'll add a comment too that we have the for loop to make the diff less to Gentoo upstream.

net-firewall/iptables/iptables-1.8.7.ebuild
sayanchowdhury: Ack

net-firewall/iptables/iptables-1.8.7.ebuild
krnowak: Is there a reason for dropping these blocks?

net-firewall/nftables/Manifest
```diff
@@ -1 +1,2 @@
-DIST nftables-0.6.tar.gz 252523 SHA256 85dd7fa4e741c0be02efddbc57b5d300e1147f09ec6f81d0399110f96dc958f0 SHA512 17f3b94687865e077dc082cf61b29ab2854fd1ffe18212a8d424f2876aef8db9780dd4d06dca8e6d093498151d47bab73e40e1f54062a83a23a3cbe75f27e921 WHIRLPOOL d15eaf81426d73bea28752f96727d291120120fb2aaa994d421d900974eb45062957435e077664fb916780f636ed9b61889dbec8b627d5d309512bae96f02874
+DIST nftables-0.9.8.tar.bz2 879516 BLAKE2B 5063090d648668f4d5ae6d4be48ebecc65dfd4b525768e94a0d90ceebbee73874c916727be8de633550db71c612d698d88cf93575931362b48d954e6ac275143 SHA512 1c5709825c8b2c13cbed0310658959ecee164c930bc9e2447618a0894598138b9a549d20509c32a5c23ce99e40438df38f9e170cf656ce993d819f365490a180
```
krnowak, line 1: Two things here:

- typo in commit ("upstrea" instead of "upstream")
- we don't need this entry in manifest, do we?

net-firewall/iptables/iptables-1.8.7.ebuild
```diff
@@ -135,18 +138,20 @@ src_install() {
...
+	local default_iptables="xtables-nft-multi"
 	if ! eselect iptables show &>/dev/null; then
 		elog "Current iptables implementation is unset, setting to ${default_iptables}"
 		eselect iptables set "${default_iptables}"
 	fi
-
-	if use nftables; then
+	# Flatcar: Drop the arptables, but retain the `for` structure in favor of lesser diff
+	# to upstream
+   	if use nftables; then
```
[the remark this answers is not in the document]
sayanchowdhury: Should be fixed in the fixup commit.

krnowak (approved):
Looks good.

net-firewall/iptables/iptables-1.8.7.ebuild
krnowak: Sounds good.

PR #1080, opened 2021-06-30
sys-libs/glibc: Revert "Update to glibc 2.33"

sayanchowdhury:
The glibc-2.33-r1 build fails with "no space left on device", so reverting kinvolk/coreos-overlay#1067

Needs to be merged with: https://github.com/kinvolk/portage-stable/pull/182


tormath1 (approved):
:rocket:

t-lo:
I think a better path forward would be to just mask ARM64 and AMD64 in the glibc-2.33 ebuild. That way the removal of the (now outdated) 2.29 ebuild is retained.

t-lo:
Alternative PR: https://github.com/kinvolk/coreos-overlay/pull/1081

sayanchowdhury:
Agreed, that would be a much better approach. Thanks! I'll close this PR.

PR #1129, opened 2021-07-22
.github: list only ebuilds with a unique VERSION_OLD for runc

dongsupark:
Before passing runc versions to `sed '/-/!{s/$/_/}'`, we need to replace `_` with `-`, because runc tarball files already have names like `1.0.0_rc2`.
Without the fix, version sort would `1.0.0` come before `1.0.0_rc2`, which is not expected in the later steps.

Due to unnecessary wildcard listings, ebuild files including all rc or beta are being listed.
Since `VERSION_OLD` is already generated as a unique version, we do not need to list multiple files to filter by running `head -n1` etc. We just need to use only the specific ebuild.
Simply list only the unique ebuild file.


dongsupark:
will merge soon to prevent another invalid PR from being created

jepio (approved):

PR #1130, opened 2021-07-23
enable arm64 SDK bootstrap

jepio:
# enable arm64 SDK bootstrap

This PR adjusts profiles and ebuild keywords to enable arm64 SDK to be bootstrapped. There is some stuff that needed fixing in the rust ebuild as well.

## How to use

Provision powerful arm64 machine.
Fetch kinvolk/flatcar-scripts#134
Fetch stage4 from https://github.com/kinvolk/Flatcar/issues/319#issuecomment-881553797.
`./bootstrap_sdk --seed_tarball=...`


dongsupark (commented):
Looks good. I managed to build an arm64 SDK from the PRs.
Though I am not sure when we should merge the PRs, as we could start building another set of releases soon.
I will let @sayanchowdhury share some thoughts.

jepio:
Merging this one before the release as well, I verified that this doesn't break amd64.

PR #1136, opened 2021-07-27
sys-kernel/coreos-modules: switch to gzip/zstd compression

pothos:
sys-kernel/coreos-modules: switch to gzip/zstd compression
    
    With the default gzip compression the 60 MB limit for the vmlinuz
    bundle of kernel+initramfs was reached. The limit comes from the size
    of the /boot partition which is 128 MB large and the kernel needs to
    fit twice, in addition to GRUB.
    
    Use zstd for the initramfs as it provides a similar speed but better
    compression. For the kernel we can't switch yet to zstd for arm64
    but for amd64 it works.

Fixes https://github.com/kinvolk/Flatcar/issues/438

## How to use/testing done

With zstd for both on amd64 I got a reduction of 13 MB for the vmlinuz file (new size is 49 MB, Alpha 2920.0.0 is 62 MB). Since that didn't work with arm64 I changed arm64 to continue to use gzip (default) for the kernel and zstd for initramfs, this just just a 5 MB reduction (new size is 54 MB, old is 59 MB).

http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3138/cldsv/


jepio (approved):
Looks good

dongsupark (approved):
Looks good.

Side note: I was wondering if we should do the same thing for arm64.
However, apparently arm64 still needs uncompressed Kernel, according to the discussion. https://bugs.gentoo.org/784599

pothos:
Thanks for the correction, I assumed it uses GZIP and was confused why switching to something similarly supported didn't work

pothos:
Ported to 2905 and 2942 maintenance branches

PR #1137, opened 2021-07-27
upgrade coreos profile 17.1

jepio:
# upgrade coreos profile to 17.1

This upgrades the coreos profile to depend on Gentoo 17.1. The 17.0 profiles have been deprecated some time ago. The only difference between 17.0. and 17.1 profiles is that `lib` no longer links to `lib64`. This is already not the case for arm64.

Fixes kinvolk/Flatcar#428
Requires kinvolk/flatcar-scripts#136
Requires kinvolk/flatcar-scripts#134

## How to use

Full bootstrap required. Starting from a 2920.0.0 SDK, stage1 needs to be brought up with the correct profile:
```
sudo ./bootstrap_sdk --stage1_overlay_ref jepio/upgrade-profile-17.1-backport-2920
```

## Testing done

SDK successfully bootstrapped. Will test the rest through CI.

### DIFF
--- a/profiles/coreos/amd64/parent
+++ b/profiles/coreos/amd64/parent
@@ -1,2 +1,2 @@
-portage-stable:default/linux/amd64/17.0/no-multilib/hardened
+portage-stable:default/linux/amd64/17.1/no-multilib/hardened
 :coreos/base


jepio:
CI running http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3140/

dongsupark:
Did CI pass? I cannot see the result.
It looks ok if CI passed.

jepio:
So bootstrapping this is sort of an issue - the CI currently does update_chroot before running `bootstrap_sdk` and that doesn't work.

Let's talk about this with the team tomorrow to figure out how to go about this.

jepio:
What I mean is: `./bootstrap_sdk` work as long as the profile is only changed after running `./setup_board` or `./update_chroot`.

jepio:
I think I found a way to get this to pass CI, currently running: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3371/.

jepio:
Closing this for now - the change itself is not the issue, its figuring out how to make it work with catalyst and the bootstrap process. I don't expect to have time for this any time soon.

PR #1145, opened 2021-07-29
coreos-base/coreos-init: prevent networkd interference with cilium_vxlan

pothos:
This pulls in
https://github.com/kinvolk/init/pull/43

**Note:** To be ported to all channels, either with introducing separate branches for `init` or by bringing the following changes along:
https://github.com/kinvolk/init/pull/41
https://github.com/kinvolk/init/pull/42


jepio (approved):
I'd say its fine to pull in the other changes into the stabel channels as well.

PR #1146, opened 2021-07-29
coreos-base/oem-gce: grant CAP_NET_ADMIN to set routes for LB

pothos:
With the switch from rkt to systemd-nspawn the ability for the service
to set the routing entries for the TCP load balancer got lost,
resulting in an unreachable LB as reported in
https://github.com/kinvolk/Flatcar/issues/459

The fix also reported there is to retain CAP_NET_ADMIN when starting
the service.

## Testing done

Regular release tests: http://localhost:9091/job/os/job/manifest/3159/cldsv/


jepio (approved):

pothos:
Picked for 2942, 2920, 2905 (Alpha, Beta, Stable)

PR #1150, opened 2021-07-30
dev-libs/expat: Sync with Gentoo upstream; updates to 2.4.1 

sayanchowdhury:
# dev-libs/expat: Sync with Gentoo upstream; updates to 2.4.1 

## How to use

```bash
emerge-amd64-usr dev-libs/expat
```

## Testing done

Jenkins CI is running, ~http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3171/cldsv/~
http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3189/cldsv/


sayanchowdhury:
CI Passed. Ready for review.

jepio (dismissed):
Do i understand correctly that usr_ldscript is only used for freebsd and that's why it has been removed? If that's everything then looks good.

krnowak:
Would it be a problem to just move it back to portage-stable? The only differences between upstream and our ebuild are:

1. We do not import usr-ldscript
2. We dropped this part: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/expat/expat-2.4.1.ebuild#n77 (which should be a no-op for us).

sayanchowdhury:
> Would it be a problem to just move it back to portage-stable? The only differences between upstream and our ebuild are:
> 
>     1. We do not import usr-ldscript
> 
>     2. We dropped this part: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/expat/expat-2.4.1.ebuild#n77 (which should be a no-op for us).

CoreOS folks dropped the usr-ldscript inherit, which much information in the commit about why? (Given back then too it seemed that usr-ldscript was mostly non-invasive) https://github.com/coreos/coreos-overlay/commit/b39dd320ca6670ee491ff6d161e2d4189ff11e48

I'm +1 to move to portage-stable, but given we would like to trigger the Alpha release. I could create another PR with the move and appropriate testing, on if there are repercussions. Thoughts?

krnowak (approved):
Alright, let's do the moving later to avoid blocking the release any further.

PR #1162, opened 2021-08-03
amazon-ssm-agent: re-add dependency to oem-ec2-compat

jepio:
# amazon-ssm-agent: re-add dependency to oem-ec2-compat

The binaries were too big to fit on our 128MB OEM partition. Now that the OEM partition is compressed-btrfs it fits again, so re-add the dependency to oem-ec2-compat.

The other half of the PR are cross-compilation fixes. The ebuild was not exporting GOARCH and was making use of latest go. When built for arm64, host binaries were produced. The ebuild has been fixed to cross-compile correctly.

Related: https://github.com/kinvolk/Flatcar/issues/107

## How to use

```
emerge-arm64-usr amazon-ssm-agent
./build_image
./image_to_vm --format=ami
```

## Testing done

Build tested, and tested that the binaries fit on the OEM partition.

### DIFF
--- a/app-emulation/amazon-ssm-agent/amazon-ssm-agent-2.3.1319.0-r1.ebuild
+++ b/app-emulation/amazon-ssm-agent/amazon-ssm-agent-2.3.1319.0-r1.ebuild
@@ -28,23 +28,25 @@ src_prepare() {
 }
 
 src_compile() {
+	go_export
+
 	# this is replication of commands from the vendor makefile
 	# but without network activity during build phase
 	local GO_LDFLAGS="-s -w -extldflags=-Wl,-z,now,-z,relro,-z,defs"
 	export GOPATH="${WORKDIR}/${PN}-${PV}"
 	export GO111MODULE="off"
 	# set agent release version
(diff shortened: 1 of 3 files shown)


pothos:
I think we should also try to install it now instead of leaving it unused.
You would move the `RDEPEND` entry in `coreos-base/oem-ec2-compat/oem-ec2-compat-0.1.2.ebuild` out of the comment and revert b6abb59c544be13e923a3e7240b5c9395c281fca to restore the entries in `base-ec2.ign`

jepio:
Sure

jepio:
@pothos I brought back the dependency and I probed the OEM partition of the AMI to see that it's there. Will try to run CI for it.

jepio:
http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3199

pothos:
Nice, thanks, how much space is left on the OEM partition?

jepio:
~85MB

pothos:
Seems the compression pays off, previously it didn't fit

pothos (approved):

jepio:
CI green ✔️

PR #1178, opened 2021-08-09
Update golang update job to update multiple golang versions

krnowak:
I noticed that https://github.com/kinvolk/coreos-overlay/pull/1177 only updates the latest major version of golang, so I updated go 1.15 manually. It's not critical, because we _probably_ don't use 1.15 anyway (grepping for `COREOS_GO_VERSION` didn't give me any results for 1.15), but we better keep both old-stable and stable version up-to-date.

I also dropped a patch for go1.12 - it was there just for rkt, so drop the baggage.

Currently a draft, no testing done yet.


krnowak:
Forgot to mention - this also tries to update github action to be able to upgrade several major versions of golang in one run.

dongsupark:
If we do not use Go 1.15 in `main`, then how about deleting Go 1.15 instead of keeping both versions?

On the other hand, the LTS branch `flatcar-lts-2605` still has Go 1.15.
If we would want to automatically update patchlevel for Go 1.15, we should actually do that in the LTS branch.

krnowak:
I think we could first update it, then backport some portion of this PR to LTS and then drop go-1.15 in main. I'll need to check other branches to see if go-1.15 was used for building anywhere.

dongsupark:
Now that https://github.com/flatcar-linux/coreos-overlay/pull/1304 was merged, I suppose this PR is outdated, isn't it?

krnowak:
Yeah, the parts for updating go1.15 is outdated, but I think that changes to the github action are still useful. I'll update the PR shortly. Any hints on how to actually test the github action, even if manually? With that I could finally move this PR out of draft status.

dongsupark:
> Any hints on how to actually test the github action, even if manually? With that I could finally move this PR out of draft status.

What I usually do is, go to my cloned repo like https://github.com/dongsupark/coreos-overlay , tweak `.github/workflows/go-releases-main.yml`,
```
name: Get the latest Go release for main
on:
  push:
    branches:
      - "main"
```
and push new code to `main` branch .
See also https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration .

dongsupark:
Now that we have merged https://github.com/flatcar-linux/coreos-overlay/pull/1319 , you are able to do the manual test of Github Actions quite easily.
Go to your forked repo, select workflow in the Actions tab, then simply click on "Run workflow".

krnowak:
Tested it by enabling updates of 1.12 and 1.17 versions of go, and downgrading them a bit. Also made sure that nonexistent or equal versions are left untouched. You can see [the test changes I have made](https://github.com/krnowak/coreos-overlay/commit/18a1a07f62f29290819fbbad2683b16deebbd27c), [the github action run](https://github.com/krnowak/coreos-overlay/runs/5794217699?check_suite_focus=true) and [the resulting PR](https://github.com/krnowak/coreos-overlay/pull/13).

dongsupark (approved):
Thanks for the tests.

PR #1179, opened 2021-08-10
bump etcd and flannel to the latest versions to get multiarch images for arm64 support

pothos:
A number of kola tests fail because on arm64 we don't have etcd and flannel working due to using an image tag by default that isn't a multiarch image.
See https://github.com/kinvolk/Flatcar/issues/474

## How to use

The new image tags used here are multiarch images and it should work to pass the kola tests on arm64 with them.

## Testing done

None yet, need to start a test build


dev-db/etcdctl/etcdctl-3.5.0.ebuild
```diff
@@ -0,0 +1,607 @@
...
+SLOT="0"
+KEYWORDS="amd64 arm64"
+IUSE=""
+
+S=${WORKDIR}/etcd-${PV}
+
+inherit coreos-go-utils go-module
+
+# generated by merging /go.sum and /etcdctl/go.sum
+EGO_SUM=(
```
tormath1, line 18: Following the `eclass` doc, we can omit the hash (https://devmanual.gentoo.org/eclass-reference/go-module.eclass/index.html)

> For inclusion in EGO_SUM, the h1: value and other future extensions SHOULD be omitted at this time. The EGO_SUM parser will accept them for ease of ebuild creation.


```
sed -i 's/ h1.*$/"/' dev-db/etcdctl/etcdctl-3.5.0.ebuild
```

tormath1:
@pothos for the failing tests `Error: unknown command "cluster-health" for "etcdctl"`, we can first start by prefixing the `etcdctl` command with `ETCDCTL_API=2` in order to keep using V2 APIs.

Once merged, we can open an issue to track everything related to `etcd/v3` upgrade with:
 - kola test update
 - locksmithctl update (already in progress)
 - ...

PR #1196, opened 2021-08-16
Compile Azure OEM contents for all boards

pothos:
While not used at the moment we still can lift the restriction to only
compile for the amd64 board.

## How to use/Testing done

Build all packages and run the `image_to_vm.sh` script for Azure


jepio (approved):
Looks good, no harm in enabling it. The same could be done for open-vm-tools at some point, since this exists: https://flings.vmware.com/esxi-arm-edition

PR #1200, opened 2021-08-18
app-emulation/wa-linux-agent: Update to 2.3.1.1

krnowak:
This version has an officially documented support for python3, so it
plays along our plans of removing python2 in favor of python3. When
the switch actually happens, we will need to update the ebuild to
mention the correct path to python modules. The path contains python
version, which is a hindrance. Would be nice to have it hidden behind
some variable.

There is also a version 2.4.0.2, but it's marked as a prerelease on
github, so decided to package 2.3.1.1 instead.

CI: http://localhost:9091/job/os/job/manifest/3324/cldsv/

Is there a way to test it out?


pothos:
You can add `azure` to the Jenkins FORMAT_LIST_AMD64 parameter

krnowak:
Cool, thanks. I'll restart the CI then.

krnowak:
CI passed. The failure in os/kola/azure doesn't seem to be related (`cl.etcd-member.discovery` test failed). The other os/kola/azure job succeeded.

dongsupark (approved):

PR #1208, opened 2021-08-19
sys-apps/ignition: Fix null pointer deref crash for OEM FS

pothos:
This pulls in
https://github.com/kinvolk/ignition/pull/25

## Testing done

Ongoing: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3333/


jepio:
Can we add a kola test for this?

pothos:
Good idea, it needs to be for a platform where OEM partitions exist (QEMU tests run on the generic image).

jepio:
I suggest we merge this one before https://github.com/kinvolk/coreos-overlay/pull/1198 as I expect we will want to backport this to alpha?

pothos:
Thanks, that simplifies things.

PR #1213, opened 2021-08-23
x11-drivers/nvidia-drivers: Add runtime dependencies

krnowak:
When I was renaming the coreos-base/gmerge package, I noticed that nvidia drivers is using a bunch of stuff at runtime (in `setup-nvidia`), but nothing was specified in RDEPENDS. I felt like it should be filled, but not sure.

@sayanchowdhury : Does it make sense, before trying to run this thing in CI?


sayanchowdhury:
This does make sense, given they are dependencies it's good to have them in RDEPEND.

PR #1214, opened 2021-08-23
explicitly set 'runtime_type' in containerd config

jepio:
# explicitly set 'runtime_type' in containerd config

The way the SystemdCgroup parameter was being set resulted in 'runtime_type' being unset, which prevents containerd from being used as the kubelet container runtime. Explicitly set in 'runtime_type' to fix the issue.

Fixes https://github.com/kinvolk/Flatcar/issues/484

## How to use

Rebuild image and deploy kubernetes with containerd as the runtime.

## Testing done

CI pending


tormath1:
@jepio reading the `containerd` config; there is no risk to have other unset options ? :)

jepio:
I checked two things:
```
$ containerd -c before.toml config dump >a
$ containerd -c after.toml config dump >b
$ diff -up a b
--- a   2021-08-24 07:52:17.000000000 +0000
+++ b   2021-08-24 07:52:13.000000000 +0000
@@ -1,5 +1,5 @@
 disabled_plugins = []
-imports = ["before.toml"]
+imports = ["after.toml"]
 oom_score = -999
 plugin_dir = ""
 required_plugins = []
@@ -99,17 +99,7 @@ version = 2
           runtime_type = "io.containerd.runc.v2"

           [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
-            BinaryName = ""
-            CriuImagePath = ""
-            CriuPath = ""
-            CriuWorkPath = ""
-            IoGid = 0
-            IoUid = 0
-            NoNewKeyring = false
-            NoPivotRoot = false
-            Root = ""
-            ShimCgroup = ""
-            SystemdCgroup = false
+            SystemdCgroup = true

       [plugins."io.containerd.grpc.v1.cri".containerd.untrusted_workload_runtime]
         base_runtime_spec = ""

```
So the other fields have default zero values that will be kept.
Kops uses the same configuration for SystemdCgroup: https://github.com/kubernetes/kops/blob/master/nodeup/pkg/model/containerd.go#L433-L462

jepio:
CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3357/

sayanchowdhury (approved):

PR #1217, opened 2021-08-24
switch docker pkgs to go1.16

jepio:
# switch docker pkgs to go1.16

Docker upstream started building with go1.16, so we should do the same. If we also get rid of old versions of
container/docker-runc/docker-proxy ebuilds, we can remove go1.13 and go1.7 from the tree.

## How to use

```
./build_package
./build_image
````

## Testing done

CI running: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3369/

### CODE REVIEW

**Comment by jepio** [comment]
CI passed except for `cl.etcd-member.v2-backup-restore`, which is unrelated and got fixed here: https://github.com/kinvolk/mantle/pull/215.

**Comment by tormath1** [comment]
@jepio I restarted the tests only for `cl.etcd-member.v2-backup-restore` :)

File: app-emulation/docker-cli/docker-cli-20.10.8.ebuild [hunk 1]
```diff
@@ -34,7 +34,6 @@ src_prepare() {
 src_compile() {
 	# Flatcar: override go version
```
tormath1, line 35: is this comment still relevant ?

jepio: Yes, it refers to `go_export` configuring `PATH` and other environment variables in a way that `go` refers to a fixed go version, specified by COREOS_GO_VERSION. This decouples us from using the latest go version. We will soon add go1.17 to the tree.

PR #1218, opened 2021-08-24
dev-libs/openssl: Sync with Gentoo upstream; bump to 1.1.1l

sayanchowdhury:
# dev-libs/openssl: Sync with Gentoo upstream; bump to 1.1.1l

## How to use
```bash
emerge-amd64-usr openssl
```

## Testing done

Jenkins CI test not done. Locally tested.


dev-libs/openssl/openssl-1.1.1l.ebuild
```diff
@@ -27,9 +27,9 @@ SRC_URI="mirror://openssl/source/${MY_P}.tar.gz
 LICENSE="openssl"
 SLOT="0/1.1" # .so version of libssl/libcrypto
 [[ "${PV}" = *_pre* ]] || \
-KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86 ~x86-linux"
-IUSE="+asm bindist elibc_musl rfc3779 sctp cpu_flags_x86_sse2 sslv3 static-libs test tls-compression tls-heartbeat vanilla"
-RESTRICT="!bindist? ( bindist )
+KEYWORDS="~alpha amd64 arm arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86 ~x86-linux"
```
krnowak, line 30: Do we care about `arm` keyword?

dev-libs/openssl/openssl-1.1.1l.ebuild
```diff
@@ -27,9 +27,9 @@ SRC_URI="mirror://openssl/source/${MY_P}.tar.gz
...
+IUSE="+asm bindist elibc_musl rfc3779 sctp cpu_flags_x86_sse2 sslv3 static-libs test tls-compression tls-heartbeat vanilla zlib"
```
krnowak, line 31: Why adding `zlib` back? I think it got replaced with `tls-compression` use flag.

dev-libs/openssl/openssl-1.1.1l.ebuild
```diff
@@ -47,7 +47,6 @@ PDEPEND="app-misc/ca-certificates"
 
 PATCHES=(
 	"${FILESDIR}"/${PN}-1.1.0j-parallel_install_fix.patch #671602
-	"${FILESDIR}"/${PN}-1.1.1i-riscv32.patch
```
krnowak: Why dropping it? Let's keep the diff minimal as possible.

dev-libs/openssl/openssl-1.1.1l.ebuild
```diff
@@ -288,8 +287,12 @@ multilib_src_install_all() {
...
 	dodoc CHANGES* FAQ NEWS README doc/*.txt doc/${PN}-c-indent.el
 
-	# create the certs directory
-	keepdir ${SSL_CNF_DIR}/certs
+	# This is crappy in that the static archives are still built even
+	# when USE=static-libs.  But this is due to a failing in the openssl
+	# build system: the static archives are built as PIC all the time.
+	# Only way around this would be to manually configure+compile openssl
+	# twice; once with shared lib support enabled and once without.
+	use static-libs || rm -f "${ED}"/usr/lib*/lib*.a
```
krnowak: This already seems to be a part of `multilib_src_install`.

dev-libs/openssl/metadata.xml
```diff
@@ -8,7 +8,6 @@
 <use>
  <flag name="asm">Support assembly hand optimized crypto functions (i.e. faster run time)</flag>
  <flag name="bindist">Disable/Restrict EC algorithms (as they seem to be patented) -- note: changes the ABI</flag>
- <flag name="ktls">Enable support for Kernel implementation of TLS (kTLS)</flag>
```
krnowak: I guess you could keep this line. I know it's out of date, but that would minimize the diff with upstream.

dev-libs/openssl/openssl-1.1.1l.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI="7"
 
-inherit flag-o-matic toolchain-funcs multilib-minimal
+inherit flag-o-matic toolchain-funcs multilib multilib-minimal systemd
```
krnowak: We probably don't need to inherit the multilib eclass.

krnowak (changes requested):

PR #1219, opened 2021-08-25
Update dev-lang/python-oem to 3.6.5

krnowak:
This updates python used by oem stuff to the same version we use in SDK - 3.6.5. Later, when we update python3 to newer version, we could update both at the same time. Also try documenting the changes we have made here.

Currently making it a draft, because os/kola/azure tests seem to have problems passing. I don't think it's related to this PR, but I want them to be able to run, since azure seems to be the primary user of python.

CI: http://localhost:9091/job/os/job/manifest/3467/cldsv/


krnowak:
CI mostly passed - arm64 build failed because of #1242, but we don't have azure images for arm64 anyway.

app-emulation/wa-linux-agent/wa-linux-agent-2.3.1.1.ebuild
```diff
@@ -12,15 +12,23 @@ LICENSE="Apache-2.0"
 SLOT="0"
 IUSE=""
 
-RDEPEND="dev-lang/python-oem"
+RDEPEND="
+dev-lang/python-oem
```
jepio, line 16:
```suggestion
dev-lang/python-oem:3.6
```
so that the "remember to update" comment becomes a build time failure.

dev-python/distro-oem/distro-oem-1.6.0.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+DESCRIPTION="OS platform information API"
+HOMEPAGE="https://github.com/python-distro/distro"
+SRC_URI="${HOMEPAGE}/releases/download/v${PV}/${MY_P}.tar.gz"
+
+LICENSE="Apache-2.0"
+KEYWORDS="amd64 arm64"
+
+SLOT="0"
+RDEPEND="dev-lang/python-oem"
```
jepio, line 17:
```suggestion
RDEPEND="dev-lang/python-oem:3.6"
```
So that this fails at build time if someone forgets to update.

jepio (approved):

jepio:
Great to see this working - if we find the time later we could also submit the wa-linux-agent patch upstream.

app-emulation/wa-linux-agent/wa-linux-agent-2.3.1.1.ebuild
krnowak: That's a good idea, I'll test it locally.

krnowak:
> Great to see this working - if we find the time later we could also submit the wa-linux-agent patch upstream.

Yeah, it's on my todo list.

app-emulation/wa-linux-agent/wa-linux-agent-2.3.1.1.ebuild
krnowak: That's a neat trick and it works like a charm, thanks! I actually have put `3.6` into the `DEP_PYVER` variable, so I only need to change the number in one place in ebuild.

dev-python/distro-oem/distro-oem-1.6.0.ebuild
krnowak: Updated it too with the `DEP_PYVAR` stuff.

jepio:
Feel free to merge

PR #1232, opened 2021-08-30
coreos-base: hard-host-depends: remove unused go versions 1.7 & 1.8 & 1.13

jepio:
# remove go 1.7,1.8 & 1.13 from SDK dependencies

In https://github.com/kinvolk/coreos-overlay/pull/1215, while grepping for occurences of go 1.13 and 1.7 I missed the slot specifiers in hard-host-depends, which broke the SDK bootstrap. Remove go 1.13 and 1.7 from the dependency list. Additionally, remove go 1.8 which is also not needed.

## Testing done

```
# emerge -uav hard-host-depends
# emerge -ca
...
 dev-lang/go
    selected: 1.7.6
   protected: none
     omitted: 1.8.7 1.12.17 1.15.11 1.16.7
...
 dev-lang/go
    selected: 1.13.15
   protected: none
     omitted: 1.8.7 1.12.17 1.15.11 1.16.7
```


coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
```diff
@@ -39,10 +39,8 @@ RDEPEND="${RDEPEND}
 	dev-lang/nasm
 	dev-lang/swig
 	dev-lang/yasm
-	dev-lang/go:1.7
 	dev-lang/go:1.8
```
dongsupark, line 42: I suppose we can also delete Go 1.8.

dongsupark (commented):
It looks good.
Nit below:

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
jepio: Removed go 1.8 as well.

dongsupark (approved):

PR #1233, opened 2021-08-30
profiles: enable 'audit' use flag for sys-libs/pam

jepio:
# build sys-libs/pam with audit support

This change results in building the pam_tty_audit module additionally, nothing else.
Related to https://github.com/kinvolk/Flatcar/issues/485.

## Testing done

```
./build_packages
./build_image
```

### DIFF
--- a/profiles/coreos/targets/generic/package.use
+++ b/profiles/coreos/targets/generic/package.use
@@ -17,7 +17,7 @@ sys-apps/smartmontools	minimal
 sys-block/parted        device-mapper
 sys-fs/lvm2		-lvm1 -readline
 sys-libs/ncurses	minimal
-sys-libs/pam        -berkdb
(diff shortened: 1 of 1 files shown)


tormath1:
I see we ignore everything from audit rules: https://github.com/kinvolk/coreos-overlay/blob/main/sys-process/audit/files/rules.d/99-default.rules#L4-L5 except for SELinux: https://github.com/kinvolk/coreos-overlay/blob/main/sys-process/audit/files/rules.d/80-selinux.rules#L4.

Maybe we should allow `audit_tty`: https://github.com/torvalds/linux/blob/master/include/uapi/linux/audit.h#L104 ?

jepio:
I kind of expect people to override audit rules when they have specialized needs.

tormath1 (approved):

PR #1234, opened 2021-08-30
dev-vcs/repo: update to 2.8 for python3 (lts)

dongsupark:
The [repo v2.10](https://groups.google.com/g/repo-discuss/c/rpSfMCl83Sk) was released dropping python2 support. As a result, every `repo init` failed to run.
To unblock CI builds, we released mantle [v0.15.2](https://github.com/kinvolk/mantle/releases/tag/v0.15.2), including a workaround to set the target branch to [`maint`](https://gerrit.googlesource.com/git-repo/+/refs/heads/maint), which still supports python2. Now with cork v0.15.2, `cork create` or`cork update` will work well for now.

However, the current state is quite fragile. It will get broken again when the upstream `maint` branch changes. We should update `dev-vcs/repo` in coreos-overlay to 2.x with python3, and get it included in Flatcar SDK, so we could later set the target branch in mantle back to `stable`.

At the moment, none of the source repos has the tarball for repo 2.10, neither GCS nor Gentoo distfiles. So for now we update it to 2.8.
It will be linked to python 3 in Flatcar SDK.

Also note that we do not have to keep `files/repo-1.25` script in thecoreos-overlay repo, because the script is simply identical to the upstream `repo` script.
I am not sure why the third-party script was there in the first place. So simply remove the script.

Also accept keywords `amd64` and `arm64` for Flatcar.
It is based on the previous commit [ea5698d5879f](https://github.com/kinvolk/coreos-overlay/commit/ea5698d5879f).

This PR is basically the same as https://github.com/kinvolk/coreos-overlay/pull/740 , but it is only to update LTS.
Doing that, we can finally remove workaround for maint branch in mantle.

## Testing done

CI passed, http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3344/cldsv except an unrelated `cl.verity`.


krnowak (approved):
Looks good. 2.8 it's a bit old, but I suppose it's fine for LTS.

dongsupark:
Thanks for the review.
Next week, I will add a changelog entry and merge.

krnowak:
> Thanks for the review. Next week, I will add a changelog entry and merge.

I don't think that the lts branch has a changelog directory. Maybe it's something we forgot to do.

dongsupark:
Added changelog

PR #1240, opened 2021-08-31
coreos-base/coreos/coreos: add user binaries to explicit list

pothos:
The image contents are defined by the list in this package and the
dependencies pulled in. Once we would lose some dependency due to
a package change, that would also meant that this dependency's
binaries are not available to the user anymore. To prevent user
binaries from being lost we have to explicitly list them in this
package.
Add the packages that have binaries relevant to the user and are
currently installed (seen in flatcar_production_image_packages.txt
and checked manually). Also add sys-apps/acl which got lost when
    removing rkt.


jepio (approved):
Could you also bump the revision of the symlink? It helps with portage dependency resolution.

pothos:
Updated

sayanchowdhury (approved):

PR #1241, opened 2021-08-31
sys-kernel/coreos-firmware: fix symlink to Intel ICE firmware

dongsupark:
Now that Kernel config `CONFIG_ICE` is [enabled](https://github.com/kinvolk/coreos-overlay/pull/1237), its corresponding firmware file needs to be also in place.
However, upstream linux-firmware tarball does not contain a correct symlink to `intel/ice/ddp/ice-1.3.26.0.pkg`, but `modinfo ice.ko` shows it requires `ice.pkg`.
So we need to create the symlink to avoid failures at the firmware scanning stage like below:

```
Missing firmware: intel/ice/ddp/ice.pkg (ice.ko.xz)
```

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3461/cldsv


krnowak (approved):
Urgh, my python3-oem build got bitten by this problem. Thanks for fixing it.

The changes make sense.

sayanchowdhury (approved):
We would need this patch to release v2983.0.0

dongsupark:
cherry-picked to `flatcar-2983`.

PR #1244, opened 2021-09-01
mdadm: migrate cron.weekly to systemd.timer

aniruddha2000:
# Replace cron.weekly with systemd.timer

Add a mdadm.timer unit and associate it with the mdadm package and remove the cron.weekly.

## Testing done

### DIFF
--- a/sys-fs/mdadm/files/mdadm.service
+++ b/sys-fs/mdadm/files/mdadm.service
@@ -2,4 +2,4 @@
 Description=Initiates a check run of an MD array's redundancy information.
 
 [Service]
-ExecStart=/usr/sbin/checkarray
+ExecStart=/usr/sbin/checkarray --cron --all --idle --quiet
--- a/sys-fs/mdadm/files/mdadm.service
+++ b/sys-fs/mdadm/files/mdadm.service
@@ -2,4 +2,5 @@
 Description=Initiates a check run of an MD array's redundancy information.
 
 [Service]
+Type=oneshot
 ExecStart=/usr/sbin/checkarray --cron --all --idle --quiet
--- a/sys-fs/mdadm/files/mdadm.service
+++ b/sys-fs/mdadm/files/mdadm.service
@@ -1,0 +1,5 @@
+[Unit]
+Description=Initiates a check run of an MD array's redundancy information.
+
+[Service]
+ExecStart=/usr/sbin/checkarray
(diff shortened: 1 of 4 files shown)


aniruddha2000:
@pothos can you give me a little code review that what I am missing. Because when I have built an image the `systemctl status *timer` is not giving me the mdadm.timer

sys-fs/mdadm/files/mdadm.service
```diff
@@ -0,0 +1,5 @@
+[Unit]
+Description=Initiates a check run of an MD array's redundancy information.
+
+[Service]
+ExecStart=/usr/sbin/checkarray
```
pothos, line 5:
```suggestion
ExecStart=/usr/sbin/checkarray --cron --all --idle --quiet
```
I think we should keep the same flags

sys-fs/mdadm/mdadm-4.1.ebuild
```diff
@@ -78,8 +78,7 @@ src_install() {
 	insinto /etc/default
 	newins "${FILESDIR}"/etc-default-mdadm mdadm
 
-	exeinto /etc/cron.weekly
-	newexe "${FILESDIR}"/mdadm.weekly mdadm
+	systemd_dounit "${FILESDIR}"/mdadm.service
```
pothos, line 81:
```suggestion
	systemd_dounit "${FILESDIR}"/mdadm.service
	systemd_dounit "${FILESDIR}"/mdadm.timer
```
Both need to be added, I'll need to find how to enable the timer.

pothos: I think `systemd_enable_service timers.target mdadm.timer`

sys-fs/mdadm/files/mdadm.service
```diff
@@ -0,0 +1,5 @@
+[Unit]
+Description=Initiates a check run of an MD array's redundancy information.
+
+[Service]
+ExecStart=/usr/sbin/checkarray --cron --all --idle --quiet
```
pothos, line 5:
```suggestion
Type=oneshot
ExecStart=/usr/sbin/checkarray --cron --all --idle --quiet
```
Making it a oneshot service unit also makes sense (cf. , e.g.,`fstrim.service` as shipped in Fedora/Debian)

pothos:
I started a test build for this (build id is 3483)

pothos (approved):

pothos:
Thanks, the timer works

PR #1245, opened 2021-09-01
profiles: Enable selinux for all targets

pothos:
Move the USE options out of the amd64 path, specify selinux
packages as explicit dependency, and add accept keywords.

Replaces https://github.com/kinvolk/coreos-overlay/pull/135

## How to use

The goal is to enable the selinux tests again in kola.

## Testing done

Ongoing, have to check if selinux use flags get enabled and whether removing the kola exemption for selinux on arm64 works
http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3470/cldsv/

### DIFF
--- a/coreos-base/coreos/coreos-0.0.1.ebuild
+++ b/coreos-base/coreos/coreos-0.0.1.ebuild
@@ -65,13 +65,11 @@ IUSE="selinux"
 
 RDEPEND=">=sys-apps/baselayout-3.0.0"
 
-# Optionally enable SELinux and pull in policy for containers
+# Optionally enable SELinux for dbus and systemd (but always install packages and pull in the SELinux policy for containers)
 RDEPEND="${RDEPEND}
 	sys-apps/dbus[selinux?]
 	sys-apps/systemd[selinux?]
-	selinux? (
-		sec-policy/selinux-virt
-	)"
+	"
 
 # Only applicable or available on amd64
 RDEPEND="${RDEPEND}
@@ -141,9 +139,14 @@ RDEPEND="${RDEPEND}
 	net-misc/wget
 	net-misc/whois
 	net-vpn/wireguard-tools
+	sec-policy/selinux-virt
+	sec-policy/selinux-base
+	sec-policy/selinux-base-policy
+	sec-policy/selinux-unconfined
 	sys-apps/acl
 	sys-apps/attr
 	sys-apps/coreutils
+	sys-apps/checkpolicy
 	sys-apps/dbus
 	sys-apps/diffutils
 	sys-apps/ethtool
@@ -163,6 +166,7 @@ RDEPEND="${RDEPEND}
 	sys-apps/rng-tools
 	sys-apps/sed
 	sys-apps/seismograph
+	sys-apps/semodule-utils
 	sys-apps/shadow
 	sys-apps/usbutils
 	sys-apps/util-linux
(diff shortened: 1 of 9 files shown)


pothos:
The selinux use flags are still not enabled dbus and systemd on arm64, need to find out why

tormath1:
a PR has been created for mantle: https://github.com/kinvolk/mantle/pull/222

pothos:
The use flags don't get picked up which causes this failure as SELinux is always disabled and not in permissive mode:
```
2021-09-02T09:18:29Z kola: creating new machine for semver check: machine "12fe8823-4ca4-42b9-9aa9-6605b0c8aebb" failed to enable selinux: Unable to enable SELinux: Process exited with status 1: setenforce: SELinux is disabled
```

coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -141,9 +139,13 @@ RDEPEND="${RDEPEND}
 	net-misc/wget
 	net-misc/whois
 	net-vpn/wireguard-tools
+	sec-policy/selinux-virt
+	sec-policy/selinux-base
+	sec-policy/selinux-base-policy
```
tormath1, line 144: note: we can also add `sec-policy/selinux-unconfined`

pothos: Done, I'm still having the problem that the selinux use flag doesn't get applied on arm64 (dbus, systemd, and I think even tar and other packages)

krnowak: @pothos : Have a look at `profiles/coreos/amd64/generic/use.mask` - `selinux` is masked there.

pothos:
I am a bit confused by the solution but unmasking for all boards in the same way helped:
```
Calculating dependencies... done!
[ebuild   R    ] sys-apps/coreutils-8.32-r1::coreos to /build/arm64-usr/ USE="selinux* xattr -acl -caps -gmp -hostname -kill -multicall -nls -split-usr -static -test -vanilla" 5424 KiB
[ebuild   R    ] app-arch/tar-1.34::portage-stable to /build/arm64-usr/ USE="minimal selinux* xattr -acl -nls" 0 KiB
[ebuild   R    ] sys-process/lsof-4.89::portage-stable to /build/arm64-usr/ USE="ipv6 rpc selinux* -examples -static" 767 KiB
[ebuild   R    ] sys-apps/systemd-247.9:0/2::coreos to /build/arm64-usr/ USE="build cryptsetup curl gcrypt http hwdb idn importd kmod lz4 lzma nat pam pcre policykit resolvconf seccomp selinux* (sysv-utils) (vanilla) zstd -acl (-apparmor) (-audit) -dns-over-tls -elfutils -gnuefi -homed -pkcs11 -pwquality -qrcode -repart -split-usr -ssl -test -xkb" 0 KiB
[ebuild   R    ] sys-apps/dbus-1.12.20-r1::coreos to /build/arm64-usr/ USE="selinux* (systemd) -X -debug -doc (-elogind) -static-libs -test -user-session" 0 KiB
[ebuild   R    ] coreos-base/coreos-0.0.1-r300::coreos to /build/arm64-usr/ USE="selinux*" 0 KiB
```

pothos:
Now the new failure is this here when running kola tests after kola turns SELinux to enforcing mode:
```
semver check: machine "de5db1e9-6dfa-4a65-bf1e-46436371bde4" failed basic checks: some systemd units failed:
��� sys-kernel-tracing.mount loaded failed failed Kernel Trace File System
```
But in general we have SELinux installed and somehow present, finally after this started in Dec. 2019.

krnowak (approved):

pothos:
Actually it was a temporary unit failure due to qemu being a bit slow on the test machine

PR #1246, opened 2021-09-01
app-misc/ca-certificates: consider system encoding when opening file

dongsupark:
Now that we started [encoding](https://github.com/kinvolk/coreos-overlay/pull/1229) strings to unicode by default, we should also take care of corner cases, where LC_CYPTE is set to a different value from the systemd default value in `/etc/locale.gen`.

For example, under a build environment with `LC_CTYPE=C`, when the UTF-8 file name is `AC_Ra�z_Certic�mara_S.A..pem`, build fails like that.

```
Traceback (most recent call last):
  File "/var/tmp/portage/app-misc/ca-certificates-3.27.1-r2/files/certdata2pem.py",
line 127, in <module>
    f = open(fname, 'w')
UnicodeEncodeError: 'ascii' codec can't encode character '\xed' in position 5: ordinal not in range(128)
 * ERROR: app-misc/ca-certificates-3.27.1-r2::coreos failed (compile phase):
```

To fix that, ignore UnicodeEncodeError when opening the file.

## How to test

```
sudo LC_CTYPE=C emerge app-misc/ca-certificates  # under the SDK
```

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3484/cldsv/


app-misc/ca-certificates/files/certdata2pem.py
```diff
@@ -124,7 +124,15 @@
...
-        f = open(fname, 'w')
+
+        # In some cases, it is not possible to encode a file name with utf8,
+        # especially when LC_CTYPE is set to a different value from the
+        # system default value in /etc/locale.gen. In that case, we need to
+        # ignore UnicodeEncodeError, to avoid running failures.
+        try:
+            f = open(fname, 'w')
+        except UnicodeEncodeError:
+            pass
```
pothos, line 135: somehow `fname` needs to be cleaned up, so that `f` gets a new file opened

pothos: some SO comment says this: `.encode(sys.getfilesystemencoding())`

krnowak: To be sure, we could do
```python
f = open(fname.encode(encoding=sys.getfilesystemencoding(), errors="ignore"))
```
That for filename like `AC_Ra�z_Certic�mara_S.A..pem` could result creating a file like `AC_Raz_Certicmara_S.A..pem`.

I'm only wondering where did you get the idea of running the script under `LC_CTYPE=C` env var? :)

pothos: Thanks!
And why is `LC_CTYPE=C` even causing utf8 usage to fail?

dongsupark: > f = open(fname.encode(encoding=sys.getfilesystemencoding(), errors="ignore"))

Thanks, updated the PR like that. CI running.

> I'm only wondering where did you get the idea of running the script under LC_CTYPE=C env var? :)

Just some dirty detective work. ;-)

With the current main branch, Jenkins CI works well, but not with my local SDK.
I found out in my local SDK, a normal user cannot run the python script, while root can do.
So I stored all env variables of both users, i.e. `env | sort`, and compared them to each other.
The diff was huge, but reading that, I suspected it had something to do with locale.

The normal user had all 14 env variables configured in my `~/.profile` etc., mostly `en_US.UTF-8` or `C`.
But root did not set custom variables, so simply followed the default one `C.UTF-8` as in `/etc/locale.gen`.

So I tested installing ca-certificates with one variable change after another, and figured out only `LC_CTYPE` makes the difference.
To make the normal user capable of running the script, I had to explicitly disable `LC_CTYPE`, like `LC_CTYPE= emerge ca-certificates`.
LC_ALL or LANG did not play any role here.

> And why is LC_CTYPE=C even causing utf8 usage to fail?

Some web searching returns like:

> LC_CTYPE defines character classes, which are named sets of characters used by a variety of programs, especially regular expression matchers. In programs such as grep that use character classes like [:alnum:], class membership varies with locale. If the writing system distinguishes upper- and lower-case letters, LC_CTYPE specifies the relationship between the two.

https://www.linux.com/news/controlling-your-locale-environment-variables/

Not sure but, I suppose its reason is locale generation in the SDK got somehow broken when LC_CTYPE is not the expected one.

dongsupark:
CI passed

pothos (approved):

PR #1261, opened 2021-09-06
Update net-misc/google-cloud-sdk to 355.0.0

krnowak:
This is to get rid of the python2 dependency.

~~I also install additional tool - bq (it's for big query stuff).~~

CI: ~~http://localhost:9091/job/os/job/manifest/3517~~ http://localhost:9091/job/os/job/manifest/3532/cldsv/


krnowak:
CI passed.

krnowak:
I dropped the dependency on pyopenssl and rerun the CI. Seems to work. pyopenssl was used only for signurl command which we only use with gsutil coming from docker image. Also added some README with some notes about the package.

CI: http://localhost:9091/job/os/job/manifest/3570/cldsv/

jepio (approved):
Love it. Getting rid of pyopenssl is also a preqrequisite of the openssl upgrade.

dongsupark:
If we can delete pyopenssl, then we can also delete `dev-python/cryptography`. :+1:

krnowak:
Yeah, I'm actually doing it in https://github.com/flatcar-linux/portage-stable/pull/214, but it's still a draft.

PR #1263, opened 2021-09-07
sys-auth/polkit: bump to 0.119 and apply duktape patchset

wrl:
This PR does two things simultaneously:
 - updates polkit to 0.119 (in part to resolve CVE-2021-3560, https://gitlab.freedesktop.org/polkit/polkit/-/issues/140)
 - applies a [patchset to use duktape](https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/35) as the Javascript interpreter engine instead of Spidermonkey

Duktape is substantially smaller than Spidermonkey, which both reduces package build time and allows us to ship a smaller image (though I have not determined how *much* smaller). More importantly, Spidermonkey and its transitive dependencies have been a substantial problem for building arm64 images, and using duktape sidesteps this issue nicely. As such, this PR *also* resolves the outstanding arm64 polkit issue (ref #863 https://github.com/kinvolk/Flatcar/issues/474 https://github.com/kinvolk/Flatcar/issues/156).

Of note is that this patchset merge request is marked as WIP and the polkit maintainer has issues with it that are still outstanding. I am going to attempt to resolve them and push this work further along. For the time being, if we ship this duktape'd polkit, we need to be clear publicly that it potentially introduces issues with custom rules.

An option would be to continue depending on Spidermonkey for amd64 and only use duktape for arm64, but a counterpoint is that shipping duktape for both architectures allows for more testing and feedback.

CI :large_blue_circle: : http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3866/cldsv/

### DIFF
--- a/profiles/coreos/arm64/package.use
+++ b/profiles/coreos/arm64/package.use
@@ -5,5 +5,3 @@ net-dns/bind-tools -gssapi
 
 # FIXME: why isn't this set by default???
 sys-libs/ncurses unicode
-
-sys-auth/polkit -introspection
--- a/profiles/coreos/arm64/package.use.force
+++ b/profiles/coreos/arm64/package.use.force
@@ -1,4 +1,3 @@
-sys-auth/polkit -introspection
 sys-apps/systemd -introspection
 
 # Matt Turner <mattst88@gentoo.org> (2020-03-28)
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -75,9 +75,6 @@ dev-libs/cyrus-sasl kerberos -berkdb -gdbm
 # don't build manpages for sssd
 sys-auth/sssd -python samba kerberos gssapi ssh sudo
 
-# needed for realmd build
-sys-auth/polkit introspection
-
 # enable logging command-line options in update_engine
 dev-cpp/glog gflags
 
@@ -147,3 +144,6 @@ dev-lang/perl minimal
 
 # Disable cgroup-hybrid as we use the unified mode
 sys-apps/systemd -cgroup-hybrid
+
+# Remove support for GObject introspection
+sys-auth/polkit -introspection


pothos:
Can you follow https://github.com/kinvolk/coreos-overlay/pull/863#pullrequestreview-598679198 to make sure we can identify the downstream changes and also know that previous downstream changes are preserved?
When that's done we can start a CI build.

wrl:
> Now find all downstream patches for the package by running git log CATEGORY/PACKAGE. If everybody followed the process of resetting before importing an upstream update, you only have to look for the commits after the last update and port them to the new version. Otherwise you have to compare the files manually to their upstream versions from older portage revisions.

@pothos What constitutes a "downstream patch", exactly? To be honest with you, I have no idea what makes https://github.com/kinvolk/coreos-overlay/commit/5c4d184e22fd93ab926878a131150047b54d0b6c a commit that needs to be forward-ported and I would appreciate any insight you can provide on this.

jepio:
"Downstream patches" are changes that were applied to the ebuild in this repository compared to the version in https://github.com/gentoo/gentoo (the upstream). Many of the "downstream" changes are things like in 5c4d184: installing Flatcar specific configuration files or files in Flatcar specific locations (keeping in mind that: /usr is read-only and gets auto-updated).

pothos:
The commits in `git log sys-auth/polkit/` which result in the difference between `polkit-0.113-r5.ebuild` and the last upstream ebuild file this was based on: https://github.com/gentoo/gentoo/blob/e9aa53351ccea3a7e9ee29f8597e9e51ec5449f4/sys-auth/polkit/polkit-0.113-r4.ebuild

The example commit I took is quite interesting because it relates to how Flatcar is updated:
```
 polkit: fix config install paths, use systemd-tmpfiles

All configs should be installed to /usr and tmpfiles should be used to
create and fix directory permissions instead of the ebuild's postinst.
```
What is auto-updated is the `/usr` partition, which means that `/etc` and `/var` is managed by the user. Everything that can be moved from `/etc` to `/usr/share` is moved by this commit, but some things have to be set up on `/etc` and `/var`. A way to set up the required files or  folders on `/etc` and `/var` from the auto-updated `/usr` partition is by shipping systemd-tempfile directives (`systemd_dotmpfilesd` uses the downstream `sys-auth/polkit/files/polkit.conf` file for this purpose).

dongsupark:
Exciting to see the duktape-based polkit! :+1: 

Ideally, we should move the native Gentoo ebuilds, i.e. without Flatcar patches, to portage-stable.
`acct-group/polkitd`, `acct-user/polkitd`, `dev-lang/duktape` belong to there.
Then we can keep only `sys-auth/polkit` in coreos-overlay.

wrl:
@jepio @pothos Okay, is there any way to, say, just get a list of outstanding downstream patches that I have to port forward, or does this just entail *manually* inspecting commit-by-commit?

pothos:
Getting the list easily requires that people follow this best practice but it wasn't done in the case of this package here. But if you update by _resetting_ the package to upstream state and then adding the downstream changes in a separate commit (one big or many small as you like), it will be easy for the next person to get the list of downstream patches.

pothos:
Or actually this package here followed this practice somehow already and you can consider all commits from fb5590cb1b45718a3ead80c193eea7318b0db825 onwards as downstream changes but the patches that are included in the new polkit release can be dropped.

krnowak:
With my python2-eradicator hat on, I'd love this to get this in, as spidermonkey is still using python2.

pothos:
The second commit should say _re-apply downstream patches_ and also include the duktape patch as it is a downstream patch compared to Gentoo.

tormath1:
CI is OK with arm64 but not with amd64, polkit pulls python through an intermediate dependency. :thinking:

krnowak:
> Looks like we'll have to disable `introspection` useflag

Not sure if that's an option, grepping for sys-auth/polkit gives me among other lines:

```
sys-auth/realmd/realmd-0.17.0.ebuild:DEPEND="sys-auth/polkit[introspection]
```

I'll need to check if realmd really needs introspection stuff from polkit.

krnowak:
I had a look at realmd's and polkit build systemd and I think that realmd's dependency on polkit with introspection is bogus - it should be fine with being just `DEPEND="sys-auth/polkit`, without the `[introspection]` part. And there's nothing in portage-stable that depends on polkit, so that ought to work.

krnowak:
And dropping introspection from polkit would be nice - it was giving me a headache when I tried to update gobject-introspection, because polkit then was trying to put or get some introspection things from `/build/amd64-usr/build/amd64-usr/blah/blah/blah` and I couldn't figure out why.

wrl:
@pothos Considering that the duktape patch is pulled from Gitlab and is therefore version-dependent, does it make sense to have it be part of the roll-up patch? It seems like that's just adding complexity next time we have to bump the version (i.e. "reformat these patches for the new version, but also re-fetch this other patch from Gitlab".)

pothos:
Yes, a separate commit would be nice, that documents how to create it.

wrl:
Have updated this PR with separate duktape patchset commit.

Note: this **will not build** with only the first commit ("bump to 0.119") applied because it depends on a version of Spidermonkey which we do not have in portage-stable.

PR #1269, opened 2021-09-09
Initial steps for python3 transition

krnowak:
The general plan for transitioning from python2 to python3 is more or less as follows:
- Update the packages, so they stop depending on python2 (which also means that they don't put `python2_7` in `PYTHON_COMPAT` array in ebuilds).
- Drop python2.
- Update python3 and python eclasses.

The problem with the first step is that usually packages that stopped supporting python2 support python3 version that we haven't packaged. Build such packages usually resulted in errors like:

```
 * ERROR: dev-lang/python-exec-2.4.8::portage-stable failed (depend phase):
 *   Invalid implementation in PYTHON_COMPAT: python3_8
```

So my usual workaround was to move the package from portage-stable to coreos-overlay, and modify the ebuild to drop `python3_8` from `PYTHON_COMPAT`. But for transition process could result in a lot of packages moved over, only to be moved back when we get an updated python3. But apparently it is enough to modify the python-utils-r1.eclass to add the missing python versions (3.8, 3.9 and 3.10), and mask them. So here I'm doing just that, and moving some packages that I moved to overlay from portage-stable some time ago, back to portage-stable.

Needs to be merged together with https://github.com/kinvolk/portage-stable/pull/212.

CI: http://localhost:9091/job/os/job/manifest/3546/


krnowak:
CI passed (had to rerun os/kola/qemu test).

profiles/coreos/base/use.mask
```diff
@@ -4,6 +4,12 @@ kdbus
 # We default to python 3.6 for now
 python_targets_python3_7
 python_single_target_python3_7
+python_targets_python3_8
```
tormath1, line 7: Maybe we can reword the commit to explain why we mask this python versions:

From the PR body:
> [...] it is enough to modify the python-utils-r1.eclass to add the missing python versions (3.8, 3.9 and 3.10), and mask them.

krnowak: Right, I have added some more words to commit messages in both PRs.

dongsupark (approved):
Thanks a lot.
With the changes, the python compat errors are gone. :tada: 

Though you need a rebase.

PR #1278, opened 2021-09-15
Add alias for bcc tools to run in docker container

aniruddha2000:
# Add alias for bcc tools to run in docker container
Added alias for bcc tool packages `tcpretrans tcpconnect tcpaccept biolatency ext4slower ext4dist` to run necessary debug tools

## Testing done
- `emerge-amd64-usr app-shells/bash`
- Run those alias in the image.

### DIFF
--- a/app-shells/bash/files/dot-bashrc
+++ b/app-shells/bash/files/dot-bashrc
@@ -16,12 +16,12 @@ fi
 
 
 # Put your fun stuff here.
-sysstat_debug_toolset=( sar iostat vmstat mpstat pidstat )
-bcc_debug_toolset = ( tcpretrans tcpconnect tcpaccept biolatency ext4slower ext4dist )
 
-for t in "${sysstat_debug_toolset[@]}"; do alias_tool "${t}"; done
-for t in "${bcc_debug_toolset[@]}"; do alias_tool "${t}"; done
-
-alias_tool() {
+alias_bcc_tool() {
     local tool="${1}"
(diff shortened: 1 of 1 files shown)


aniruddha2000:
Still getting error for the alias `iovisor-ext4dist`
```bash
cannot attach kprobe, probe entry may not exist
Traceback (most recent call last):
  File "/usr/share/bcc/tools/ext4dist", line 205, in <module>
    b.attach_kprobe(event="ext4_file_write_iter", fn_name="trace_entry")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 834, in attach_kprobe
    (fn_name, event))
Exception: Failed to attach BPF program trace_entry to kprobe ext4_file_write_iter
```

for alias `iovisor-ext4slower`
```bash
cannot attach kprobe, probe entry may not exist
Traceback (most recent call last):
  File "/usr/share/bcc/tools/ext4slower", line 318, in <module>
    b.attach_kprobe(event="ext4_file_write_iter", fn_name="trace_write_entry")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 834, in attach_kprobe
    (fn_name, event))
Exception: Failed to attach BPF program trace_write_entry to kprobe ext4_file_write_iter
```

aniruddha2000:
@pothos would you like to build an image and test it :)

pothos:
Thanks!
I started a build to verify that the alias gets set as expected (will paste the image url for build 3597 here). For testing it's enough to run the lines on a shell in the current Stable or Alpha image. I would exclude non-working commands until we get them working.

aniruddha2000:
@pothos then after your build if those commands are also not working then would you like me to remove those aliases?

pothos:
In general setting up the alias works well, but not for the root user which has no `.bashrc` file. I think it's ok to add one but for now using only `core` should also be good.

Yes, please remove the two aliases that don't work.

aniruddha2000:
In general do we have any specific implementation of the root `.bashrc` or if we are going for that then we have to implement our own from scratch?

pothos (approved):

pothos:
Thanks, there is nothing set up for the root bashrc but I see no problem in adding it

PR #1295, opened 2021-09-22
sys-apps/net-tools: Move back to portage-stable

krnowak:
We are going to update it to a newer version, which fixes the build
system issue that was a reason for putting the package in overlay.

This is to get rid of EAPI 5 in the package.

To be merged together with https://github.com/flatcar-linux/portage-stable/pull/217.

CI: http://localhost:9091/job/os/job/manifest/3646/cldsv/


krnowak:
CI passed.

sayanchowdhury (approved):
LGTM

PR #1297, opened 2021-09-22
sys-auth/realmd: Update to 0.17.0

krnowak:
This seems to be still unpackaged by gentoo, so refresh it on our own
then and do some maintainance work on it:

- Update to the recent EAPI, replacing autotools-utils with autotools,
  and systemd_get_unitdir with systemd_get_systemunitdir.

- Add a patch from upstream fixing the pkg-config detection, so our
  hack during configure phase is not necessary any more.

- Patch the configure script to put the D-Bus policy files in
  /usr/share instead of /etc. This removes a need for a hack in the
  install step.

This is to get rid of EAPI 5 in the package.

CI: http://localhost:9091/job/os/job/manifest/3643/cldsv/


krnowak:
CI passed.

sayanchowdhury (approved):
LGTM

PR #1298, opened 2021-09-22
sys-fs/lvm2: Sync with gentoo

krnowak:
This is to get rid of EAPI 5 from the package.

This also drops some of our modifications - those were about dropping circular dependencies between systemd, udev and baselayout. Those can be worked around by the `break_dep_loop` in `build_packages` script.

CI: http://localhost:9091/job/os/job/manifest/3648/cldsv/


krnowak:
CI passed.

sys-fs/lvm2/Manifest
```diff
@@ -1,2 +1 @@
 DIST LVM2.2.02.188.tgz 2421550 BLAKE2B bed90c8454cd4b20fdeec6dcbf5a9f97c9310671aea3b2252f8069cfa439fcb050f5ad95f928a7125a1734a4dc5ac985da99a4a570538e377a7205191a505476 SHA512 8c9db17c49dc8ebcab6c7f246ab85870a80658be811cf7f4d8f36abbebafa355b030bfc1e3bcbad73ccccb7fcd06d4a95ac547ca15d18d33715126da92703dca
-DIST LVM2.2.03.13.tgz 2619537 BLAKE2B 8f93896105e4d392ab88f027bbc0ea874b179a73d10ff8d0850ead2625e86468fe69d55c6fb0108f57a1efc1d3e8cc94647f1bc06636e4a714d0817224bbbfd7 SHA512 8b4fb6da5dd46ddeb754436189072bea79be594fab6d57d9be63c6ce3e3cb68d38601a2c031c493681321281cfb1f1459ace19974c00beb569364f44863cfc4a
```
sayanchowdhury: We drop the not required stuff when doing the "Sync with Gentoo"

sayanchowdhury (approved):
one comment, otherwise looks good.

sys-fs/lvm2/Manifest
krnowak: Alright, squashed the commits.

PR #1303, opened 2021-09-24
*: Update packages with EAPI <= 5 to EAPI 7

krnowak:
EAPI 5 is deprecated in gentoo and EAPI 1 to 4 are banned. Using old EAPIs may block updating eclasses in portage-stable, because many of the new versions of eclasses drop support for old EAPIs.

EAPI 6 seems to be also deprecated, but updating them is not as pressing as for lower versions.

Needs to be merged together with https://github.com/flatcar-linux/portage-stable/pull/218.

CI: http://localhost:9091/job/os/job/manifest/3661/cldsv/


krnowak:
CI passed.

krnowak:
Just to explain more the changes I've made:

- The versionator eclass is deprecated in EAPI 7 - it's replaced by some `ver_` functions.
- In EAPI 7, if src_prepare is overridden, it must call either `default` or `eapply_user`. If neither is called, build fails. `default` tries to apply patches from the `PATCHES` variable (if defined) and calls `eapply_user`. I chose to call `default`.
- In one package I had to define the `S` variable - the package is a meta-package, but has some custom `src_install`. The overridden function probably triggered `emerge` to try to enter the directory like `${WORKDIR}/${P}`, which wasn't created. Normally such directories are created as a part of unpacking a tarball or something, but in this case the package didn't have any tarball.
- The `autotools-utils` eclass is replaced by `autotools` eclass. The former is even gone by now from gentoo. It was more or less replaced by standard gentoo helper functions like emake or econf or by the builtin EAPI functionality. The other feature, running autoreconf, was put into the `autotools` eclass, so now instead of defining the `AUTOTOOLS_AUTORECONF` variable, we should just call `eautoreconf` in `src_prepare`.
- The `ROOT` to `SYSROOT` changes are described in a commit message. But in short - `ROOT` should not be used in `src_` functions, because it's basically an empty string. This is more or less a case by case thing, but in this PR it was mostly replaced by using `SYSROOT`. I had some more changes in some selinux packages, buuut I was too scared to add them here - they will end up in a separate PR. :)
- Replaced `git-2` with `git-r3`, because `git-2` is old and deprecated. But it happened only in `sys-devel/sysroot-wrappers-9999.ebuild` which is actually never used…
- Dropped `cros-debug` from one ebuild - it wasn't used.

dongsupark (approved):
Looks good.
Though you need a rebase.

PR #1310, opened 2021-09-27
.github: get correct ebuild file name from old version

dongsupark:
The bug fix https://github.com/flatcar-linux/coreos-overlay/pull/1129 caused a regression that Github Actions cannot determine a correct `$VERSION_OLD` if the old ebuild file has a suffix like `-r1`.
That's why recently Github Actions for firmware have failed to run.
We need to create a function to get a correct ebuild file name, by falling back to the most similar name, in case the expected ebuild file does not exist.

### DIFF
--- a/.github/workflows/rust-apply-patch.sh
+++ b/.github/workflows/rust-apply-patch.sh
@@ -23,9 +23,8 @@ fi
 # replace rust version in profiles/, e.g. package.accept_keywords.
 find profiles -name 'package.*' | xargs sed -i "s/=dev-lang\/rust-\S\+/=dev-lang\/rust-${VERSION_NEW}/"
 
-pushd "dev-lang/rust" >/dev/null || exit
-git mv $(ls -1 rust-${VERSION_OLD}.ebuild) "rust-${VERSION_NEW}.ebuild"
-popd >/dev/null || exit
+EBUILD_FILENAME=$(get_ebuild_filename "dev-lang" "rust" "${VERSION_OLD}")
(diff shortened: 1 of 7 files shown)


.github/workflows/rust-apply-patch.sh
```diff
@@ -23,9 +23,8 @@ fi
 # replace rust version in profiles/, e.g. package.accept_keywords.
 find profiles -name 'package.*' | xargs sed -i "s/=dev-lang\/rust-\S\+/=dev-lang\/rust-${VERSION_NEW}/"
 
-pushd "dev-lang/rust" >/dev/null || exit
-git mv $(ls -1 rust-${VERSION_OLD}.ebuild) "rust-${VERSION_NEW}.ebuild"
-popd >/dev/null || exit
+EBUILD_FILENAME=$(get_ebuild_filename "dev-lang" "rust" "${VERSION_OLD}")
+git mv "${EBUILD_FILENAME}" "rust-${VERSION_NEW}.ebuild"
```
krnowak, line 27: You do not enter into `dev-lang/rust` directory any more so target of `git mv` needs to be updated too:

```suggestion
git mv "${EBUILD_FILENAME}" "dev-lang/rust/rust-${VERSION_NEW}.ebuild"
```

krnowak (changes requested):
Looks good, but needs a fix for rust.

.github/workflows/rust-apply-patch.sh
dongsupark: Good catch. Fixed, thanks.

krnowak (approved):

PR #1312, opened 2021-09-28
app-crypt/gnupg: update to 2.2.29

dongsupark:
Update `app-crypt/gnupg` to 2.2.29.
One of the key purposes for the update is to be able to use the new default keyserver `keyserver.ubuntu.com`, which is provided by default since 2.2.29. It is due to the shutdown of the SKS keyserver pools.
See also https://bugs.gentoo.org/811828

Add Flatcar patches for accepting without UIDs, based on 3d9a9c9c3654c6b8c073e306636bf8dc64cfb657 .
Use EAPI=7, until EAPI 8 could be fully supported.
Delete unnecessary `app-crypt/gnupg` from accept_keywords.

Fixes https://github.com/flatcar-linux/Flatcar/issues/179.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/qemu/2712/cldsv


krnowak (approved):
Looks good.

PR #1319, opened 2021-09-30
.github: new Github workflow for open-vm-tools

dongsupark:
Automatically update `coreos/open-vm-tools` as well as `coreos-base/oem-vmware`.
Get the latest open-vm-tools release number, and get its build number from the Github repo, and replace the old build number with the new one.
Also sync `coreos-base/oem-vmware` in line with `open-vm-tools`.

To do that, we need to also split git repo init part into `prepare_git_repo`, to prevent some corner cases, where applying multiple patches does not work because the latter overwrites the former patch.
We need to also clean up an unnecessary patch file.

### DIFF
--- a/.github/workflows/vmware-releases-main.yml
+++ b/.github/workflows/vmware-releases-main.yml
@@ -2,6 +2,7 @@ name: Get the latest open-vm-tools release for main
 on:
   schedule:
     - cron:  '0 7 * * 3'
+  workflow_dispatch:
 
 jobs:
   get-vmware-release:
--- a/.github/workflows/vmware-releases-main.yml
+++ b/.github/workflows/vmware-releases-main.yml
@@ -1,0 +1,47 @@
+name: Get the latest open-vm-tools release for main
+on:
+  schedule:
+    - cron:  '0 7 * * 3'
+
+jobs:
+  get-vmware-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Fetch latest open-vm-tools release
+        id: fetch-latest-release
+        run: |
+          git clone https://github.com/vmware/open-vm-tools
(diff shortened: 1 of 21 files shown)


.github/workflows/vmware-releases-main.yml
```diff
@@ -0,0 +1,47 @@
...
+          PATH: ${{ steps.setup-flatcar-sdk.outputs.path }}
+          VERSION_NEW: ${{ steps.fetch-latest-release.outputs.VERSION_MAIN }}
+        run: .github/workflows/vmware-apply-patch.sh
+      - name: Create pull request for main
+        uses: peter-evans/create-pull-request@v3
+        if: steps.apply-patch-main.outputs.UPDATE_NEEDED == 1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          base: ${{ steps.fetch-latest-release.outputs.BASE_BRANCH_MAIN }}
+          branch: vmware-${{ steps.fetch-latest-release.outputs.VERSION_MAIN }}-main
```
jepio, line 41: This `main` should be `${{ steps.fetch-latest-release.outputs.BASE_BRANCH_MAIN }}`

.github/workflows/vmware-releases-main.yml
```diff
@@ -0,0 +1,47 @@
+name: Get the latest open-vm-tools release for main
+on:
+  schedule:
+    - cron:  '0 7 * * 5'
+
```
jepio, line 5: If you add ` workflow_dispatch:` as a trigger, then we can manually test it right after merging.

jepio (approved):
Looks good, had just two minor comment, but i think it can be merged

.github/workflows/vmware-releases-main.yml
dongsupark: Right, done

.github/workflows/vmware-releases-main.yml
dongsupark: Good idea, added workflow_dispatch to every workflow.

PR #1320, opened 2021-09-30
Remove util-linux entry from package.accept_keywords

aniruddha2000:
# Remove util-linux entry from package.accept_keywords


krnowak:
I updated the comment to mention the respective PR in portage-stable.

PR #1344, opened 2021-10-12
net-misc/openssh: Sync with Gentoo upstream; updates to openssh 8.8_p1

sayanchowdhury:
# net-misc/openssh: Sync with Gentoo upstream; updates to openssh 8.8_p1

Fixes [CVE-2021-41617](https://cve.circl.lu/cve/CVE-2021-41617)

Pulls in https://github.com/flatcar-linux/init/pull/54.

## How to use

```bash
emerge-amd64-usr net-misc/openssh
```

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4343/cldsv


dongsupark:
Thanks for working on that.
We should be a little more careful when testing openssh 8.8, as ssh-rsa/sha1 is disabled by default.
See also https://www.gentoo.org/support/news-items/2021-10-08-openssh-rsa-sha1.html .

dongsupark:
FYI, a rebase is needed

dongsupark:
Rebased, and updated the version to 8.8_p3.

dongsupark:
CI fails consistently like:
```
not ok - cl.update.payload
  ---
  Error: "harness.go:535: Cluster failed starting machines: machine \"928c35d4-aaae-4be3-ad10-c751709537da\" failed to start: ssh journalctl failed: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain"
```

tormath1:
@dongsupark after looking a bit into the Mantle code, I think the issue is related to: https://github.com/golang/go/issues/49952 as we can SSH into the instance with a regular OpenSSH client (tried locally with `net-misc/openssh-8.6_p1-r2`)

dongsupark:
@tormath1 Thanks for finding the tracking issue.
It looks like we should wait for Go 1.18.
Set the status to Long Term / Blocked.

tormath1:
@dongsupark in the meantime, what do you think about adding the following[^1] to the `sshd_config`:
```
HostkeyAlgorithms +ssh-rsa
PubkeyAcceptedAlgorithms +ssh-rsa
```

I tried locally, it seems to work.

It would allow us to ship `openssh 8.8_p3` and to prevent breakage for user having the same kind of issue we're having with the CI ? In the next release or after, we could start to deprecate `ssh-rsa` algorithm.

[^1]: https://www.gentoo.org/support/news-items/2021-10-08-openssh-rsa-sha1.html

dongsupark:
Yeah, that makes sense.
Will soon add the configs to sshd_config, and start another CI build.

dongsupark:
CI passed with an additional fix https://github.com/flatcar-linux/init/pull/54. :tada:

dongsupark:
As https://github.com/flatcar-linux/init/pull/54 was merged, update CROS_WORKON_COMMIT of coreos-init.
Also added changelog for openssh 8.8.

tormath1:
It looks good to me, just one thing in the Flatcar changes [commit](https://github.com/flatcar-linux/coreos-overlay/pull/1344/commits/5b02ff8c98be1f3f3a5140da705a4206321de8bc), I don't understand the following mention:
> Instead of dropping bindist, enable it with the profiles now so it
doesn't need to be modified on future updates.

dongsupark:
> Instead of dropping bindist, enable it with the profiles now so it
doesn't need to be modified on future updates.

I am not the person who wrote the commit message.
I suppose we can remove the message, because we do not touch bindist in that commit.
Will soon update the commit.

sayanchowdhury:
before leaving on the break; I was tracking this upstream [issue](https://github.com/golang/go/issues/37278) but things were a bit scattered then, good they created a tracking issue now. I was looking into [migrating to mantle](https://github.com/flatcar-linux/mantle/tree/sayan/migrate-to-ed25519) `ed25519` keys, until I realized that Azure does have support for them.

profiles/coreos/base/package.accept_keywords
```diff
@@ -31,7 +31,9 @@ dev-util/checkbashisms
 
 =net-libs/gnutls-3.7.1 ~amd64 ~arm64
 
-=net-misc/openssh-8.7_p1-r1 ~amd64 ~arm64
+=net-libs/libmicrohttpd-0.9.55 ~amd64 ~arm64
```
tormath1, line 34: We should remove this no ? It seems it has been dropped in this PR: https://github.com/flatcar-linux/coreos-overlay/pull/1408/files since we're about to ship `net-libs/libmicrohttpd-0.9.73`.

dongsupark: Good catch, thanks.
Probably something went wrong during the rebase.
Fixed.

tormath1 (approved):
Nit-pick in the commit title:
> profiles: accept ~arm64, ~amd64 for openssh 8.8_p3

We could update "openssh 8.8_p3" to "openssh 8.8_p1" - otherwise it looks good to me, let's ship this :rocket:

dongsupark:
Fixed the title and commits. Thanks!

PR #1356, opened 2021-10-18
profiles: Drop keywords for sys-fs/mtools

krnowak:
We just updated this package to 4.0.35, which is declared stable for
both amd64 and arm64.

To be merged together with https://github.com/flatcar-linux/portage-stable/pull/226.

CI: http://localhost:9091/job/os/job/manifest/3869/cldsv/


krnowak:
CI passed.

PR #1357, opened 2021-10-19
Adapt to new pkg-config implementation

krnowak:
Update dbus-glib - it includes some fix for cross compiling, but we need to add another one. I think this could be upstreamed too.

update-engine also needed a fix for using a proper build tool.

New pkg-config implementation broke my hack in systemd ebuild for getting systemd unit directory, so I rewrote it to stop using systemd.eclass functionality.

I needed to move some packages (app-crypt/p11-kit, sys-fs/btrfs-progs) from portage-stable to overlay to fix build issues. I think that some of those fixes could be upstreamed to gentoo later. I used that opportunity to update those packages at the same time.

This came up when updating python packages (I needed to update libxslt, which pulled in a newer version of virtual/pkgconfig, which obsoleted pkgconfig in favor of pkgconf, which caused some build issues during cross-compilation).

To be merged with https://github.com/flatcar-linux/portage-stable/pull/227.

CI: http://localhost:9091/job/os/job/manifest/4140/cldsv/


krnowak:
CI more or less passed (I'll keep rerunning qemu-uefi).

krnowak:
qemu-uefi tests passed too.

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -247,6 +247,10 @@ src_configure() {
 	multilib-minimal_src_configure
 }
 
+get_rootprefix() {
+	usex split-usr "${EPREFIX:-/}" "${EPREFIX}/usr"
+}
```
dongsupark, line 252: The code itself looks good.
Ideally we should minimize difference with Gentoo.
I believe you would send the changes to upstream in the future. :-)

dev-libs/dbus-glib/dbus-glib-0.112.ebuild
```diff
@@ -42,9 +51,16 @@ multilib_src_configure() {
...
 
-	# Use host tool when building cross.
-	if ! use cros_host; then
-		myconf+=("--with-dbus-binding-tool=/usr/bin/dbus-binding-tool")
+	# Configure a CBUILD directory to make a native build tool.
+	if tc-is-cross-compiler; then
+		mkdir "${BUILD_DIR}-build" || die
+		cd "${BUILD_DIR}-build" || die
+		ECONF_SOURCE="${S}" econf_build
+		myconf+=( --with-dbus-binding-tool="$PWD/dbus/dbus-binding-tool" )
```
dongsupark, line 59: So we do not need to keep the Flatcar [patch](https://github.com/flatcar-linux/coreos-overlay/commit/3a7d063c729e) for dbus-binding-tool, because Gentoo ebuild is already adding the `--with-dbus-binding-tool` option.
Nice.

app-crypt/p11-kit/p11-kit-0.23.22.ebuild
```diff
@@ -0,0 +1,69 @@
...
+	fi
+	default
+}
+
+multilib_src_configure() {
+	# Flatcar: Override the detection of the systemd user unit
+	# directory and bash completion directory with these
+	# environment variables.
+	local -x systemduserunitdir=$(systemd_get_userunitdir)
+	local -x bashcompdir=$(get_bashcompdir)
```
dongsupark, line 51: I am wondering if there would be a simpler approach, e.g.

```
econf \
...
--with-systemduserunitdir="$(systemd_get_userunitdir)" \
--with-bashcompdir="$(get_bashcompdir)"
```

sys-fs/lvm2/lvm2-2.02.188-r2.ebuild
```diff
@@ -123,6 +123,13 @@ src_prepare() {
 		sed -i -e 's/^\(MAN7+=.*\) $(LVMTHINMAN) \(.*\)$/\1 \2/' man/Makefile.in || die
 	fi
 
+	# Flatcar: The configure script tries to detect the systemd
+	# util dir without providing a way for us to override it, so
+	# modify the script.
+	sed -i \
+	    -e "s#^\(systemdutildir=\).*#\1$(systemd_get_utildir)#" \
+	    configure{.ac,} || die
```
dongsupark, line 131: I am not sure I like the hack.
But looking at the lvm2 source tree, I think it is not possible to pass an option like `--with-systemdutildir`. 
So it is ok.

sys-fs/btrfs-progs/btrfs-progs-5.10.1-r1.ebuild
```diff
@@ -0,0 +1,131 @@
...
+		AT_M4DIR=m4 eautoreconf
+		mkdir config || die
+		local automakedir="$(autotools_run_tool --at-output automake --print-libdir)"
+		[[ -e ${automakedir} ]] || die "Could not locate automake directory"
+		ln -s "${automakedir}"/install-sh config/install-sh || die
+		ln -s "${EPREFIX}"/usr/share/gnuconfig/config.guess config/config.guess || die
+		ln -s "${EPREFIX}"/usr/share/gnuconfig/config.sub config/config.sub || die
+	fi
+	# Flatcar: Replace udevdir variable with proper udev directory.
+	sed -i -e 's#^\(udevdir\s\+=\).*#\1 $(get_udevdir)#' Makefile.inc.in
```
dongsupark, line 99: Same. btrfs-progs does not provide an option like `--with-udevdir`.
It is ok.

dongsupark (commented):
Thanks for the excellent work.

See comments below.

sys-apps/systemd/systemd-9999.ebuild
krnowak: I think that systemd ebuild is the most customized forked ebuild that we have in overlay. So I think we would need to do a bit of work on this ebuild to make some part of it upstreamable either to gentoo or maybe even to systemd.

dev-libs/dbus-glib/dbus-glib-0.112.ebuild
krnowak: True. But then another issue popped up and I still needed to patch the ebuild. :) I'll add upstreaming the changes I have made to gentoo to my TODO list. Maybe at some point we will end up moving dbus-glib into portage-stable even.

Even better would be to port update_engine to use D-Bus functionality implemented in GLib instead, so dbus-glib could be removed, but this is a story for another time.

app-crypt/p11-kit/p11-kit-0.23.22.ebuild
krnowak: I rechecked the p11-kit configure script to be sure. There are only options like `--with-bash-completion` and `--with-systemd`, and those simply enable or disable things without taking any values. I can see that `systemduserunitdir` and `bashcompdir` are set up with the [`PKG_CHECK_VAR`](https://autotools.io/pkgconfig/variables.html#idm46482193783040) macro, which adds the following snippet to the `configure --help` output:

```
Some influential environment variables:
  …
  systemduserunitdir
              value of systemduserunitdir for systemd, overriding pkg-config
  bashcompdir value of completionsdir for bash-completion, overriding
              pkg-config
```

So looks like that environment variables are the only way to override the detection. It's less elegant than passing some `--with-foo` options, but I suppose it's still within "that's how we do stuff in autotools idio(ma)tically". :)

sys-fs/lvm2/lvm2-2.02.188-r2.ebuild
krnowak: I don't like the hack too - it's just plain terrible and ugly. But [as you saw](https://sourceware.org/git/?p=lvm2.git;a=blob;f=configure.ac;h=6cdf1a7e652a982cd648e8b194d774ebfeb9aa41;hb=HEAD#l1636), there isn't a way to override it. I could try sending a patch to lvm2 to add such an option, but I remember that Kai's patch that we still carry also wasn't acknowledged, so no high hopes here. I'll put it on my TODO list.

sys-fs/btrfs-progs/btrfs-progs-5.10.1-r1.ebuild
krnowak: Yeah, I could provide a patch to btrfs-progs to have an option like `--with-udevdir`. I'll add it to my TODO list. This change in ebuild can also be upstreamed to gentoo, which also will be on my TODO list.

dongsupark (approved):
Thanks for the explanation.

PR #1358, opened 2021-10-20
ca-certificates: update to NSS 3.70

guillomep:
For HTTPS to work and also for it to deliver the security promises we
need to ship the latest certificate database.
Update the package version symlink in oder to fetch the database from
the newest NSS release under
https://ftp.mozilla.org/pub/security/nss/releases/
and do a "ebuild ca-certificates-3.70.ebuild manifest" run.

## How to use

Check the image content changes and some HTTPS connections.
Port the update to all channels.
Optional: Automate this as GitHub Action (see flatcar-linux/Flatcar#533)


pothos:
Thanks, we should really automate this :)
I'll make a test build, too, before merging.

PR #1382, opened 2021-10-29
selinux-base: fix build issue during boostrap

jepio:
# selinux-base: fix build issue during boostrap

selinux-base needs python with xml support, but the dependency was implicit through policycoreutils, and we removed that dependency downstream. This shows up during stage3 of bootstrap_sdk. Fix the issue by adding an explicit dependency and removing the previous attempt at fixing the issue (forcing sequential build, which does nothing).

## How to use

`./boostrap_sdk`.

## Testing done

Will let CI run.


dongsupark (approved):
According to a different [CI run](http://jenkins.infra.kinvolk.io:8080/job/os/job/sdk/1448/), this PR indeed fixes the build issue.
Thanks a lot!

PR #1389, opened 2021-11-02
`dev-libs/openssl`: enable `fips` support

tormath1:
In this PR, we:
* sync `dev-libs/openssl` with the upstream to fetch this change: https://bugs.gentoo.org/820173
* cherry-pick the Flatcar changes to re-apply them
* add `fips` to `package.use` to enable `FIPS` provider for OpenSSL

## How to use

```
emerge-amd64-usr dev-libs/openssl
```

## Testing done

CI :large_blue_circle: : http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4047/cldsv/

FIPS provider can be correctly installed now since the module `.so` is available :
```
$ ls /usr/lib64/ossl-modules
fips.so  legacy.so
$ sudo openssl fipsinstall -out /etc/ssl/fipsmodule.cnf -module /usr/lib64/ossl-modules/fips.so
$ cat /etc/ssl/fipsmodule.cnf
[fips_sect]
activate = 1
...
```


jepio (approved):
Nicely done!

tormath1:
cherry-picked to:
* flatcar-3046

PR #1386, opened 2021-11-02
*: use https instead of git for Github URLs

dongsupark:
Replace `git://` with `https://` for Github URLs, because Github now rejects an unauthenticated git access.
See also https://github.blog/2021-09-01-improving-git-protocol-security-github/.

* app-admin/locksmith
* app-admin/mayday
* app-admin/sdnotify-proxy
* app-admin/toolbox
* app-admin/updateservicectl
* app-arch/torcx
* app-crypt/go-tspi
* app-emulation/acbuild
* app-emulation/actool
* coreos-base/afterburn
* coreos-base/coreos-cloudinit
* coreos-base/coreos-init
* coreos-base/emerge-gitclone
* coreos-base/nova-agent-watcher
* coreos-base/update-ssh-keys
* coreos-base/update_engine
* coreos-devel/fero-client
* coreos-devel/mantle
* sys-apps/baselayout
* sys-apps/ignition
* sys-apps/seismograph
* sys-boot/grub
* sys-boot/shim
* sys-kernel/bootengine
* sys-libs/nss-usrfiles

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/238.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4043/cldsv

PR #1394, opened 2021-11-03
*: *: use https instead of git for Github URLs (take 2)

dongsupark:
(This PR is another attempt of https://github.com/flatcar-linux/coreos-overlay/pull/1386, which was reverted)

Replace `git://` with `https://` for Github URLs, because Github now rejects an unauthenticated git access.
See also https://github.blog/2021-09-01-improving-git-protocol-security-github/.

In addition, fix a build issue during SDK stage2, where git + libcurl could not fetch from `https://` URLs.
To resolve the issue, we need to install curl with `BOOTSTRAP_USE=ssl` before trying to install baselayout.
Also we need to set `CURL_SSL=openssl` as required by curl. Using a USE_EXPAND variable `curl_ssl_openssl` in `BOOTSTRAP_USE`, we can specify the correct `CURL_SSL` variable in curl.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/240.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4063/cldsv

### DIFF
--- a/app-admin/locksmith/locksmith-9999.ebuild
+++ b/app-admin/locksmith/locksmith-9999.ebuild
@@ -4,7 +4,7 @@
 EAPI=7
 CROS_WORKON_PROJECT="flatcar-linux/locksmith"
 CROS_WORKON_LOCALNAME="locksmith"
-CROS_WORKON_REPO="git://github.com"
+CROS_WORKON_REPO="https://github.com"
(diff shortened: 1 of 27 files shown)


tormath1 (approved):

dongsupark:
cherry-picked to `flatcar-lts-2605`, `flatcar-2983`, `flatcar-3033`, `flatcar-3046`.

PR #1395, opened 2021-11-03
sys-apps/baselayout: Add sgx group

krnowak:
Systemd during the initrd stage was complaining about the missing
group, which resulted in ignoring some of the udev rules. Let's
placate it by adding sgx to baselayout, so the group is available
during the initrd stage too.

Pulls in https://github.com/flatcar-linux/baselayout/pull/20.

Tested locally.


dongsupark (approved):
Looks good.

A minor thing.
Can you please add https://github.com/flatcar-linux/baselayout/pull/20 in the description?
That would help so much the person who creates release notes, to track the relation between the PRs.

krnowak:
> Looks good.
> 
> A minor thing. Can you please add [flatcar-linux/baselayout#20](https://github.com/flatcar-linux/baselayout/pull/20) in the description? That would help so much the person who creates release notes, to track the relation between the PRs.

Sure, done. Thanks for the review.

PR #1396, opened 2021-11-03
dev-libs/libgpg-error: Remove obsolete package

krnowak:
We have a newer copy in portage-stable, which is being picked up
during build.

To be merged together with https://github.com/flatcar-linux/portage-stable/pull/241.

CI: http://localhost:9091/job/os/job/manifest/4065/cldsv/


krnowak:
CI passed.

jepio (approved):
Looks like the only downstream change is a cross-compile patch that is included in gentoo versions. (e8fe1ef7f2c3279f48b0e9ed77bf2ce23f62a218). Looks good.

PR #1397, opened 2021-11-03
sys-kernel/bootengine: Stop using deprecated actions in veritysetup

krnowak:
The "create" action became "open", and "remove" became "close". Also
reorder the parameters accordingly (it's a bit different for "open" vs
"create"). Also put the options before specifying the action.

Pulls in https://github.com/flatcar-linux/bootengine/pull/31.

CI: http://localhost:9091/job/os/job/manifest/4066/cldsv/


krnowak:
Bootengine PR merged.

krnowak:
> Thanks, it looks good.

Thanks for the review.

> So if I understand correctly, `veritysetup create` does not work any more with cryptsetup 2.3.6, right? Then should we probably backport the fix to stable, beta, and alpha?

It has worked so far (so still works with cryptsetup 2.4.1 that we currently have in main). It's just deprecated. I only noticed it because I was investigating a hang during boot. So no need to backport.

PR #1445, opened 2021-11-25
profiles: delete unnecessary keywords for gcc

dongsupark:
Now that gcc 9.4.0 already has stable keywords both amd64 and arm64, we do not need to accept keywords in profiles.
Simply delete.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/252 .

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4238/cldsv

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) - done in https://github.com/flatcar-linux/portage-stable/pull/252


krnowak (approved):
Looks good, thanks.

PR #1456, opened 2021-12-01
app-arch/torcx: Fix wrong reference to Docker 19.03

pothos:
When selecting the docker-1.12-no profile, torcx failed because the
profile looked for 19.03 instead of 20.10.

Make the docker-1.12-no profile identical to the vendor profile so
that we don't have to update it.

## How to use

Check that
```
echo docker-1.12-no | sudo tee /etc/torcx/next-profile
sudo reboot
```
works again.

Backport to Stable, Beta, Alpha.

## Testing done

build scheduled, ran the above test

### DIFF
--- a/app-arch/torcx/files/docker-1.12-no.json
+++ b/app-arch/torcx/files/docker-1.12-no.json
@@ -4,7 +4,7 @@
         "images": [
             {
                 "name": "docker",
-                "reference": "19.03"
+                "reference": "com.coreos.cl"
             }
         ]
     }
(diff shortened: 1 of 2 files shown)


jepio (approved):
What's the role of the `docker-1.12` profile anyway?

pothos:
> What's the role of the `docker-1.12` profile anyway?

Torcx knows about the file `/etc/flatcar/docker-1.12` which can have `yes` or `no` and I think `/etc/torcx/next-profile` can also have `docker-1.12-yes` or `docker-1.12-no` :shrug:

PR #1470, opened 2021-12-02
.github: Try to get a lwn link for kernel release

krnowak:
Tested on my fork. Results in a PR like this: https://github.com/krnowak/coreos-overlay/pull/2/files

### DIFF
--- a/.github/workflows/kernel-apply-patch.sh
+++ b/.github/workflows/kernel-apply-patch.sh
@@ -36,7 +36,17 @@ done
 
 popd >/dev/null || exit
 
-URL="https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tag/?h=v${VERSION_NEW}"
+if ! curl -sA 'Chrome' -L 'http://www.google.com/search?hl=en&q=site%3Alwn.net+linux+'"${VERSION_NEW}" -o search.html; then
+  echo 'curl failed'
+  touch search.html
+fi
+# can't use grep -m 1 -o … to replace head -n 1, because all the links
+# seem to happen in one line, so grep prints all the links in the line
+URL=$({ grep -o 'https://lwn.net/Articles/[0-9]\+' search.html || true ; } | head -n 1)
+if [[ ! "${URL}" ]]; then
+    URL="https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tag/?h=v${VERSION_NEW}"
+fi
+rm search.html
 
 generate_update_changelog 'Linux' "${VERSION_NEW}" "${URL}" 'linux'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/kernel-apply-patch.sh
```diff
@@ -36,7 +36,12 @@ done
 
 popd >/dev/null || exit
 
-URL="https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tag/?h=v${VERSION_NEW}"
+curl -sA 'Chrome' -L 'http://www.google.com/search?hl=en&q=site%3Alwn.net+linux+'"${VERSION_NEW}" -o search.html
+URL=$(grep -o 'https://lwn.net/Articles/[0-9]\+' search.html | head -n 1)
```
pothos, line 40: The `set -euo pipefail` will make this line fail if grep doesn't find anything, and the script exits.
```suggestion
URL=$(grep -m 1 -o 'https://lwn.net/Articles/[0-9]\+' search.html || true)
```

krnowak:
```
$ curl -sA "Chrome" -L 'http://www.google.com/search?hl=en&q=site%3Alwn.net+linux+5.15.6' -o search.html
$ grep -m 1 -o 'https://lwn.net/Articles/[0-9]\+' search.html 
https://lwn.net/Articles/877286
https://lwn.net/Articles/876861
https://lwn.net/Articles/877285
https://lwn.net/Articles/876976
https://lwn.net/Articles/484685
```

So, didn't work, because apparently everything is just one line. I'll look into making it more robust.

pothos: Ok, updated the suggestion

krnowak: Heh, I went with parentheses instead of braces, but in the end applied your suggestion too. But surely braces have some weird syntax quirks.

.github/workflows/kernel-apply-patch.sh
```diff
@@ -36,7 +36,17 @@ done
 
 popd >/dev/null || exit
 
-URL="https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tag/?h=v${VERSION_NEW}"
+if ! curl -sA 'Chrome' -L 'http://www.google.com/search?hl=en&q=site%3Alwn.net+linux+'"${VERSION_NEW}" -o search.html; then
```
pothos, line 39:
```suggestion
if ! curl -sfA 'Chrome' -L 'http://www.google.com/search?hl=en&q=site%3Alwn.net+linux+'"${VERSION_NEW}" -o search.html; then
```
For the return code to be set on `404` you also need `-f`.

krnowak: Sure. Curl can fail for multitude of reasons, which is why I've put it into an `if`. `404` will result in grep not finding anything.

pothos (approved):

PR #1087, opened 2021-07-01
Upgrade to glibc-2.33

t-lo:
This PR upgrades (once more) to glibc-2.33. 

In addition to https://github.com/kinvolk/coreos-overlay/pull/1067, it fixes an issue with `newtmpfies` which led to silent breakage during build - we now use `systemd_newtmpfilesd` as we should have from the start.

It depends on https://github.com/kinvolk/portage-stable/pull/183, which fixes an issue with `scanelf` and glibc-2.33, leading to silent breakage in *every* package build (i.e. binaries were not stripped).

CI :large_blue_circle: : http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2946/cldsv/

PR #1475, opened 2021-12-03
Port packages using tmpfiles from systemd eclass to tmpfiles eclass

krnowak:
Functions dealing with tmpfiles in systemd eclass were deprecated in favor of the same functionality in tmpfiles eclass. This PR moves the packages using the deprecated things to shiny new things.

Should be merged together with https://github.com/flatcar-linux/portage-stable/pull/257.

CI: http://localhost:9091/job/os/job/manifest/4403/cldsv/


krnowak:
CI passed.

sys-libs/glibc/glibc-2.33-r8.ebuild
```diff
@@ -1392,7 +1392,7 @@ glibc_do_src_install() {
 		sed -i "${nscd_args[@]}" "${ED}"/etc/init.d/nscd
 
 		use systemd && systemd_dounit nscd/nscd.service
-		systemd_newtmpfilesd nscd/nscd.tmpfiles nscd.conf
+		newtmpfiles nscd/nscd.tmpfiles nscd.conf
```
dongsupark, line 1395: Confused.
Gentoo uses the default `newtmpfiles`.
But Flatcar has been applying its own [patch](https://github.com/flatcar-linux/coreos-overlay/pull/1087/commits/8d040f93c289cd696db714ac8ae92d55e04171ac) to replace `newtmpfiles` with `systemd_newtmpfilesd`.
Does it mean that we do not need to keep the Flatcar patch for glibc?

krnowak: So it seems that the reason for replacing `newtmpfiles` with `systemd_newtmpfilesd` was to avoid circular dep between glibc and system, because `tmpfiles` eclass was adding `virtual/tmpfiles` into dependencies (`virtual/tmpfiles` pulled in systemd), while `systemd` eclass was not adding any new deps by inheriting it. The `tmpfiles` eclass got recently a new option (`TMPFILES_OPTIONAL`) that allows us to skip adding `virtual/tmpfiles`, so we can drop this change from glibc.

dongsupark: Ok, I believe you are correct. We can drop the newtmpfiles from the glibc patch.
Still, I am afraid, the next person who upgrades glibc could probably get into a rabbit hole to figure out which parts of the big chunk of the [patch](https://github.com/flatcar-linux/coreos-overlay/pull/1087/commits/8d040f93c289cd696db714ac8ae92d55e04171ac) are still valid.
Ideally we should have 2 commits for glibc, one for plain sync commit with Gentoo, the other for all Flatcar changes.

Same applies maybe for other ones in the other 20 commits.

Also I am not sure I like bumping the revision like `-r7` to `-r8` for every commit, especially for glibc.
In the past revisions in glibc 2.33 in Gentoo had quite special meaning. `-r5` was to address specific CVEs, and `-r7` was to address other CVEs, etc.
Worse, that revision number did not match with `PATCH_VER` to be used for the patch tarball.
Now we bumped the revision to -r8, and let's say Gentoo suddenly releases 2.33-r8 tomorrow. That must be fun.
Not sure that would be the case as well for glibc 2.34 though.

krnowak: > Ok, I believe you are correct. We can drop the newtmpfiles from the glibc patch. Still, I am afraid, the next person who upgrades glibc could probably get into a rabbit hole to figure out which parts of the big chunk of the [patch](https://github.com/flatcar-linux/coreos-overlay/pull/1087/commits/8d040f93c289cd696db714ac8ae92d55e04171ac) are still valid. Ideally we should have 2 commits for glibc, one for plain sync commit with Gentoo, the other for all Flatcar changes.
> 
> Same applies maybe for other ones in the other 20 commits.

Sure, I can do that.

> 
> Also I am not sure I like bumping the revision like `-r7` to `-r8` for every commit, especially for glibc. In the past revisions in glibc 2.33 in Gentoo had quite special meaning. `-r5` was to address specific CVEs, and `-r7` was to address other CVEs, etc. Worse, that revision number did not match with `PATCH_VER` to be used for the patch tarball. Now we bumped the revision to -r8, and let's say Gentoo suddenly releases 2.33-r8 tomorrow. That must be fun. Not sure that would be the case as well for glibc 2.34 though.

I would say that when an ebuild is modified, its revision should be bumped. But it's icky in case of forked packages like glibc, true. Maybe we could use one of two ways out of it:

1. For packages that are forked from gentoo, multiply their revision by 10 or 100 and bump the revision every time we modify it. Cons: things like `PATCH_VER` or generally the `PVR` and `PF` variables don't have the contents expected by the build (for example it expected `PVR` to be `2.33-r7`, not `2.33-r701`).
2. Modify portage by adding subrevision? Something like `glibc-2.33-r7.1`. That would still be newer than `r7`, but older than `r8`. Cons: another patch to carry in portage, I suppose.

My proposal would be that for now let's not worry about it and then get the revbumping sorted out in new year.

dongsupark: > For packages that are forked from gentoo, multiply their revision by 10 or 100 and bump the revision every time we modify it. Cons: things like PATCH_VER or generally the PVR and PF variables don't have the contents expected by the build (for example it expected PVR to be 2.33-r7, not 2.33-r701).
> Modify portage by adding subrevision? Something like glibc-2.33-r7.1. That would still be newer than r7, but older than r8. Cons: another patch to carry in portage, I suppose.

TBH both look ugly. ;-)

Ideally we should update each ebuild to the next version (like glibc 2.34) as often as possible, without having to worry about damn complicated revision numbers.
And we are already doing that for many packages.
However, this is glibc, one of the heart of OS. That's why frequent version bumps are not trivial.
I would also hope the Gentoo glibc maintainers stop using different PATCH_VER in the first place. But anyway there is nothing we can do about that.

krnowak: > > For packages that are forked from gentoo, multiply their revision by 10 or 100 and bump the revision every time we modify it. Cons: things like PATCH_VER or generally the PVR and PF variables don't have the contents expected by the build (for example it expected PVR to be 2.33-r7, not 2.33-r701).
> > Modify portage by adding subrevision? Something like glibc-2.33-r7.1. That would still be newer than r7, but older than r8. Cons: another patch to carry in portage, I suppose.
> 
> TBH both look ugly. ;-)

:grin:

I had a look at [ebuild revisions doc](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html) and actually not all changes warrant a revbump. TIL. I also had an idea of using [the `_p` suffix in package version](https://devmanual.gentoo.org/ebuild-writing/file-format/index.html#file-naming-rules), but that's worse than the above, so I'll shut up now. :)

> 
> Ideally we should update each ebuild to the next version (like glibc 2.34) as often as possible, without having to worry about damn complicated revision numbers. And we are already doing that for many packages. However, this is glibc, one of the heart packages of Linux OS. That's why frequent version bumps are not trivial. I would also hope the Gentoo glibc maintainers stop using different PATCH_VER in the first place. But anyway there is nothing we can do about that.

Alright, I'll drop the revbump for glibc package then. The change I made is rather trivial, anyway.

krnowak:
@dongsupark: So for most of the packages I went with updating by resetting the package to vanilla state from gentoo (without actually updating it to a newer version) and then applying our changes. But I really wasn't sure what to do with some revision bumps that happened before (see the renames of `net-fs/samba`, `sys-libs/glibc` and `sys-libs/libsemanage`). For packages that aren't forks of gentoo packages I bumped the revision (with an exception of `sys-apps/systemd` - I revbumped that one too - it is technically a gentoo fork, but basically a rewrite).

dongsupark:
Thanks for updating that.

> But I really wasn't sure what to do with some revision bumps that happened before (see the renames of net-fs/samba, sys-libs/glibc and sys-libs/libsemanage).

Changes in `sys-libs/glibc` and `sys-libs/libsemanage` are ok.

But the rename of `net-fs/samba` looks strange. From 4.12.9-r3 to 4.12.9-r1?
I would say, let's make it 4.12.9-r2, the same as flatcar-3087 and many other previous releases.

krnowak:
> Thanks for updating that.
> 
> > But I really wasn't sure what to do with some revision bumps that happened before (see the renames of net-fs/samba, sys-libs/glibc and sys-libs/libsemanage).
> 
> Changes in `sys-libs/glibc` and `sys-libs/libsemanage` are ok.

Cool.

> 
> But the rename of `net-fs/samba` looks strange. From 4.12.9-r3 to 4.12.9-r1? I would say, let's make it 4.12.9-r2, the same as flatcar-3087 and many other previous releases.

Alright. Just for the record - r1 to r2 happened in 29f37d36731f665879a839d17b7827149df1086d, and r2 to r3 happened (unnecessarily) in c56f51654268e459ba154a3203adeb435a3c9e31.

krnowak:
CI passed.

dongsupark (approved):

PR #1484, opened 2021-12-07
.github/wf: add mirror-calico action

tormath1:
this action will sync our ghcr calico images with upstream docker
images.

<hr>

`kubeadm.*.calico` tests are using Calico images from GHCR in order to circumvent Docker hub pulling rate limit, the issue is that we can't really pin the image tag used by Calico tests (see: https://github.com/projectcalico/calico/issues/5121) so each time Calico is releasing, CI is failing while images are not updated into GHCR.
It should be a temporary thing :tm: while https://github.com/projectcalico/calico/issues/5121 is being handled :) 

Script is based on internal custom scripts we were currently using. 

Tested on my fork: https://github.com/tormath1/coreos-overlay/actions/runs/1548773225

### DIFF
--- a/.github/workflows/mirror-calico.yml
+++ b/.github/workflows/mirror-calico.yml
@@ -1,0 +1,26 @@
+name: Sync GHCR Calico images with Docker Hub
+on:
+  schedule:
+    # run every 12h
+    - cron:  '0 */12 * * *'
+  workflow_dispatch:
+
+jobs:
+  get-docker-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login to GitHub Container Registry (ghcr)
+        run: echo ${{ secrets.GHCR_PASSWORD }} | docker login ghcr.io -u ${{ secrets.GHCR_USERNAME }} --password-stdin
+      - name: Fetch latest Calico release
+        id: fetch-latest-release
+        run: |
+          set -ex
+
+          git clone https://github.com/projectcalico/calico calico
+          # get latest version
+          version=$(git -C "$_" tag | sort -V | tail -1)
+
+          pushd .github/workflows/
+          ./mirror-calico.sh $version
+          popd
--- a/.github/workflows/mirror-to-ghcr.sh
+++ b/.github/workflows/mirror-to-ghcr.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# This generic script aims to mirror an image from Docker hub to another registry.
+# Authentication to the registry must be done before.
+
+image=$1
+imagetag=$2
+registry=${3:-kinvolk}
+
+# we want both arch for running tests
+platforms=( amd64 arm64 )
+
+# tags will hold the mirrored images
+tags=()
+
+for platform in "${platforms[@]}"; do
+  # we first fetch the image from Docker Hub
+  var=$(docker pull $image:$imagetag --platform=linux/$platform -q)
+  # we prepare the image to be pushed into another registry
+  tag=ghcr.io/$registry/$image:$imagetag-$platform
+  # we tag the image to create the mirrored image
+  docker tag $var $tag
+  docker push $tag
+  tags+=( $tag )
+done
+
+docker manifest create ghcr.io/$registry/$image:$imagetag "${tags[@]}"
+# some images have bad arch specs in the individual image manifests :(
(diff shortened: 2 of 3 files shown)


.github/workflows/mirror-calico.yml
```diff
@@ -0,0 +1,57 @@
...
+    # run every 12h
+    - cron:  '0 */12 * * *'
+  workflow_dispatch:
+
+jobs:
+  get-docker-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to GitHub Container Registry (ghcr)
+        run: echo ${{ secrets.GHCR_PASSWORD }} | docker login ghcr.io -u ${{ secrets.GHCR_USERNAME }} --password-stdin
```
jepio, line 13: Where do these secrets come from? The other way to do this would be to link the packages to this repository and then specify permissions on the default github token for the workflow.

tormath1:
> Would it be ok to copy the body of the script into a separate bash file in `.github/workflows` and call that?

@jepio in this case I would be in favor of moving the two scripts into `./github/worflows` repository in order to not mix public/private repositories.

.github/workflows/mirror-calico.yml
tormath1: This secret needs to be manually added from the GH Admin settings page.

> specify permissions on the default github token for the workflow.

IMHO, I would prefer to split privileges in case we have a token leak.

.github/workflows/mirror-calico.yml
```diff
@@ -0,0 +1,57 @@
...
+
+          # we want both arch for running tests
+          platforms=(
+            amd64
+            arm64
+          )
+
+          # we iterate over the images we want to mirror
+          for image in "${images[@]}"; do
+            # tags will hold the images the mirrored images
```
krnowak, line 40: Not sure about this one.

```suggestion
            # tags will hold the names of the mirrored images
```

krnowak (commented):
Looks alright, found two small issues.

.github/workflows/mirror-calico.yml
jepio: ok

jepio:
> > Would it be ok to copy the body of the script into a separate bash file in `.github/workflows` and call that?
> 
> @jepio in this case I would be in favor of moving the two scripts into `./github/worflows` repository in order to not mix public/private repositories.

Yeah that makes sense

.github/workflows/mirror-calico.yml
tormath1: It's actually:
```suggestion
            # tags will hold the mirrored images
```

krnowak: Sounds good.

tormath1:
@jepio @krnowak thanks for you reviews. Added and commented scripts from `stash` into `.github/workflows` and updated the action to use this scripts.

.github/workflows/mirror-calico.yml
```diff
@@ -0,0 +1,27 @@
...
+        id: fetch-latest-release
+        run: |
+          set -ex
+
+          git clone https://github.com/projectcalico/calico calico
+          # get latest version
+          version=$(git -C "$_" tag | sort -V | tail -1)
+
+          pushd .github/workflows/
+          chmod u+x ./mirror-to-ghcr.sh ./mirror-calico.sh
```
krnowak, line 25: This should not be necessary, right?

.github/workflows/mirror-to-ghcr.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# This generic script aims to mirror an image from Docker hub to another registry.
+# Authentication to the registry must be done before.
+
+image=$1
+imagetag=$2
+registry=${3:-kinvolk}
+: ${tag:=latest}
```
krnowak, line 9: This line is not necessary - `tag` variable will be clobbered in the loop anyway.

.github/workflows/mirror-calico.yml
tormath1: That's right, scripts already hold `x` flag: 
```
$ ls -ali .github/workflows/mirror-*.sh
73843545 -rwxr-xr-x 1 mathieu mathieu  439 Dec  7 14:22 .github/workflows/mirror-calico.sh
73842802 -rwxr-xr-x 1 mathieu mathieu 1021 Dec  7 14:18 .github/workflows/mirror-to-ghcr.sh
```

krnowak (approved):
Looks good.

PR #1495, opened 2021-12-13
sys-auth/google-oslogin: accept ssh-rsa in sshd_config for openssh 8.8

dongsupark:
Temporarily accept `ssh-rsa` algorithm in `sshd_config` for openssh >= 8.8, until most ssh clients could deprecate ssh-rsa.

It is the same fix as https://github.com/flatcar-linux/init/pull/54.
However, we should do that again for GCE, because the google-oslogin ebuild overwrites the existing `sshd_config`.

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4383/cldsv


dongsupark:
CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4383/cldsv

dongsupark:
cherry-picked to flatcar-3087

PR #1502, opened 2021-12-15
sys-apps/policycoreutils: update correct SELinux store

t-lo:
The policycoreutils ebuild calls `semodule` in postinst to update SELinux stores.
It does not, however, tells `semodule` the correct ROOT to use, so builds that go into `/build/[arch]-usr` end up updating the SDK's store.
Fixes 
   libsemanage.semanage_commit_sandbox: Error while renaming /var/lib/selinux/targeted/active to /var/lib/selinux/targeted/previous. (Invalid cross-device link)
observed when using the SDK Container to build the OS image.
It now also updates the correct store, which it previously did not.

We should consider cherry-picking this patch to flatcar-3033, flatcar-3066, and flatcar-3087.

<hr>

CI (running): http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4397/cldsv/

### DIFF
--- a/changelog/bugfixes/2021-12-16-policycoreutils-fix-semodule-postinst.md
+++ b/changelog/bugfixes/2021-12-16-policycoreutils-fix-semodule-postinst.md
@@ -1,0 +1,10 @@
+The policycoreutils ebuild calls semodule in postinst to update SELinux stores.
+It does not, however, tells semodule the correct ROOT to use, so builds that go into /build/[arch]-usr end up updating the SDK's store.
+This patch resolves the following error message:
+```
+$ emerge-amd64-usr policycoreutils
+[...]
+libsemanage.semanage_commit_sandbox: Error while renaming /var/lib/selinux/targeted/active to /var/lib/selinux/targeted/previous. (Invalid cross-device link)
+```
+The error is observed when using the SDK Container to build an OS image.
+The `semanage` run in policycoreutilsi' `postinst`  now also updates the correct store, which it previously did not.
--- a/sys-apps/policycoreutils/policycoreutils-3.1-r3.ebuild
+++ b/sys-apps/policycoreutils/policycoreutils-3.1-r3.ebuild
@@ -219,7 +219,7 @@ pkg_postinst() {
 	for POLICY_TYPE in ${POLICY_TYPES} ; do
 		# There have been some changes to the policy store, rebuilding now.
 		# https://marc.info/?l=selinux&m=143757277819717&w=2
-		einfo "Rebuilding store ${POLICY_TYPE} (without re-loading)."
-		semodule -s "${POLICY_TYPE}" -n -B || die "Failed to rebuild policy store ${POLICY_TYPE}"
+		einfo "Rebuilding store ${POLICY_TYPE} in '${ROOT}' (without re-loading)."
+		semodule -S "${ROOT}" -s "${POLICY_TYPE}" -n -B || die "Failed to rebuild policy store ${POLICY_TYPE}"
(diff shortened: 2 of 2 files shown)


sys-apps/policycoreutils/policycoreutils-3.1-r3.ebuild
```diff
@@ -219,7 +219,7 @@ pkg_postinst() {
 	for POLICY_TYPE in ${POLICY_TYPES} ; do
 		# There have been some changes to the policy store, rebuilding now.
 		# https://marc.info/?l=selinux&m=143757277819717&w=2
-		einfo "Rebuilding store ${POLICY_TYPE} (without re-loading)."
-		semodule -s "${POLICY_TYPE}" -n -B || die "Failed to rebuild policy store ${POLICY_TYPE}"
+		einfo "Rebuilding store ${POLICY_TYPE} in '${ROOT}' (without re-loading)."
+		semodule -S "${ROOT}" -s "${POLICY_TYPE}" -n -B || die "Failed to rebuild policy store ${POLICY_TYPE}"
```
krnowak, line 223: I think that ROOT can be empty if root is actually `/`, so I think we should guard against this case too (if we want to upstream this to gentoo too).

```suggestion
		einfo "Rebuilding store ${POLICY_TYPE} in '${ROOT%/}' (without re-loading)."
		semodule -S "${ROOT%/}" -s "${POLICY_TYPE}" -n -B || die "Failed to rebuild policy store ${POLICY_TYPE}"
```

krnowak: Eeek, I got it wrong, should be:
```suggestion
		einfo "Rebuilding store ${POLICY_TYPE} in '${ROOT:-/}' (without re-loading)."
		semodule -S "${ROOT:-/}" -s "${POLICY_TYPE}" -n -B || die "Failed to rebuild policy store ${POLICY_TYPE}"
```

krnowak: See the [second implication in the ultimate EAPI7 guide](https://mgorny.pl/articles/the-ultimate-guide-to-eapi-7.html#d-ed-root-eroot-no-longer-have-a-trailing-slash).

t-lo: Good catch Krzesimir! I'll go with it.

t-lo:
Rebased on latest main (for the containerd-1.5.8 tarball checksum fix).

t-lo:
Added a bugfix changelog file.

t-lo:
Build is successful, will merge and cherry-pick for flatcar-3033, flatcar-3066, and flatcar-3087.

changelog/bugfixes/2021-12-16-policycoreutils-fix-semodule-postinst.md
```diff
@@ -0,0 +1,10 @@
+The policycoreutils ebuild calls semodule in postinst to update SELinux stores.
```
pothos, line 1: This changelog entry format is not really useful for the release notes. This paragraph is good for the PR description or a commit message, but here should be a bullet point for the release notes, e.g., `- SDK: Fixed build error popping up in the new SDK Container because `policycoreutils` used the wrong ROOT to update the SELinux store ([PR#1502](https://github.com/flatcar-linux/coreos-overlay/pull/1502))`
I'll file a PR for that here: https://github.com/flatcar-linux/coreos-overlay/pull/1537

PR #1517, opened 2021-12-20
sys-kernel/bootengine: make dracut install loop driver

dongsupark:
Unlike with Kernel 5.10, dracut does not automatically install `loop.ko` with Kernel 5.15.
Explicitly install the loop module from the dracut command line.

Pulls in https://github.com/flatcar-linux/bootengine/pull/32

## Testing done

CI passed for amd64: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4434/cldsv/

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) (done in https://github.com/flatcar-linux/bootengine/pull/32)


pothos:
When testing you need to generate not only the `packet` image but also the `pxe` image.
Since we touch something initramfs related I suggest to do a test for all platforms.

tormath1 (approved):
:rocket:

dongsupark:
cherry-picked to flatcar-3087

PR #1518, opened 2021-12-20
.github: Reflect more kernel versions in the changelog

krnowak:
It happens that kernel update PRs are created faster than we merge
them. In such case we create version gaps in the changelog. Remedy
that by adding links to all the released kernel versions between the
current one and the just released one.

For an example of a PR created by these changes please see https://github.com/krnowak/coreos-overlay/pull/6


pothos (approved):
Thanks!

PR #1519, opened 2021-12-21
eclass: Some updates for newer EAPI versions

krnowak:
Also drop the unused `git` eclass (we only use `git-r3` now).

To be merged with https://github.com/flatcar-linux/portage-stable/pull/262.

CI: http://localhost:9091/job/os/job/manifest/4540/cldsv/


krnowak:
CI passed.

jepio (approved):

PR #1525, opened 2021-12-21
Get rid of EAPI 0

krnowak:
With the recent eclass update I noticed that I missed one class of packages - packages whose ebuilds didn't even define the EAPI variable (so technically their EAPI was zero). Fortunately there were not that many of them.

Here we reflect the removal of `virtual/cdrtools` in favor of direct use of `app-cdr/cdrtools`, as gentoo is doing. Also, we drop `sys-libs/db` package completely - it's ancient (and unmodified since the import from chromiumos), and nothing is actually pulling it (if something would, we would certainly be getting deps problems, as some of the conditional deps insist on having sys-apps/db-4.8, which we didn't even have).

To be merged together with https://github.com/flatcar-linux/portage-stable/pull/264.

CI: http://localhost:9091/job/os/job/manifest/4592/cldsv/

- [x] https://github.com/flatcar-linux/coreos-overlay/pull/1548 is merged


krnowak:
CI passed.

tormath1 (approved):
OK LGTM, `virtual/cdrtools` is only explicitly required by `coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild` - too bad the CI is gone though.

PR #1542, opened 2022-01-04
sys-kernel/coreos-modules: explicitly enable EFI_GENERIC_STUB_INITRD_CMDLINE_LOADER

jepio:
# enable processing of `initrd=0` kernel param on all platforms

This is fixes tests on c3.large.arm machines on Equinix Metal.

## How to use

Rebuild kernel: `emerge-arm64-usr coreos-kernel coreos-modules coreos-sources`.

## Testing done

CI running, instances boot on EM.

- [X] no changelog entries because this is a bugfix for an issue that hasn't been released yet.


sys-kernel/coreos-modules/files/commonconfig-5.15
```diff
@@ -170,6 +170,7 @@ CONFIG_EDAC=m
 CONFIG_EFI=y
 CONFIG_EFIVAR_FS=y
 CONFIG_EFI_STUB=y
+CONFIG_EFI_GENERIC_STUB_INITRD_CMDLINE_LOADER=y
```
krnowak, line 173: Please keep the file contents sorted, so I think this line should go up one line. Or use the `sort_config.sh` script in the same directory. :)

pothos (approved):

PR #1543, opened 2022-01-05
Migrate to libxcrypt

krnowak:
The crypt functionality in glibc is deprecated, let's move to libxcrypt. Gentoo already did that, so we would need to eventually follow suit. Let's do it now then.

Should be merged with https://github.com/flatcar-linux/portage-stable/pull/267.

CI: http://localhost:9091/job/os/job/manifest/4524/cldsv/

### DIFF
--- a/profiles/coreos/base/package.mask
+++ b/profiles/coreos/base/package.mask
@@ -14,3 +14,10 @@
 # Since version 2, it tries to write liblto symlinks with absolute paths that
 # don't work when building for the board root directories.
 >=sys-devel/gcc-config-2
+
+# Overwrite portage-stable mask. We are delaying the transition to
+# libxcrypt, because we need to figure out how to solve the dep loop
+# that results from the migration (python -> virtual/libcrypt ->
+# libxcrypt -> glibc -> python), and also we need to update gcc to
+# version 10 or later.
+>=virtual/libcrypt-2
--- a/profiles/coreos/base/package.unmask
+++ b/profiles/coreos/base/package.unmask
@@ -5,3 +5,10 @@
 # gentoo. We still need it, since sys-libs/libsemanage still requires
 # it. When we update selinux, this can be dropped.
 =dev-libs/ustr-1.0.4-r8
+
+# Overwrite portage-stable mask. We are delaying the transition to
+# libxcrypt, because we need to figure out how to solve the dep loop
+# that results from the migration (python -> virtual/libcrypt ->
+# libxcrypt -> glibc -> python), and also we need to update gcc to
+# version 10 or later.
+=virtual/libcrypt:0/1
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -142,3 +142,7 @@ dev-libs/openssl fips
 
 # enables ELF support to e.g. allow tc to handle BPF filters.
 sys-apps/iproute2 elf
+
+# disable crypt functionality from glibc in favor of libxcrypt
+sys-libs/glibc -crypt
+sys-libs/libxcrypt system
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -142,7 +142,3 @@ dev-libs/openssl fips
 
 # enables ELF support to e.g. allow tc to handle BPF filters.
 sys-apps/iproute2 elf
-
-# disable crypt functionality from glibc in favor of libxcrypt
-sys-libs/glibc -crypt
-sys-libs/libxcrypt system
--- a/profiles/coreos/base/package.use.force
+++ b/profiles/coreos/base/package.use.force
@@ -1,10 +1,6 @@
 # Copyright (c) 2014 The CoreOS Authors. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
-# Force libcrypt so it's included in stage-1 of the SDK build.
-# +crypt was introduced in glibc-2.30 and above.
-sys-libs/glibc crypt
-
 # Do not force this flag, we don't need XATTR_PAX
 sys-apps/portage -xattr
 
--- a/profiles/coreos/base/package.use.force
+++ b/profiles/coreos/base/package.use.force
@@ -1,6 +1,10 @@
 # Copyright (c) 2014 The CoreOS Authors. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
+# Force libcrypt so it's included in stage-1 of the SDK build.
+# +crypt was introduced in glibc-2.30 and above.
+sys-libs/glibc crypt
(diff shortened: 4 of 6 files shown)


krnowak:
Not gonna happen for now - need to figure out how to break a new dep loop (`dev-lang/python` -> `virtual/libcrypt` -> `sys-libs/libxcrypt` -> `sys-libs/glibc` -> `dev-lang/python`), maybe we will be able to break it at `sys-libs/libxcrypt` or add an extra step to script to follow the `Early migration` part in the [libxcrypt migration Gentoo news](https://www.gentoo.org/support/news-items/2021-10-18-libxcrypt-migration-stable.html). Second blocker is that `sys-libs/libxcrypt` seems to require gcc >= 10, while we are still at 9.

thesamesam:
- On the circular dependency, this is kind of interesting. It should only matter on upgrades where you:
1. have a newer `dev-lang/python` that is not yet emerged
2. ... said Python version is in `PYTHON_COMPAT` of e.g. glibc (you could try dropping it from libxcrypt instead)

You can either drop the relevant Python version from `PYTHON_COMPAT` for a bit (in Gentoo, we assume everyone should have Python 3.9 by now, or will soon enough, but we cannot assume Python 3.10, so we dropped it from glibc's for now: see [here](https://gitweb.gentoo.org/repo/gentoo.git/commit/sys-libs/glibc?id=3d081cb997d18d200c71e1ff394e17db2f01f38d)).

This behaviour is related to bug [702806](https://bugs.gentoo.org/702806).

But you should be able to try that change to the ebuilds or masking the newer version it tries to pull in for a bit or --exclude python:3.10 (or whatever the version is).

TL;DR: It tries to use the latest possible Python listed in `PYTHON_COMPAT` even if it's not set as a target, so you can workaround it by either removing it from the listing, masking it temporarily, or simply `--excluding` it as a one-off (future world upgrades will be fine).

- You should be okay with GCC 9 for libxcrypt *provided you're not using distcc* (because it breaks detection of some of the support needed). In theory, it'd be fine if you had GCC 9 locally and on all of the distribution nodes too, FWIW. It's just that we had enough [reports](https://bugs.gentoo.org/823179) of it for me to add the warning.

krnowak:
> * On the circular dependency, this is kind of interesting. It should only matter on upgrades where you:
> 
> 
>     1. have a newer `dev-lang/python` that is not yet emerged
> 
>     2. ... said Python version is in `PYTHON_COMPAT` of e.g. glibc (you could try dropping it from libxcrypt instead)
> 
> 
> You can either drop the relevant Python version from `PYTHON_COMPAT` for a bit (in Gentoo, we assume everyone should have Python 3.9 by now, or will soon enough, but we cannot assume Python 3.10, so we dropped it from glibc's for now: see [here](https://gitweb.gentoo.org/repo/gentoo.git/commit/sys-libs/glibc?id=3d081cb997d18d200c71e1ff394e17db2f01f38d)).
> 
> This behaviour is related to bug [702806](https://bugs.gentoo.org/702806).
> 
> But you should be able to try that change to the ebuilds or masking the newer version it tries to pull in for a bit or --exclude python:3.10 (or whatever the version is).
> 
> TL;DR: It tries to use the latest possible Python listed in `PYTHON_COMPAT` even if it's not set as a target, so you can workaround it by either removing it from the listing, masking it temporarily, or simply `--excluding` it as a one-off (future world upgrades will be fine).

The loop that I had looks like this:

```
…

[nomerge       ] virtual/libcrypt-2:0/2::portage-stable [1:0/1::portage-stable] USE="-abi_riscv_lp64 -abi_riscv_lp64d -static-libs*" 
[ebuild  N     ]  sys-libs/libxcrypt-4.4.25-r1:0/1::portage-stable  USE="(compat) (split-usr) (system) -abi_riscv_lp64 -abi_riscv_lp64d -static-libs -test" 601 KiB
[ebuild  rR    ]   dev-lang/perl-5.34.0-r2:0/5.34::portage-stable  USE="minimal -berkdb -debug -doc -gdbm -ithreads" 12708 KiB
[ebuild   R    ]   sys-libs/glibc-2.33-r7:2.2::coreos  USE="multiarch nscd (ssp) (static-libs) systemd -audit -caps -cet -compile-locales (-crypt*) -custom-cflags -doc -gd -headers-only (-multilib) -multilib-bootstrap -profile -selinux -static-pie -suid -systemtap -test (-vanilla)" 16705 KiB
[ebuild  NS    ]    dev-lang/python-3.9.8:3.9::portage-stable [2.7.15:2.7::portage-stable, 3.6.5:3.6/3.6m::portage-stable] USE="gdbm hardened ncurses readline sqlite ssl xml -bluetooth% -build -examples -test -tk -verify-sig% -wininst (-ipv6%*) (-libressl%) (-threads%*)" 18714 KiB
[ebuild  r  U  ]     virtual/libcrypt-2:0/2::portage-stable [1:0/1::portage-stable] USE="-abi_riscv_lp64 -abi_riscv_lp64d -static-libs*" 0 KiB

…

 * Error: circular dependencies:

(dev-lang/python-3.9.8:3.9/3.9::portage-stable, ebuild scheduled for merge) depends on
 (virtual/libcrypt-2:0/2::portage-stable, ebuild scheduled for merge) (buildtime_slot_op)
  (sys-libs/libxcrypt-4.4.25-r1:0/1::portage-stable, ebuild scheduled for merge) (runtime)
   (sys-libs/glibc-2.33-r7:2.2/2.2::coreos, ebuild scheduled for merge) (buildtime)
    (dev-lang/python-3.9.8:3.9/3.9::portage-stable, ebuild scheduled for merge) (buildtime)
```

From the emerge output it looks like it's updating from python 3.6? Shouldn't be the case, an update to python 3.9.8 happened a while ago, so the SDK should have the python 3.9.8 already. Maybe the rebuild happens because of the virtual/libcrypt change. Anyway, for now I did the magic described in the Gentoo repo news to keep using the deprecated crypt functionality from glibc and that seemed to work. But I'll have a look again at the migration again soon. Thanks for help!
 
> * You should be okay with GCC 9 for libxcrypt _provided you're not using distcc_ (because it breaks detection of some of the support needed). In theory, it'd be fine if you had GCC 9 locally and on all of the distribution nodes too, FWIW. It's just that we had enough [reports](https://bugs.gentoo.org/823179) of it for me to add the warning.

Ha, right. I wrote about the blocker because I saw `if [[ ${BUILD_TYPE} != "binary" ]] && tc-is-gcc && [[ $(gcc-major-version) -lt 10 ]] ; then die "…"; fi` in the ebuild, but I missed that it was inside `if has "distcc" ${FEATURES} ; then` which I think is false for us. So it's actually probably not a blocker, but still not a reason to not update gcc to something newer. ;)

PR #1550, opened 2022-01-07
profiles/coreos/arm64: build target packages with '-moutline-atomics'

jepio:
# profiles/coreos/arm64: build target packages with '-moutline-atomics'

This should benefit more modern ARM systems (Altra/Graviton2) while keeping compatibility with older platforms (RPi4).
See commit message for a longer description.

## How to use

## Testing done


krnowak (approved):
CI seems to have passed, so the change looks good.

jepio:
If the GCC update gets merged this will be obsolete, as GCC-10 enables this option by default.

https://github.com/flatcar-linux/Flatcar/issues/549

krnowak:
I suppose this can be closed now?

PR #1562, opened 2022-01-12
changelog/README.md: specify current security fix section format

pothos:
The entries added in changelog/security/ do not follow our existing
security section in the release notes:
  https://www.flatcar.org/releases/#release-3033.2.0

Document the structure and an example to use the right format that we
need for release note generation.

## How to use

Pick this change for the scripts and portage-stable repos


pothos:
I'll also fix the existing entries now since we need them for channel progressions

dongsupark (approved):

sayanchowdhury (approved):

PR #1564, opened 2022-01-13
.github: find latest calico tag when mirroring

jepio:
# .github: find latest calico tag when mirroring

Workflow is currently failing because it finds the wrong tag, this commit fixes that.

## How to use

## Testing done

```
$ git clone https://github.com/projectcalico/calico calico
$ git -C calico tag --sort=-authordate | head -1
v3.21.3
```

### DIFF
--- a/.github/workflows/mirror-calico.yml
+++ b/.github/workflows/mirror-calico.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  get-docker-release:
+  mirror-calico:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror-calico.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
           git clone https://github.com/projectcalico/calico calico
           # get latest version
-          version=$(git -C "$_" tag | sort -V | tail -1)
+          version=$(git -C "$_" tag --sort=-authordate | head -1)
```
pothos, line 22: Do we need the latest X entries in case this job doesn't run often enough?

tormath1: The challenge with this mirroring thing is that calico is releasing in parallel v3.21.x and v3.22.x - we currently have only interest with v3.21.x.
So we could even add the following:
```suggestion
          version=$(git -C "$_" tag --sort=-authordate --list "v3.21*" | head -1)
```

pothos: If we don't want to maintain this here, we should probably mirror all release tags. I would take the last 10 entries and iterate over them.

jepio: How about i fetch the tigera operator manifest and from there figure out what calico version is getting deployed? Let me give it a try.

jepio:
@tormath1 @pothos I'd like to merge the updated version. It correctly determines the currently used calico version (v3.21.3) and will fix the failing Azure tests.

tormath1 (approved):
That looks good to me - we can merge this !

PR #1572, opened 2022-01-17
dev-libs/openssl: sync with the upstream

tormath1:
In this PR, we upgrade OpenSSL from 3.0.0 to 3.0.1. See: https://www.openssl.org/news/changelog.html#openssl-30

## How to use

```
emerge-amd64-usr -av dev-libs/openssl
```

## Testing done

* CI (:large_blue_circle:): http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4618/cldsv/

## Note for reviewers

* Nothing special, on the upstream there is no diff between the two ebuilds.
* After a first CI run, I notified that `::gentoo` has masked `openssl-3.0*` while fixing the dependencie's issues: https://bugs.gentoo.org/797325 since openssl and reverse dependencies are building fine for us, let's unmask it.


dev-libs/openssl/Manifest
```diff
@@ -4,4 +4,6 @@ DIST openssl-1.0.2u.tar.gz 5355412 BLAKE2B b2ff2a10e5851af5aca4093422a9a072c794e
...
 DIST openssl-1.1.1k.tar.gz 9823400 BLAKE2B e9bd90f17bc819c4960d07bbee04346e8a7adb87a764a09d033ef76f1d638c67b180c4f2beb84ec25fbff54ccc9c14c13b9b16a27cac231a5dd22b02635d5cec SHA512 73cd042d4056585e5a9dd7ab68e7c7310a3a4c783eafa07ab0b560e7462b924e4376436a6d38a155c687f6942a881cfc0c1b9394afcde1d8c46bf396e7d51121
 DIST openssl-1.1.1l.tar.gz 9834044 BLAKE2B 9e8739015db63a013c05587e3d164d67c3f65f1f6c5fc75e4592bcd038c036cde88a7bc95fbc1f1b4ed876f6124ca4dabcd4f5dbb45d1b84299f2efe1a59431a SHA512 d9611f393e37577cca05004531388d3e0ebbf714894cab9f95f4903909cd4f45c214faab664c0cbc3ad3cca309d500b9e6d0ecbf9a0a0588d1677dc6b047f9e0
+DIST openssl-1.1.1m.tar.gz 9847315 BLAKE2B 163262933df11afdb7b0c58fbbf0454b05e02951d28ed24e2c530affa18dee884d86555f7314506852ebfcc092bb509b6f9cd33893e30dab67bfb6f5713946eb SHA512 ba0ef99b321546c13385966e4a607734df38b96f6ed45c4c67063a5f8d1482986855279797a6920d9f86c2ec31ce3e104dcc62c37328caacdd78aec59aa66156
```
krnowak, line 7: Since this is overlay, could we drop all the unused ebuilds and tarballs, and keep only 3.0.1 and files that are used by that ebuild?

profiles/coreos/base/package.accept_keywords
```diff
@@ -12,8 +12,8 @@
 =dev-lang/rust-1.57.0-r1 ~amd64 ~arm64
 =virtual/rust-1.57.0 ~amd64 ~arm64
 
-# dev-libs/openssl-3.0.0 is still in testing phase at this point
-=dev-libs/openssl-3.0.0 ~amd64 ~arm64
+# dev-libs/openssl-3.0.1 is still in testing phase at this point
```
krnowak, line 15: I suppose that since openssl is in overlay, we could mark it as stable in the ebuild. What do you think?

profiles/coreos/base/package.unmask
```diff
@@ -12,3 +12,7 @@
 # libxcrypt -> glibc -> python), and also we need to update gcc to
 # version 10 or later.
 =virtual/libcrypt-1-r1
+
+# Overwrite portage-stable mask. OpenSSL-3* is building fine on Flatcar
+# and Flatcar's dependencies are building fine against.
```
krnowak, line 17:
```suggestion
# and Flatcar's dependencies are building fine against it.
```

profiles/coreos/base/package.accept_keywords
tormath1: Most of the time, I try to keep drift between `::gentoo` and `::coreos-overlay` as minimal as possible. Even to mark a package as stable.

Now you say it, if we update directly the ebuild, we could decrease the number of file to modified when we update a package (`package.accept_keywords` in this case) and so the mistake probability.

On the other hand, in the documentation[^1], we explicitely mention `package.accept_keywords` file update:
> To proceed, add the package name and version, and its masked architectures to the package.accept_keywords file inside the coreos profile. 


[^1]: https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-modifying-flatcar/#add-or-update-a-package

dev-libs/openssl/Manifest
tormath1: Done in a `fixup!` commit to the `Flatcar modifications` commit.

profiles/coreos/base/package.accept_keywords
krnowak: > Most of the time, I try to keep drift between `::gentoo` and `::coreos-overlay` as minimal as possible. Even to mark a package as stable.
> 
> Now you say it, if we update directly the ebuild, we could decrease the number of file to modified when we update a package (`package.accept_keywords` in this case) and so the mistake probability.
> 

Yeah. It's a tradeoff between minimizing the diff, but having changes in couple places vs having a bigger diff, but limited to a single location.

But yeah, I wasn't sure, so I asked for your opinion.


> On the other hand, in the documentation[1](#user-content-fn-1-c098c53cc832f1db184b6d2c38340798), we explicitely mention `package.accept_keywords` file update:
> 
> > To proceed, add the package name and version, and its masked architectures to the package.accept_keywords file inside the coreos profile.

I'd say it this is what we want to do for portage-stable packages (we shouldn't be modifying them, of course).

> 
> ## Footnotes
> 
>     1. https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-modifying-flatcar/#add-or-update-a-package [leftwards_arrow_with_hook](#user-content-fnref-1-c098c53cc832f1db184b6d2c38340798)

tormath1: Ok - then let's give a try by updating directly the ebuild to mark it as stable, it's consistent with the fact of cleaning the ebuild too.
`package.accept_keywords` should only be used for `::portage-stable` ebuilds.

tormath1:
@krnowak comments have been addressed, commits squashed and reworded.

krnowak (approved):

PR #1573, opened 2022-01-17
Changes related to app-arch updates

krnowak:
This is to remove EAPI 5 from portage-stable

This moves app-arch/bzip2 back to portage-stable. We used to have a patch for bzip2 for version 1.0.6, but we are actually using vanilla Gentoo version 1.0.8 without modifications for over a year (since Nov 2020).

Also, do some small keywords cleanup for arm64.

Needs to be merged together with https://github.com/flatcar-linux/portage-stable/pull/272.

CI: http://localhost:9091/job/os/job/manifest/4611/cldsv/


krnowak:
CI passed.

tormath1 (approved):

PR #1575, opened 2022-01-18
Update linux to 5.16.9

krnowak:
This updates both linux, and updates keywords for kernel-related packages. dev-util/perf has no update for 5.16 yet in Gentoo.

Should be merged together with https://github.com/flatcar-linux/portage-stable/pull/275.

CI: http://localhost:9091/job/os/job/manifest/4854/cldsv/


pothos (commented):
Two meta comments: for systemd please use at least two commits, one with the Gentoo upstream state, one with the changes we have on top.
The other comment is about the merge time. We have 5.15 which is an LTS kernel currently in the Alpha channel and only had one major release. It would be good to iterate on this major release first and see how it goes, maybe we can bring 5.15 to Stable first before we update the kernel in Alpha and effectively need to block the channel transitions at that point and also have to backport any wanted small updates manually to the Beta branches.

jepio:
I'm with @pothos on this, but I also wanted to ask: what is the motivation behind wanting to go to 5.16 so quickly?

krnowak:
> Two meta comments: for systemd please use at least two commits, one with the Gentoo upstream state, one with the changes we have on top.

Yeah, I was wondering about it - Gentoo still has 250.2, not 250.3, but that's a minor detail. I was mostly wondering how to treat the systemd package - as something that entirely our own (due to the size of modifications that we have made) or a fork from Gentoo that we occasionally sync from. I went with the former, but I can do the latter, no problem.

> The other comment is about the merge time. We have 5.15 which is an LTS kernel currently in the Alpha channel and only had one major release. It would be good to iterate on this major release first and see how it goes, maybe we can bring 5.15 to Stable first before we update the kernel in Alpha and effectively need to block the channel transitions at that point and also have to backport any wanted small updates manually to the Beta branches.

Sure, I'll split the PR then and keep linux update blocked for now.

krnowak:
> I'm with @pothos on this, but I also wanted to ask: what is the motivation behind wanting to go to 5.16 so quickly?

Certainly keeping the distro up to date and have some early testing of 5.16. But the channel transitions certainly throw a wrench into the works here. It can wait then.

sayanchowdhury:
I agree with @pothos. Ideally, we should have the two commits way in all the coreos-overlay packages that track upstream. In v249 update, I adjusted systemd to have two commits.

We can have Gentoo sync commit to have v250.2, and then v250.3 in the Flatcar patch.

krnowak:
systemd update is here: [portage-stable](https://github.com/flatcar-linux/portage-stable/pull/276) and [coreos-overlay](https://github.com/flatcar-linux/coreos-overlay/pull/1579).

krnowak:
This is outdated, and we won't be updating kernel to 5.16. Most likely the update will happen to version 6.0 or what will be the next LTS version, but that can happen in a new PR.

PR #1579, opened 2022-01-20
sys-apps/systemd: Update to 250.3

krnowak:
This synces the package with Gentoo and we apply our modifications. To make it a bit more consistent with Gentoo I replaced the ebuild setup that we had (9999 is a file, 249.7 is a symlink to 9999) with the one used in Gentoo (250.3 is a file, not a symlink). Also, I commented out the lines that we don't want instead of removing them, so it should be easier next time if something within the removed lines has changed that is relevant to us. I brought back the Gentoo-provided patches for systemd to minimize the diff in the ebuild. We won't apply them, because we set the `vanilla` USE flag in our profiles.

Should be merged together with PORTAGE_OVERLAY_PR

CI: http://localhost:9091/job/os/job/manifest/4642/cldsv/


krnowak:
CI passed.

sys-apps/systemd/files/gentoo-journald-audit.patch
```diff
@@ -0,0 +1,40 @@
...
+diff --git a/src/journal/journald-server.c b/src/journal/journald-server.c
+index 5865bf9809..163be685a8 100644
+--- a/src/journal/journald-server.c
++++ b/src/journal/journald-server.c
+@@ -2208,7 +2208,7 @@ int server_init(Server *s, const char *namespace) {
+                 .compress.threshold_bytes = (uint64_t) -1,
+                 .seal = true,
+ 
+-                .set_audit = true,
++                .set_audit = -1,
```
pothos, line 34: Did we rely on this?

pothos:
`/usr/bin/bootctl` disappeared from the image, not sure if this was an upstream systemd intention or if we need to set some flag now to install it.

jepio:
Looks like its part of systemd-boot (https://github.com/gentoo/gentoo/blob/master/sys-boot/systemd-boot/systemd-boot-249.9.ebuild). But bootctl is not relevant to Flatcar so far.

pothos:
> But bootctl is not relevant to Flatcar so far.

It has some generic features like `bootctl status`, `bootctl reboot-to-firmware` or `bootctl random-seed`. We should not add and delete binaries randomly, and if we do, add a changelog entry for users to be aware that something got dropped.

pothos:
Since we don't ship `efibootmgr` I would say that `bootctl` has a value on Flatcar and if we can keep it, we should.

sys-apps/systemd/files/gentoo-journald-audit.patch
pothos: Seem we are fine, I still see audit logs in the kernel.

krnowak: This is a part of a Gentoo-specific patch, that gets applied if the `vanilla` USE flag is disabled. We enable the flag in `profiles/features/systemd/package.use.force`, so we are not getting this change.

The sole reason for having these patches is to minimize the diff between our ebuild and Gentoo's.

jepio:
OK, I see the value in bootctl (and efibootmgr), we can add the systemd-boot ebuild and play with `INSTALL_MASK` to restrict the installed files.

krnowak:
Before the update, `bootctl` was built if we requested general support for efi, now it needs also requesting support for gnu-efi in the build system. This can be done by enabling the `gnuefi` use flag for systemd. Will rerun the CI.

krnowak:
Changelog updated, but the CVE has no entry in NVD, yet. The [entry on mitre](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-3997) provides no useful information.

pothos:
The NVD links will work later, it's ok to use them for consistency, one could also add a second link to currently available info

pothos:
Looks good now, the remaining question I have is about the content of this change
```
-/usr/lib64/sysusers.d/basic.conf
-/usr/lib64/sysusers.d/README
-/usr/lib64/sysusers.d/systemd.conf
-/usr/lib64/sysusers.d/systemd-remote.conf
+/usr/lib64/sysusers.d/acct-*.conf
```
and what part in Flatcar sets the user accounts up (do then end up in the `/usr`'s `passwd` or in `/etc/passwd`)? Need to check this.

pothos:
Not ideal for Flatcar what the current result is with the Gentoo acct way of things (`acct-user-systemd-oom.conf` and friends now added to the image):
`/etc/passwd` got:
```
systemd-coredump:x:194:194:System user; systemd-coredump:/dev/null:/sbin/nologin
systemd-oom:x:198:198:System user; systemd-oom:/dev/null:/sbin/nologin
systemd-timesync:x:195:195:System user; systemd-timesync:/dev/null:/sbin/nologin
```
`/usr/share/baselayout/passwd` has (from https://github.com/flatcar-linux/baselayout/blob/flatcar-master/baselayout/passwd):
```
[…]
systemd-journal-remote:x:242:242:Journal Remote:/dev/null:/sbin/nologin
systemd-network:x:244:244:Network Management Service:/dev/null:/sbin/nologin
systemd-resolve:x:245:245:Network Name Resolution:/dev/null:/sbin/nologin
systemd-bus-proxy:x:246:246:Legacy D-Bus Proxy:/dev/null:/sbin/nologin
```
Normally the `/etc/passwd` file is left completely to the user, that's why the nss-altfiles glibc plugin is used to inject things from `/usr/share/baselayout/passwd` at runtime. I don't know how systemd-sysusers handles conflicting existing entries in `/etc/passwd` if the IDs change over the releases.

pothos:
I suggest to add the users to the baselayout file to prevent creation through `systemd-sysusers.service` at image boot. We can do this manually as done before or move to some dynamic way where after the package installation step in the image building we copy the resulting passwd file to the `/usr/share/baselayout/passwd` location (idea is that we don't have to maintain this but get it for free from the sysusers way in gentoo).

krnowak:
Yeah, will do that. The conf files from acct-user and acct-group are going to be still there, in `/usr/lib/sysusers.d`, but I'll verify if they have matching UIDs/GIDs and that they are no-ops.

pothos:
Please create and merge the baselayout PR and update the commit ID to the merge commit

krnowak:
Updated, had to add more overrides (for messagebus, ntp, sshd, polkitd and tss).

pothos:
Thanks!

PR #1588, opened 2022-01-27
sys-auth/polkit: sync with `::gentoo`

tormath1:
In this PR, we upgrade `sys-auth/polkit` to `0.120-r2` including patch for `cve-2021-4034`.

## Testing done

CI (:large_blue_circle:): http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4701/cldsv/

## Note for reviewers

* Duktape patch has been moved from [MR#35](https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/35) to [MR#97](https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/97)
* Patch has been reworked to remove this commit: https://gitlab.freedesktop.org/polkit/polkit/-/commit/c7fc4e1b61f0fd82fc697c19c604af7e9fb291a2

Related-To: https://github.com/flatcar-linux/Flatcar/issues/606


sys-auth/polkit/polkit-0.120-r2.ebuild
```diff
@@ -0,0 +1,131 @@
...
+QA_MULTILIB_PATHS="
+	usr/lib/polkit-1/polkit-agent-helper-1
+	usr/lib/polkit-1/polkitd"
+
+src_prepare() {
+	local PATCHES=(
+		"${FILESDIR}/polkit-0.120-meson.patch"
+		"${FILESDIR}/polkit-0.120-CVE-2021-4043.patch"
+
+		# from https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/35
```
pothos, line 69: https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/97 is the one that got merged now, should we switch to it? I didn't compare both but in question it seems to be better to take the merged one

tormath1: Nice, I missed that :) Will update.

tormath1: Patch from the MR#97 is now applied.
Reworked a bit the patch by removing the `.gitlab-ci.yml` section since it's not present in the archive, commit has been amended consequently and a new CI has been triggered.

pothos: Yeah, good, it should be the last update where we drag the downstream patch anyway

krnowak (commented):
Could you keep only the ebuild we actually modify and drop the rest? It is usually confusing to see several ebuilds in a directory and only one of them is modified for our needs.

tormath1:
@krnowak you're right. It's done in the `fixup!` commit. I applied the following patch in the meantime: https://github.com/gentoo/gentoo/pull/23980

pothos:
> Could you keep only the ebuild we actually modify and drop the rest? It is usually confusing to see several ebuilds in a directory and only one of them is modified for our needs.

Thanks, another nice side effect is an easier to understand diff in the overall PR

pothos (approved):
Assuming the CI passes without related failures

krnowak (approved):

PR #1602, opened 2022-01-31
sys-kernel: allow fips mode to be enabled

pothos:
With this kernel config, users can boot with fips=1 set in
`/usr/share/oem/grub.cfg`:
```
set linux_append="fips=1"
```

Which triggers various behaviors, for FIPS 200 certification.

with this config compiled in, and that boot parameter, users can can
that fips is enabled with:
```
flatcar ~ # cat /proc/sys/crypto/fips_enabled
1
```

## Testing done

scheduled


krnowak (approved):
Looks good. But I guess that the non-interactive first boot is going to be with fips mode disabled, even if we patch up grub.cfg with ignition, as the patched grub.cfg would take effect only after a reboot.

PR #1622, opened 2022-02-07
sys-apps/systemd: add downstream patch to disable foreign route mgmt

pothos:
This new downstream patch disables the ManageForeignRoutes and
ManageForeignRoutingPolicyRules systemd-netword settings by default to
ensure that CNIs don't get their routes or routing policy rules
discarded on network reconfiguration events.

https://github.com/flatcar-linux/Flatcar/issues/620

## How to use

Also port to Stable, maybe while also bumping to the latest systemd point release.

Needs coordination with the systemd 250 update PR to ensure this doesn't get lost.

## Testing done

Manually tested turning one of the bonded interfaces down and it worked well

### DIFF
--- a/sys-apps/systemd/files/0001-networkd-disable-managing-of-foreign-routes-rules-by-default.patch
+++ b/sys-apps/systemd/files/0001-networkd-disable-managing-of-foreign-routes-rules-by-default.patch
@@ -1,0 +1,63 @@
+From 513429b47f0852d17ba721ad5d55baa985f48ddb Mon Sep 17 00:00:00 2001
+From: Kai Lueke <kailuke@microsoft.com>
+Date: Mon, 7 Feb 2022 17:39:23 +0100
+Subject: [PATCH] networkd: disable managing of foreign routes/rules by default
+
+While systemd-networkd follows the principle of a declarative network
+configuration and thus needs a way to ensure that unwanted routes or
+routing policy rules get discarded, the interfacing with procedural
+network management from CNIs like Cilium is limited, so that when the
+interface is set to "unmanaged" through a networkd unit, any routing
+policies there would also be ignored and discarded unless they would
+be defined for a new unit for a dummy network interface. This means
+the only option left is to disable the discarding of foreign rules
+globally.
+
+Change the default for ManageForeignRoutes and
+ManageForeignRoutingPolicyRules to "no" to ensure that we don't
+interfere with the network management of the CNIs. Users that rely on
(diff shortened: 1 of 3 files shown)


sys-apps/systemd/files/0001-networkd-disable-managing-of-foreign-routes-rules-by-default.patch
```diff
@@ -0,0 +1,63 @@
...
+--- a/src/network/networkd.conf
++++ b/src/network/networkd.conf
+@@ -15,8 +15,8 @@
+ [Network]
+ #SpeedMeter=no
+ #SpeedMeterIntervalSec=10sec
+-#ManageForeignRoutingPolicyRules=yes
+-#ManageForeignRoutes=yes
++#ManageForeignRoutingPolicyRules=no
++#ManageForeignRoutes=no
```
ymmt2005, line 57: Why don't we just disable these without patching `networkd-manager.c`?

```suggestion
-#ManageForeignRoutingPolicyRules=yes
-#ManageForeignRoutes=yes
+ManageForeignRoutingPolicyRules=no
+ManageForeignRoutes=no
```

pothos: This config file gets placed as initial version to `/etc` and from there on the user owns it and will not get a new version of it when updating.

pothos: I've searched for a better solution than patching the C file but there are no meson build switches for these two settings.

ymmt2005: I see.  Thank you.

jepio (approved):
I suggest bumping the revision of the ebuild symlink too.

pothos:
> I suggest bumping the revision of the ebuild symlink too.

For the sake of easy cherry picking without conflicts I'm a bit reluctant to do this

jepio:
OK

sys-apps/systemd/files/0001-networkd-disable-managing-of-foreign-routes-rules-by-default.patch
pothos: Actually there is a way to use drop-in files for the networkd.conf file, I'll submit a PR.

PR #1628, opened 2022-02-09
coreos-base/oem-ec2-compat: set correct console on arm64

jepio:
# coreos-base/oem-ec2-compat: set correct console on arm64

The default console (console=ttyAMA0) that is set by grub is wrong for EC2 arm64, set it to console=ttyS0 manually to allow capturing console output. With the wrong setting the kola console log is empty.

## How to use

Start an arm64 EC2 instance.

## Testing done

CI passed and more data is found in console.txt (http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4797/cldsv/)


jepio:
EC2 buffers the last 64kB, we fetch the console twice  before the test ends. It would need an asynchronous console-fetcher in kola to overcome this limitation, so yeah, out of scope.

jepio:
New issue here https://github.com/flatcar-linux/Flatcar/issues/625

pothos:
Picked for flatcar-3033

PR #1636, opened 2022-02-10
Add auditd package and systemd unit

JAORMX:
# Add auditd package and systemd unit

This includes the `auditd` binary and systemd unit as part of the
distro. While journald is also able to handle logs from the linux audit
subsystem, auditd provides audit-specific capabilities that are
necessary in deployments subject to regulatory compliance.

For one, an administrator is able to configure audit log writing policy
to ensure that logs land on disk and nothing is missed (`flush`). We
wouldn't want such policy through journald as it woudl sync and ensure
all logs which might be undesirable and too resource intensive. In
short, this allows us to configure different management policies for
audit logs compared to general logs.

It allows us to explicitly configure the node's reaction to errors such
as the disk beign full, the disk having other issues or space constraints.

While Flatcar is not Common Criteria certified which would require the
system to shut down if audit logs present issues (not written or
collected), some FedRAMP environments do require actions such as
notifications (which could be achieved via syslog). This can be
explicitly done with auditd as well.

## How to use

## Testing done

### DIFF
--- a/changelog/changes/2022-02-15-auditd.md
+++ b/changelog/changes/2022-02-15-auditd.md
@@ -1,0 +1,2 @@
+- Added `auditd.service` but left it disabled by default, a custom configuration can be created by removing the `/etc/audit/auditd.conf` and replacing it with an own file ([PR#1636](https://github.com/flatcar-linux/coreos-overlay/pull/1636))
+
--- a/changelog/changes/2022-02-15-auditd.md
+++ b/changelog/changes/2022-02-15-auditd.md
@@ -1,2 +1,2 @@
-- Added `auditd.service` but left it disabled by default, a custom configuration can be created by removing the `/etc/audit/auditd.conf` and replacing it with an own file ([PR#1636](https://github.com/flatcar-linux/coreos-overlay/pull/1636))
+- Added `auditd.service` but left it disabled by default, a custom configuration can be created by removing `/etc/audit/auditd.conf` and replacing it with an own file ([PR#1636](https://github.com/flatcar-linux/coreos-overlay/pull/1636))
 
--- a/sys-process/audit/audit-3.0.6.ebuild
+++ b/sys-process/audit/audit-3.0.6.ebuild
@@ -42,17 +42,13 @@ src_prepare() {
 	# Disable installing sample rules so they can be installed as docs.
 	echo -e '%:\n\t:' | tee rules/Makefile.{am,in} >/dev/null
 
-	# Flatcar: Do not build daemon stuff.
-	sed -e '/^SUBDIRS =/s/audisp//' \
-		-i Makefile.am || die
 	# Flatcar: Some legacy stuff is being installed when systemd
 	# is enabled. Drop all the lines that try doing it.
 	sed -e '/${DESTDIR}${initdir}/d' \
 		-e '/${DESTDIR}${legacydir}/d' \
 		-i init.d/Makefile.am || die
 	# Flatcar: Do not build daemon stuff.
-	sed -e '/^sbin_PROGRAMS =/s/auditd//' \
-		-e '/^sbin_PROGRAMS =/s/aureport//' \
+	sed -e '/^sbin_PROGRAMS =/s/aureport//' \
 		-e '/^sbin_PROGRAMS =/s/ausearch//' \
 		-i src/Makefile.am || die
 
@@ -144,6 +140,11 @@ multilib_src_install_all() {
 	# newinitd "${FILESDIR}"/auditd-init.d-2.4.3 auditd
 	# newconfd "${FILESDIR}"/auditd-conf.d-2.1.3 auditd
 
+	# Flatcar: install sample configuration
+	insinto /usr/share/auditd
+	doins "${S}"/init.d/auditd.conf
+
+
 	# Flatcar: We are not installing audisp too.
 	# [ -f "${ED}"/sbin/audisp-remote ] && \
 	# dodir /usr/sbin && \
@@ -151,7 +152,6 @@ multilib_src_install_all() {
 
 	# Flatcar: Do not install gentoo rules.
 	# Gentoo rules
-	# insinto /etc/audit
 	# newins "${FILESDIR}"/audit.rules-2.1.3 audit.rules
 	# Flatcar: We are installing our own rules.
 	insinto /usr/share/audit/rules.d
@@ -160,14 +160,15 @@ multilib_src_install_all() {
 	# doins "${FILESDIR}"/audit.rules.stop*
 
 	# audit logs go here
-	# Flatcar: This is where auditd puts its logs. We don't have
-	# the daemon, so get rid of the unnecessary directory.
-	# keepdir /var/log/audit
+	keepdir /var/log/audit
 
 	find "${ED}" -type f -name '*.la' -delete || die
 
 	# Security
 	lockdown_perms "${ED}"
+
+	# Flatcar: We add the systemd unit but don't enable it.
+	systemd_dounit init.d/auditd.service
 
 	# Flatcar: Our systemd stuff.
 	newtmpfiles "${FILESDIR}"/audit-rules.tmpfiles audit-rules.conf
@@ -182,11 +183,13 @@ pkg_postinst() {
 lockdown_perms() {
 	# Upstream wants these to have restrictive perms.
 	# Should not || die as not all paths may exist.
-	# Flatcar: No lockdown of permissions - it's probably only
-	# related to auditd.
-	# local basedir="${1}"
+	# Flatcar: We don't include ausearch and aureport
+	# so they're removed from the hardening list
+	local basedir="${1}"
 	# chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,report,search,trace} 2>/dev/null
-	# chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
+	chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,trace} 2>/dev/null
+	chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
 	# chmod 0640 "${basedir}"/etc/audit/{auditd.conf,audit*.rules*} 2>/dev/null
+	rm -f  "${basedir}"/etc/audit/auditd.conf 2>/dev/null
 	:
 }
--- a/sys-process/audit/files/audit-rules.tmpfiles
+++ b/sys-process/audit/files/audit-rules.tmpfiles
@@ -1,5 +1,6 @@
-d   /etc/audit                          -   -   -   -   -
-d   /etc/audit/rules.d                  -   -   -   -   -
-L   /etc/audit/rules.d/00-clear.rules   -   -   -   -   /usr/share/audit/rules.d/00-clear.rules
-L   /etc/audit/rules.d/80-selinux.rules -   -   -   -   /usr/share/audit/rules.d/80-selinux.rules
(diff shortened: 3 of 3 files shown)


sys-process/audit/audit-3.0.6.ebuild
```diff
@@ -182,11 +178,12 @@ pkg_postinst() {
...
-	# local basedir="${1}"
+	# Flatcar: We don't include ausearch and aureport
+	# so they're removed from the hardening list
+	local basedir="${1}"
 	# chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,report,search,trace} 2>/dev/null
-	# chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
-	# chmod 0640 "${basedir}"/etc/audit/{auditd.conf,audit*.rules*} 2>/dev/null
+	chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,trace} 2>/dev/null
+	chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
+	chmod 0640 "${basedir}"/etc/audit/{auditd.conf,audit*.rules*} 2>/dev/null
```
pothos, line 187: Thanks for the changes. I think the location of the config and the default permissions will be a bit tricky. Further above the location used is ` /usr/share/audit/rules.d` which is due to the auto update only covering the `/usr` partition and not `/etc` or `/var`. If we need files under `/etc` or `/var` we normally set them up through systemd-tmpfile directives to ensure that they exist on updated nodes, too. One possible solution for the config file would be to place the config file `/etc/audit/auditd.conf` under `/usr` and  set up symlinks from `/etc/` to it which is currently done for the rules under `/etc/audit/rules.d/` already (see `files/audit-rules.tmpfiles`).

pothos: (The user would still be able to customize the config by placing a regular file there and systemd-tmpfiles would not touch it to create a symlink)

pothos: E.g., for the log directory it would be an entry like `d   /var/log/audit        0750    -   -   -   -`

pothos: If the `auditd.conf` file actually just holds the default settings we can even try to delete it

sys-process/audit/audit-3.0.6.ebuild
```diff
@@ -182,10 +178,11 @@ pkg_postinst() {
...
 	# Should not || die as not all paths may exist.
-	# Flatcar: No lockdown of permissions - it's probably only
-	# related to auditd.
-	# local basedir="${1}"
+	# Flatcar: We don't include ausearch and aureport
+	# so they're removed from the hardening list
+	local basedir="${1}"
 	# chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,report,search,trace} 2>/dev/null
+	chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,trace} 2>/dev/null
 	# chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
```
pothos, line 186:
```suggestion
	chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
```
Let's enable this here and check the result of the test build

sys-process/audit/audit-3.0.6.ebuild
```diff
@@ -144,14 +140,18 @@ multilib_src_install_all() {
 	# newinitd "${FILESDIR}"/auditd-init.d-2.4.3 auditd
 	# newconfd "${FILESDIR}"/auditd-conf.d-2.1.3 auditd
 
+	# Flatcar: install sample configuration
+	insinto /usr/share/auditd
+	doins "${FILESDIR}"/init.d/auditd.conf
+	newtmpfiles auditd.conf
```
pothos, line 146: This here is about the tmpfile .conf, so
```suggestion
	newtmpfiles "${FILESDIR}/auditd-conf.tmpfiles" auditd-conf.conf
```
and a new file `auditd-conf.tmpfiles` would contain `L   /etc/audit/auditd.conf   -       -       -       -       /usr/share/auditd/auditd.conf`

sys-process/audit/audit-3.0.6.ebuild
```diff
@@ -182,11 +183,12 @@ pkg_postinst() {
...
-	# local basedir="${1}"
+	# Flatcar: We don't include ausearch and aureport
+	# so they're removed from the hardening list
+	local basedir="${1}"
 	# chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,report,search,trace} 2>/dev/null
-	# chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
-	# chmod 0640 "${basedir}"/etc/audit/{auditd.conf,audit*.rules*} 2>/dev/null
+	chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,trace} 2>/dev/null
+	chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
+	chmod 0640 "${basedir}"/etc/audit/{auditd.conf,audit*.rules*} 2>/dev/null
```
pothos, line 192: Instead of `chmod … /etc/auditd/auditd*rules` we need to change `sys-process/audit/files/audit-rules.tmpfiles` to have `640` where the first `-` column is.

JAORMX: Done.

sys-process/audit/audit-3.0.6.ebuild
```diff
@@ -144,14 +140,19 @@ multilib_src_install_all() {
 	# newinitd "${FILESDIR}"/auditd-init.d-2.4.3 auditd
 	# newconfd "${FILESDIR}"/auditd-conf.d-2.1.3 auditd
 
+	# Flatcar: install sample configuration
+	insinto /usr/share/auditd
+	doins "${FILESDIR}"/init.d/auditd.conf
+	newtmpfiles "${FILESDIR}/auditd-conf.tmpfiles" auditd-conf.conf
```
pothos, line 146:
```suggestion

```
Since you added it to the rules tmpfile this here is redundant (and doesn't work because the file does not exist)

sys-process/audit/audit-3.0.6.ebuild
pothos: I tried to use it without the file and it somehow worked, just an error message in the log but it seemed to do what it should, no idea.
Anyway, I've now started a test build of the current state which I think is very good. When that's done we can add a changelog entry and that's it.

sys-process/audit/audit-3.0.6.ebuild
```diff
@@ -144,14 +140,18 @@ multilib_src_install_all() {
 	# newinitd "${FILESDIR}"/auditd-init.d-2.4.3 auditd
 	# newconfd "${FILESDIR}"/auditd-conf.d-2.1.3 auditd
 
+	# Flatcar: install sample configuration
+	insinto /usr/share/auditd
+	doins "${FILESDIR}"/init.d/auditd.conf
```
pothos, line 145: This path here is not yet correct: `FileNotFoundError: [Errno 2] No such file or directory: b'/build/amd64-usr/var/tmp/portage/sys-process/audit-3.0.6/files/init.d/auditd.conf'`
I think we need `WORKDIR` or `S` instead of `FILESDIR`, I'll have to give it a try

JAORMX: Let me know if using `WORKDIR` works for you

pothos:
```suggestion
	doins "${S}"/init.d/auditd.conf
```
This here works

pothos:
The changelog entry needs a file like `changelog/changes/2022-02-15-auditd.md` with a single markdown list entry like that:
```
- Added `auditd.service` but left it disabled by default, a custom configuration can be created by removing the `/etc/audit/auditd.conf` and replacing it with an own file ([PR#1636](https://github.com/flatcar-linux/coreos-overlay/pull/1636))
```

JAORMX:
> The changelog entry needs a file like `changelog/changes/2022-02-15-auditd.md` with a single markdown list entry like that:
> 
> ```
> - Added `auditd.service` but left it disabled by default, a custom configuration can be created by removing the `/etc/audit/auditd.conf` and replacing it with an own file ([PR#1636](https://github.com/flatcar-linux/coreos-overlay/pull/1636))
> ```

Done. Thanks @pothos for all your help and patience!

changelog/changes/2022-02-15-auditd.md
```diff
@@ -0,0 +1,2 @@
+- Added `auditd.service` but left it disabled by default, a custom configuration can be created by removing the `/etc/audit/auditd.conf` and replacing it with an own file ([PR#1636](https://github.com/flatcar-linux/coreos-overlay/pull/1636))
```
pothos, line 1:
```suggestion
- Added `auditd.service` but left it disabled by default, a custom configuration can be created by removing the `/etc/audit/auditd.conf` symlink and replacing it with an own file ([PR#1636](https://github.com/flatcar-linux/coreos-overlay/pull/1636))
```

pothos:
It's a bit rough because of the downstream changes… I hope we can create an automatism soon that allows us to use unmodified Gentoo ebuild files.

pothos:
Maybe the `/etc/audit/auditd.conf` file still gets created from the package installation. In this case we can add something like this:
```
@@ -189,6 +189,7 @@ lockdown_perms() {
 	# chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,report,search,trace} 2>/dev/null
 	chmod 0750 "${basedir}"/sbin/au{ditctl,ditd,trace} 2>/dev/null
 	chmod 0750 "${basedir}"/var/log/audit 2>/dev/null
 	# chmod 0640 "${basedir}"/etc/audit/{auditd.conf,audit*.rules*} 2>/dev/null
+	rm -f  "${basedir}"/etc/audit/auditd.conf 2>/dev/null
 	:
 }
```

pothos:
Thanks a lot, can you squash the commits?

pothos:
Oh no… Now that I tried it again I get `Error opening config file (Too many levels of symbolic links)` :(
Seems we can't use the symbolic link approach to keep the file updated. Either we can try to use the copy tmpfile directive (`C     /target/to/create                        -    -    -     -           /source/to/copy` see https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html), or we do a compromise then and ship the file only for new installations.

sys-process/audit/files/audit-rules.tmpfiles
```diff
@@ -1,5 +1,6 @@
-d   /etc/audit                          -   -   -   -   -
-d   /etc/audit/rules.d                  -   -   -   -   -
-L   /etc/audit/rules.d/00-clear.rules   -   -   -   -   /usr/share/audit/rules.d/00-clear.rules
-L   /etc/audit/rules.d/80-selinux.rules -   -   -   -   /usr/share/audit/rules.d/80-selinux.rules
-L   /etc/audit/rules.d/99-default.rules -   -   -   -   /usr/share/audit/rules.d/99-default.rules
+d   /etc/audit                          750   -   -   -   -
+L   /etc/audit/auditd.conf              640  -   -   -   /usr/share/auditd/auditd.conf
```
pothos, line 2:
```suggestion
C   /etc/audit/auditd.conf              640  -   -   -   /usr/share/auditd/auditd.conf
```
This will create the file if it doesn't exist. Unfortunately this means we won't update it but hopefully we soon find a way to do so.

pothos (approved):
Waiting to start it once on the built image, then we are ready to merge

PR #1638, opened 2022-02-10
coreos-base/update_engine: fix flatcar-postinst migration notice

pothos:
This pulls in
https://github.com/flatcar-linux/update_engine/pull/16
to fix a small template error in the cgroup v2 migration notice.

## How to use

Decide if we want to add this to Stable 3033 which would also pull in the creation of the kured `/run/reboot-required` file

## Testing done

None


jepio (approved):
I almost feel as though this isn't worthy of a changelog entry - its a typo in a comment that few people will read. But you can leave it.

pothos:
Yeah, I'm fine dropping it from the release notes to make room for other stuff

PR #1649, opened 2022-02-16
profiles/coreos: move gnuefi systemd USE flag to target

jepio:
# profiles/coreos: move gnuefi systemd USE flag to target

SDK bootstrap is failing with:

  Message: sbat-distro (from ID):

  ../systemd-stable-250.3/src/boot/efi/meson.build:189:24: ERROR: Problem encountered: Required sbat-distro option not set and autodetection failed

The gnuefi USE flag controls whether bootctl and systemd-boot are built, but we
only need those on the target. Currently the USE flag is set for SDK as well,
so move it to coreos/targets/generic.

## How to use

Rebootstrap.

## Testing done

CI started: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4861/

### DIFF
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -28,10 +28,9 @@ net-analyzer/nmap ncat -system-lua
 # removes mta dependencies
 app-admin/sudo -sendmail
 
-# use lzma which is the default on non-gentoo systems, use gnuefi for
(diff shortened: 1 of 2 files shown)


krnowak (approved):
Oops, looks good, thanks!

jepio:
os/sdk build succeeded and `USE=gnuefi` is still present in target packages. Merging.

PR #1651, opened 2022-02-16
Replace systemd patch by a drop-in file

pothos:
This pulls in
https://github.com/flatcar-linux/init/pull/61
and
https://github.com/flatcar-linux/baselayout/pull/22
to use a drop-in file instead of the systemd patch.

## How to use

No need to port this to other channels

## Testing done

[Done here](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4866/) and by setting up Cilium on a QEMU VM and doing `networkctl down eth0 ; networkctl up eth0` without impact (while on Stable 3033.2.2 it results in loosing the routing policy rules).


krnowak (approved):
Looks good.

PR #1658, opened 2022-02-17
profiles,coreos-base: changes needed by bpftool

dongsupark:
Get `dev-util/bpftool` included in the production images by default.

As `dev-util/bpftool` does not have a keyword `arm64`, we need to force enable arm64 keyword for bpftool.

Add `dev-python/docutils`, `dev-util/patchutils` to hard-host-depends.
Without adding those in the SDK, the new package `dev-util/bpftool` would end up pulling in the new dependencies into the production images, which should not happen.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/291.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4881/cldsv

- Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) - done in https://github.com/flatcar-linux/portage-stable/pull/291


krnowak:
We probably could file an issue on Gentoo to enable and stabilize bpftool on arm64 when it reaches stable.

PR #1659, opened 2022-02-17
dev-db/sqlite: skip `sqlite3` installation

tormath1:
In this PR, we add `/usr/bin/sqlite3` to `INSTALL_MASK` in order to skip `sqlite3` cli installation.

## Testing done

- CI (:large_blue_circle:): http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4888/cldsv/

## Note for reviewers

- Not sure we need a changelog entry since sqlite3 has not been introduced yet in an alpha release.

### DIFF
--- a/profiles/coreos/targets/generic/make.defaults
+++ b/profiles/coreos/targets/generic/make.defaults
@@ -12,6 +12,7 @@ FEATURES="nodoc noinfo noman"
 # Exclude even more documentation
 # Remove bash-completion files as we don't install bash-completion.
 # Remove locale LC_MESSAGES files.
+# Remove sqlite3 binary installation provided by dev-db/sqlite as we only need the libs.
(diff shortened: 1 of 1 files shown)


jepio:
As @pothos said, I would add it here: https://github.com/flatcar-linux/coreos-overlay/blob/main/profiles/coreos/targets/generic/make.defaults, there's no need to fork an ebuild to remove a single file. What do you think?

dongsupark:
In the particular case of this PR, I am fine with either way, INSTALL_MASK or not.

In case of Flatcar's coreos-overlay, however, `INSTALL_MASK` settings in various `make.defaults` are so ugly.
A typical example of growing code base over many years. I feel even scary about touching those files. :-(
If someone has free cycles, please clean up unnecessary parts in the masks.

pothos (approved):
Thanks

PR #1660, opened 2022-02-18
sys-apps/shadow: update to 4.11.1

dongsupark:
Sync with Gentoo to update `sys-apps/shadow` to 4.11.1, mainly to address [CVE-2013-4235](https://nvd.nist.gov/vuln/detail/CVE-2013-4235).

Based on Gentoo commit [defe2a377e43](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=defe2a377e43a756441b183b66e2c4aae2be27b5).

Flatcar changes are based on commit [f118a42e854f](https://github.com/flatcar-linux/coreos-overlay/commit/f118a42e854f9596bff629c864216aa4e9b4b1c2).

We also need to enable `su` USE flag for sys-apps/shadow. Ideally we should enable `su` for util-linux, and disable the flag in shadow. However, to avoid pulling in unnecessary dependencies into the Flatcar SDK, for now we enable `su` for shadow and disable `su` for util-linux.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/294.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4922/cldsv


sys-apps/shadow/shadow-4.11.1.ebuild
```diff
@@ -1,19 +1,20 @@
...
+SRC_URI="https://github.com/shadow-maint/shadow/releases/download/v${PV}/${P}.tar.xz"
 
 LICENSE="BSD GPL-2"
-SLOT="0"
-KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv s390 sparc x86"
-IUSE="acl audit bcrypt cracklib nls pam selinux skey split-usr +su xattr"
+# Subslot is for libsubid's SONAME.
+SLOT="0/4"
+KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv ~s390 sparc x86"
+IUSE="acl audit bcrypt cracklib nls pam selinux skey split-usr su xattr"
```
thesamesam, line 17: Just a heads up:
1. They broke ABI (so remember to add := to consumers and revbump them, if they use libsubuid -- not many of these exist. Most things just use shadow binaries, so not need this).
2. We changed default from +su -> su and now use util-linux to provide su. You may want to re-enable su for shadow and disable it for util-linux instead. This is because shadow was deprecating it entirely but now will be keeping it for non-PAM users.

Hope those hints might be useful somehow.

dongsupark: Thanks for the heads-up.

> They broke ABI (so remember to add := to consumers and revbump them, if they use libsubuid -- not many of these exist. Most things just use shadow binaries, so not need this). IIRC this is just podman and maybe one or two other container-related packages.

Currently no binary in the Flatcar production images is linking against libsubid.
(If we could somehow add podman, then it might be a different story, but not in the short future.)
For now I think there is no need to worry.

> We changed default from +su -> su and now use util-linux to provide su. You may want to re-enable su for shadow and disable it for util-linux instead if you want to avoid churn. This is because shadow was deprecating it entirely but now will be keeping it for non-PAM users. But my recommendation is to use util-linux su as that is the preference of shadow upstream too.

That's an interesting point.
I agree that Flatcar should also use su in util-linux instead of shadow.

However, it looks like, changes like that seem to open up a can of worms.
First, under the `/build/amd64-usr`, it looks ok to rely on su from util-linux.
OTOH, in the SDK, installation of `util-linux[pam,su]` requires `pam_sss.so` to be installed in the SDK, because baselayout of Flatcar requires that.
I am not sure it is a good idea to add the whole sssd into the SDK, only for the pam_sss module. :thinking:

dongsupark:
CI passed

krnowak (approved):
Looks alright, but I guess we will have to have a look at our modifications at some point:

- split-usr could be just disabled in a profile instead of dropping the code from ebuild
- arch-specific cruft could be reduced (we don't care about ppc, hppa or sh).

But I think I'm going to do it as a part of the rework of /etc handling, so let's get this in.

PR #1661, opened 2022-02-18
sys-kernel/bootengine: fix containerised builds

t-lo:
Bump `CROS_WORKON` to include https://github.com/flatcar-linux/bootengine/pull/36 to fix an issue with dracut in containerised builds.

This fix should be cherry-picked to all active maintenance branches, i.e. `flatcar-3033`, `flatcar-3066`, and `flatcar-3139`.

### DIFF
--- a/sys-kernel/bootengine/bootengine-9999.ebuild
+++ b/sys-kernel/bootengine/bootengine-9999.ebuild
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="https://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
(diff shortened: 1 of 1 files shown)


pothos:
Another alternative is to add an ebuild patch file for Stable

t-lo:
> Another alternative is to add an ebuild patch file for Stable

Uh that's quite a lot actually; there are wipefs patches that go back 'til September last year. I'll just port this patch then and create separate PRs for 3033, 3066, and 3139.

t-lo:
@pothos https://github.com/flatcar-linux/bootengine/pull/37 and https://github.com/flatcar-linux/bootengine/pull/38

t-lo:
Cherry-picked for flatcar-3139.

PR #1675, opened 2022-02-25
build fixes (shim/polkit)

jepio:
# build fixes (shim/polkit)

* mark sys-boot/shim as amd64 only dependency (the package doesn't build for arm64)
* polkit dependency on duktape needs to be stricter when using binpkgs

## How to use

## Testing done

CI running.


tormath1 (approved):
good catch for the polkit rebuild ! Let's merge this as soon as the CI is :large_blue_circle:

PR #1676, opened 2022-02-25
core-base/oem-azure(-pro): enable earlycon on all platforms

jepio:
# core-base/oem-azure(-pro): enable earlycon on all platforms

Both architectures and VM generations.

## How to use

## Testing done

CI running


jepio:
Checked the CI logs, and all configurations correctly enable early console now.

krnowak (approved):

PR #1677, opened 2022-02-25
coreos-devel/mantle: update to latest state

pothos:
This updates the internal kola version that is now used for tests from
the SDK container pipeline.

## How to use

Pick for all channels

Maybe we should set up a GitHub Action for that or do this as part of the nightly submodule update

### DIFF
--- a/coreos-devel/mantle/mantle-9999.ebuild
+++ b/coreos-devel/mantle/mantle-9999.ebuild
@@ -11,7 +11,7 @@ COREOS_GO_MOD="vendor"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm64"
 else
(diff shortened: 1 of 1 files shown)


tormath1:
Maybe we can cut a release of mantle instead of pulling flatcar-master ?

pothos:
This is going to be updated daily from now on, or maybe even for every commit

tormath1 (approved):

dongsupark:
This PR uncovered a hidden build issue in flatcar-3033 and flatcar-3066.
To fix that issue, cherry-picked https://github.com/flatcar-linux/coreos-overlay/commit/5f7a593137c136a692b302445f7c7f44ecbe5ba7 to flatcar-3033 and flatcar-3066.

PR #1682, opened 2022-03-02
sys-kernel: Revert change to forbid using xfrm id 0

pothos:
The changes broke userspace (e.g., Cilium is affected because it used
id 0 for the dummy state https://github.com/cilium/cilium/pull/18789)
and we decided to revert them to give the affected software more time
to adapt (cf. https://marc.info/?t=164607426900002&r=1&w=2).

## How to use

Test that, e.g.,
```
sudo ip link add ipsec0  type xfrm dev lo
```
works again.

Backport to all channels

## Testing done

[Ongoing](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5006/cldsv/)


borkmann (approved):
Thanks for your help, Kai! Much appreciated!

jepio (approved):

jepio:
Reading this (https://marc.info/?l=linux-netdev&m=164614735600855&w=2) it appears like only one patch needs to be reverted, not both?

pothos:
For Cilium yes, only the state change needs to be reverted but I think that in general we should not start failing something like `sudo ip link add ipsec0 type xfrm dev lo` if it worked before.

pothos:
Picked for 3033, 3139, 3165

PR #1683, opened 2022-03-02
coreos-base/coreos-init: move processes to root cgroup before unbinding controllers

jepio:
# coreos-base/coreos-init: move processes to root cgroup before unbinding controllers

This fixes the issue we saw on EM, where pid controller is still busy after initramfs and it was not possible to unbind it.

Pulls in https://github.com/flatcar-linux/init/pull/63

## How to use

## Testing done

http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5007/cldsv/


jepio:
Thanks for the quick review

jepio:
Sorry, will fix the commit reference to master.

PR #1699, opened 2022-03-08
containerd: Enable SELinux labeling support by default

JAORMX:
# containerd: Enable SELinux labeling by default

This enables containerd to do appropriate SELinux labeling of containers
and files by default. This should not be problematic as Flatcar ships with
SELinux permissive by default.

This was based on the [sample config](https://github.com/containerd/cri/blob/deprecated/master/docs/config.md) provided by containerd

### DIFF
--- a/app-emulation/containerd/files/config.toml
+++ b/app-emulation/containerd/files/config.toml
@@ -27,6 +27,10 @@ runtime = "runc"
 # live restore is not supported
 no_shim = false
 
+[plugins."io.containerd.grpc.v1.cri"]
+# enable SELinux labeling
+enable_selinux = true
+
 [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
 # setting runc.options unsets parent settings
 runtime_type = "io.containerd.runc.v2"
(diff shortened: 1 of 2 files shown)


jepio:
CI has passed; I'll let @tormath1 take a look, as he might think of some dependency between this and the container selinux profiles.

tormath1:
Thanks @JAORMX for this contribution. To me this change looks good and relevant with SElinux enabled by default with Docker: https://github.com/flatcar-linux/coreos-overlay/blob/613277c65ea9288941b0db5d9683c97724d3f347/app-emulation/docker/files/docker.service#L11

Currently Flatcar is booting in permissive mode so it should not prevent application to run normally - and as mentioned in SELinux documentation:
> Check the system logs for any messages containing avc: denied. Such messages indicate that an enforcing SELinux would prevent the container from performing the logged operation. Please open an issue on Github, including the full avc log message. [^1]

Would you mind adding a changelog entry in the following folder: https://github.com/flatcar-linux/coreos-overlay/tree/main/changelog/changes ?

[^1]: https://www.flatcar.org/docs/latest/setup/security/selinux/#check-a-containers-compatibility-with-selinux-policy

JAORMX:
@tormath1 done.

PR #1700, opened 2022-03-08
coreos-base/coreos: remove rng-tool dependency

jepio:
# coreos-base/coreos: remove rng-tool dependency

rng-tools does not appear to be necessary for booting in virtual machine
environments in 2022. Back in the day the boot process would block if
there was not enough entropy to seed the system random pool, but over
the years the linux kernel made sure that the pool is force seeded if
userspace does not do so one it's own. Remove rng-tool as it is not
needed and it would require work to make sure it works (detection of
tpm/hwrng/intel cpu instructions).

## How to use

## Testing done

### DIFF
--- a/changelog/changes/2022-03-09-rngd.md
+++ b/changelog/changes/2022-03-09-rngd.md
@@ -1,0 +1,1 @@
+- Removed rngd.service because it is not essential anymore for the kernel to boot fast in VM environments ([PR#1700](https://github.com/flatcar-linux/coreos-overlay/pull/1700))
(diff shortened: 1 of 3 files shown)


JAORMX (approved):

pothos (commented):
This is about rngd.service, or? A changelog entry would be good, like `- Removed rngd.service because it is not essential anymore for the kernel to boot fast in VM environments`

jepio:
Added the changelog entry

pothos (approved):

JAORMX (approved):

PR #1688, opened 2022-03-03
Profile cleanups for updated packages from 2018

krnowak:
Clean up the profiles for the updated packages.

~Blocked on #1681.~

Needs to be merged together with https://github.com/flatcar-linux/portage-stable/pull/302.

CI: http://localhost:9091/job/os/job/manifest/5082/cldsv/

PR #1708, opened 2022-03-10
Profile cleanups for updated packages from 2019

krnowak:
Should be merged together with https://github.com/flatcar-linux/portage-stable/pull/305.

~Blocked on #1688.~

CI: http://localhost:9091/job/os/job/manifest/5207/cldsv/


JAORMX (approved):

krnowak:
CI passed (with the help of https://github.com/flatcar-linux/mantle/pull/309).

eclass/coreos-cargo.eclass
```diff
@@ -76,15 +76,36 @@ coreos-cargo_src_unpack() {
...
+				config_lines+=("${ar_line}")
+				config_lines+=("${linker_line}")
+				target_rust_target_amended=1
+				;;
+		esac
+	done <"${ECARGO_HOME}/config"
+	if [[ "${build_amended}" -eq 0 ]]; then
+	    config_lines+=('' "${b_header}" "${target_line}")
+	fi
+	if [[ "" -eq 0 ]]; then
```
jepio, line 105: what is this?

jepio (approved):
The config.toml update code is slightly scary but looks correct, other than the one comment.

eclass/coreos-cargo.eclass
krnowak: Previously, coreos-cargo appended a `[build]` and `[target.whatever]` sections to some cargo config and that worked. But after updating cargo eclass in portage stable, the config has changed too - it has a `[build]` section in the middle of the file. So coreos-cargo appended another `[build]` section and cargo bailed out complaining about an invalid toml file.

This is a workaround to keep just one `[build]` section until I maybe have some deeper look at our way of building rust packages. I hope that this is just some nonsense now and can be dropped in favor of cargo.

krnowak: Oh, wait. You mean the `[[ "" -eq 0 ]]` part. Whoops.

krnowak: Funny - `[[ "" -eq 0 ]]` seems to be always true, so we were always appending the `[target.whatever]` section to the end of file. Which is fine, because it did not exist in config generated by cargo eclass. But will fix, it was supposed to be `[[ "${target_rust_target_amended}" -eq 0 ]]`.

krnowak:
> The config.toml update code is slightly scary but looks correct, other than the one comment.

Thanks. I'm rerunning the packages matrix job with the fix.

krnowak:
CI passed after the fix too.

dongsupark (approved):

PR #1709, opened 2022-03-10
sys-apps/systemd: enable systemd-sysext.service

pothos:
The systemd-sysext.service activates sysext images on boot.
Enable it by default.

## How to use

Check that it is on

## Testing done
Manually booted the image and check that the service got started (enabled is not show here because Flatcar uses hardcoded symlinks)
```
$ systemctl status --no-pager systemd-sysext
○ systemd-sysext.service - Merge System Extension Images into /usr/ and /opt/
     Loaded: loaded (/usr/lib/systemd/system/systemd-sysext.service; disabled; vendor preset: disabled)
     Active: inactive (dead)
  Condition: start condition failed at Thu 2022-03-10 17:31:01 UTC; 18s ago
       Docs: man:systemd-sysext.service(8)

Mar 10 17:31:01 localhost systemd[1]: Merge System Extension Images into /usr/ and /opt/ was skipped because all trigger condition checks failed.
```

[started](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5092/cldsv/) but failed because I removed the other branch too early

### DIFF
--- a/changelog/changes/2022-03-10-systemd-sysext-service.md
+++ b/changelog/changes/2022-03-10-systemd-sysext-service.md
@@ -1,0 +1,1 @@
+- Enabled the `systemd-sysext.service` by default, to disable you have to mask it
--- a/sys-apps/systemd/systemd-250.3.ebuild
+++ b/sys-apps/systemd/systemd-250.3.ebuild
@@ -560,6 +560,8 @@ multilib_src_install_all() {
 	if use gnuefi; then
 		builddir_systemd_enable_service sysinit.target systemd-boot-update.service
 	fi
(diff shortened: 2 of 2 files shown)


jepio:
Should we be doing the systemd-sysext activation from the initramfs and overlaying on /sysroot? We should give this some thought.

pothos:
I would use the standard service to merge the extensions at the same time as other distros

pothos:
Not sure if it even works from the initramfs because systemd uses some `/run` paths to set up the overlay mount. I think sysinit.target is good for our purposes and if we want we can even try to start it as early as possible before we would go the route of setting up the overlays before the final init system starts.

PR #1712, opened 2022-03-11
Add sssd to list of SELinux modules enabled

JAORMX:
# Include `sssd` SELinux module in base policy

sssd it being provided by flatcar and it was missing from the list.

This inclusion should appropriately label sssd-related files.

Related-Bug: https://github.com/flatcar-linux/Flatcar/issues/673

## How to use

## Testing done

### DIFF
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -58,7 +58,7 @@ sys-libs/glibc nscd
 dev-libs/cyrus-sasl kerberos -berkdb -gdbm
 
 # don't build manpages for sssd
-sys-auth/sssd -python samba kerberos gssapi ssh sudo
+sys-auth/sssd -python samba kerberos gssapi ssh sudo selinux
 
 # enable logging command-line options in update_engine
 dev-cpp/glog gflags


krnowak:
Hm, I see in the gentoo repo that there is a `sec-policy/selinux-sssd` package. Maybe this should be brought into portage-stable instead and then selinux USE flag should be also enabled for sssd in profiles?

JAORMX:
@krnowak that could be an option! Although, it's already included by default in Flatcar, so why not include it to the base list? the package is merely a `MODS` addition after all https://gitweb.gentoo.org/repo/gentoo.git/tree/sec-policy/selinux-sssd/selinux-sssd-2.20200818-r2.ebuild

krnowak:
The reason is that we aim for having the least amount of divergence between flatcar and gentoo, so we have less things to maintain.

JAORMX:
@krnowak done

krnowak:
@JAORMX: Thanks!

I see that you added the `sec-policy/selinux-sssd` package to [coreos-overlay](https://github.com/flatcar-linux/coreos-overlay), but you made no modifications to it. In such case, we add such unmodified packages to [the portage-stable repo](https://github.com/flatcar-linux/portage-stable) instead.

Another thing is that `sec-policy/selinux-sssd` is pulled by `sys-auth/sssd` only if `selinux` USE flag is enabled for this package, which is not the case currently - we enable it for selected packages and `sys-auth/sssd` is not one of them. You can enable it in `profiles/coreos/base/package.use` - [please find a line for sssd and add `selinux` there](https://github.com/flatcar-linux/coreos-overlay/blob/main/profiles/coreos/base/package.use#L62).

With that done, I can start a test build in our CI.

JAORMX:
@krnowak created https://github.com/flatcar-linux/portage-stable/pull/307

krnowak:
Cool, CI is crunching now those PRs. Thanks!

krnowak:
There as a dev-container failure, that's unrelated to this PR, otherwise CI passed (had to rerun qemu tests, because some of them were flaky).

You could download the [qemu image](https://bucket.release.flatcar-linux.net/flatcar-jenkins/developer/developer/boards/amd64-usr/2022.03.17%2bdev-flatcar-master-5164/flatcar_production_qemu_image.img.bz2) and [the bash script](https://bucket.release.flatcar-linux.net/flatcar-jenkins/developer/developer/boards/amd64-usr/2022.03.17%2bdev-flatcar-master-5164/flatcar_production_qemu.sh) to verify if it works for you.

JAORMX:
@krnowak I haven't been able to test this cause I'm on mac right now and haven't rewritten the script to work on mac yet. Should CI be re-ran then?

krnowak:
No, no need for rerunning the tests. dev-container failure is unrelated and I have a fix for it, but didn't manage to test it and merge yet.

I'll check the contents of the image.

krnowak:
Files in `/var/lib/sss` and `/etc/sssd` are labeled as [its refpolicy](https://github.com/SELinuxProject/refpolicy/blob/master/policy/modules/services/sssd.fc) mandates, `/var/log/sssd`, `/usr/bin/sssd` are still `unlabeled_t` stuff. So it's some improvement.

krnowak:
Thanks for the PR! I think we will eventually also need to focus on having more coverage of SELinux labeling.

tormath1:
@krnowak yes, that's the goal of this issue: https://github.com/flatcar-linux/Flatcar/issues/673 and I revamped this PR: https://github.com/flatcar-linux/scripts/pull/66 to get a fully labeled system but it breaks everything in enforcing mode :)

PR #1713, opened 2022-03-11
profiles: disable SYMLINK_LIB

pothos:
WIP

## Testing done


profiles/coreos/targets/generic/make.defaults
```diff
@@ -98,3 +98,6 @@ ACCT_USER_SYSTEMD_RESOLVE_ID=245
 # tss seems to be one of those users with a mismatching UID/GID
 ACCT_GROUP_TSS_ID=252
 ACCT_USER_TSS_ID=236
+
+# Disable creation of /usr/lib as symlink
+SYMLINK_LIB="no"
```
jepio, line 103: Could we switch to 17.1 profile instead?

pothos: Maybe, I'm not sure what other changes this implies. My idea was to limit the changes for now and only align the amd64 board with the arm64 board and leave the SDK out.

jepio: i would love it if we could make the jump to 17.1 independently, but I acknowledge that it's more work.

sys-apps/systemd/systemd-250.3.ebuild
```diff
@@ -560,6 +560,8 @@ multilib_src_install_all() {
 	if use gnuefi; then
 		builddir_systemd_enable_service sysinit.target systemd-boot-update.service
 	fi
+	# Flatcar: enable systemd-sysext.service
+	builddir_systemd_enable_service sysinit.target systemd-sysext.service
```
krnowak, line 564: It's a nitpick, but could you put that below the `enable reboot.target` comment separated with an empty line? This section is basically following the contents of upstream's `90-systemd.preset`, so I would like to avoid mixing in our custom stuff like enabling systext here. Thanks!

krnowak (approved):

changelog/changes/2022-03-10-systemd-sysext-service.md
```diff
@@ -0,0 +1 @@
+- Enabled the `systemd-sysext.service` by default, to disable you have to mask it
```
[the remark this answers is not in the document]
pothos: ah, right

sys-apps/systemd/systemd-250.3.ebuild
pothos: :+1:

PR #1720, opened 2022-03-14
sys-kernel/coreos-sources: revert commit which breaks networking on M4 instances

jepio:
# sys-kernel/coreos-sources: revert commit which breaks networking on M4 instances

We bisected m4 networking failure down to a commit that has already been backported to all stable kernels. Revert the commit while upstream discussions happen.

## How to use

If the ami from the test run is still around:
```
./bin/kola spawn --aws-type m4.large --board amd64-usr  --aws-iam-profile jenkins-test --platform=aws  --aws-ami=ami-03ecf733c5ca3afeb --aws-region=us-west-2 --verbose --aws-credentials-file ~/.aws/credentials
```

## Testing done

Successfully tested here: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/aws/1242/tapResults/.

### DIFF
--- a/changelog/bugfixes/2022-03-14-kernel-msi-revert.md
+++ b/changelog/bugfixes/2022-03-14-kernel-msi-revert.md
@@ -1,0 +1,1 @@
+- Reverted the Linux kernel commit which broke networking on AWS instances which use Intel 82559 NIC (c4/m4) ([Flatcar#665](https://github.com/flatcar-linux/Flatcar/issues/665), [PR#1720](https://github.com/flatcar-linux/coreos-overlay/pull/1720))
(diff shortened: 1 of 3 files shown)


pothos:
What target branch is this?

jepio:
This one is for flatcar-3033, and I wanted to merge it after https://github.com/flatcar-linux/coreos-overlay/pull/1715 is merged.

pothos:
Ok, a changelog entry would be good, too

jepio:
> What target branch is this?

Wanted to see this work:

> Base automatically changed from linux-5.10.105-flatcar-3033 to flatcar-3033

pothos (approved):

PR #1721, opened 2022-03-14
sys-kernel/bootengine: re-add missing modules

pothos:
This pulls in
https://github.com/flatcar-linux/bootengine/pull/40
to add the kernel modules back that disappeared compared to Stable
3033.x.y with the 5.10 kernel.

## How to use

Port to beta, alpha - maybe with a bootengine backport branch

## Testing done

[build passed](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5124/cldsv/), and manual checks in https://github.com/flatcar-linux/bootengine/pull/40

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)

PR #1723, opened 2022-03-15
sys-kernel/coreos-sources: revert commit which breaks networking on M4 instances

jepio:
# sys-kernel/coreos-sources: revert commit which breaks networking on M4 instances

This is the same as https://github.com/flatcar-linux/coreos-overlay/pull/1720 but for main. 

## How to use

## Testing done


jepio:
CI running here: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5137/cldsv/

jepio:
Depends on https://github.com/flatcar-linux/coreos-overlay/pull/1721 which is already merged into all branches.

jepio:
Some mistakes where made (tried to reuse binary packages but this pulled in stale bootengine), but final test run passed.

jepio:
Merging since this is tested and already reviewed for stable.

PR #1724, opened 2022-03-15
coreos-base/hard-host-depends: Stop pulling in intltool

krnowak:
The tool is deprecated, nothing pulls that in any more and it has a
dependency on dev-perl/XML-Parser, an updated version of which would
want to pull a bunch of new packages through dev-perl/libwww-perl.
Avoid the hassle and drop the tool.

Actually, I found out that realmd needed this tool, but it was not specified in BDEPEND. I created a patch to migrate the project from intltool to gettext. ~The patch is submitted to upstream: https://gitlab.freedesktop.org/realmd/realmd/-/merge_requests/32~ The patch got merged by upstream.

Realmd seems to need some polkit ITS rule files (used for i18n), so I added a small custom package that installs them to the SDK, so I avoid adding whole polkit there.

Probably the proper fix would be to either:
- make realmd build system to respect the `--disable-nls` flag when it generates the polkit policy file (I wasn't aware of it when I wrote the patch for migrating to gettext), or
- make gettext to find the ITS rules in `/build/amd64-usr` too, not only in `/`.

Needs to be merged together with https://github.com/flatcar-linux/portage-stable/pull/306

CI: http://localhost:9091/job/os/job/manifest/5142/cldsv/


krnowak:
CI passed.

tormath1:
Awesome, contribution to `realmd` is great. Thanks :) 
Just have a quick question regarding `dev-util/polkit-its-rules`, you say it's required by `realmd` - should we add this package to the `RDEPEND` (or `PDEPEND`) of the `realmd` ebuild so Portage can properly resolve dependency and later help to understand the _why_ of this package ?

krnowak:
> Awesome, contribution to `realmd` is great. Thanks :) Just have a quick question regarding `dev-util/polkit-its-rules`, you say it's required by `realmd` - should we add this package to the `RDEPEND` (or `PDEPEND`) of the `realmd` ebuild so Portage can properly resolve dependency and later help to understand the _why_ of this package ?

I think it should be added to `BDEPEND`, because `gettext` needs it at a build time. `gettext` was looking at the files in `/usr/share/gettext` only (this is where files from `dev-util/polkit-its-rules` end up being), ignoring similar stuff installed in `/build/amd64-usr/usr/share/gettext` (this is where the files from `sys-auth/polkit` are installed), which is why `DEPEND=sys-auth/polkit` was insufficient.

I have now found out about the `GETTEXTDATADIRS` env var, so I'll try settting that one somewhere to `/build/amd64-usr/usr/share/gettext`, so maybe `dev-util/polkit-its-rules` will not be necessary.

tormath1 (approved):

PR #1735, opened 2022-03-18
sys-libs/pam: Update to 1.5.1_p20210622 

sayanchowdhury:
# sys-libs/pam: Update to 1.5.1_p20210622 

## Testing done

Jenkins CI passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5166/cldsv/


JAORMX (approved):

dongsupark:
CI passed.
It would be really great if @krnowak could have another look.

sys-libs/pam/Manifest
```diff
@@ -1,2 +1,4 @@
...
-DIST Linux-PAM-1.5.1.tar.xz 972964 BLAKE2B a1714569587a383fa8211b23765c66b08b18dc2808c1521a904171dc2886cced56e9afa27408e8a9d5eec6226b31390dc8f14434071370f4e1147c77ce8b36ac SHA512 1db091fc43b934dde220f1b85f35937fbaa0a3feec699b2e597e2cdf0c3ce11c17d36d2286d479c9eed24e8ca3ca6233214e4dff256db47249e358c01d424837
+DIST Linux-PAM-1.5.2-docs.tar.xz 443276 BLAKE2B eb0376022bd17d472cfbb7d757acd9a7743d68929ac604bc7fbc27e87b35f0424e9459afff8110ee094c6914cef6e6f483e6eeb2fc2ec24909477ec53f600fa2 SHA512 69bbb52b3855a1ff3066ffe3ae1890440516311260ddfb20a7a7ea7beeb42484ea085080e3206f23933edf4a695a13f2317e33dffcab2b29f6e9e210d493e1f0
+DIST Linux-PAM-1.5.2.tar.xz 988784 BLAKE2B a835034cd239bc9377419c13dda45276e8e64a33fcf714a1957ff41112fbb6dce0be8e9773afc82458a04d54bf146a0c26117d7170521fecdc0c98184cef5f4f SHA512 fa16350c132d3e5fb82b60d991768fb596582639841b8ece645c684705467305ccf1302a0147ec222ab78c01b2c9114c5496dc1ca565d2b56bf315f29a815144
```
krnowak, line 2: We probably don't need these here.

krnowak (approved):
One nitpick, otherwise looks good. We probably should eventually drop this locked accounts functionality - we tried to upstream it I think, and it was more or less rejected, because there is some other/better way to do it.

sys-libs/pam/Manifest
dongsupark: Good catch. Cleaned it up.

PR #1745, opened 2022-03-21
profiles: disable su USE flag for util-linux

dongsupark:
As `sys-apps/shadow` has its own su binary, `sys-apps/util-linux` should not have its own su binary. Otherwise, build will simply fail.
Disable su USE flag for util-linux.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/310.

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5192/cldsv

- Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) (see https://github.com/flatcar-linux/portage-stable/pull/310)

### DIFF
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -143,7 +143,11 @@ sys-apps/iproute2 elf
 # Ideally util-linux should have the su binary, but that is currently not
 # possible, because of a bunch of additional dependencies in SDK like
 # pam_sssd in baselayout.
+#
+# Disable su for util-linux to avoid conflict with sys-apps/shadow, which
+# has its own su binary.
 sys-apps/shadow su
+sys-apps/util-linux -su
 
 # Enable kerberos support for NFS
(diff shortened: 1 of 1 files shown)


dongsupark:
> Do you know any reasons to use one over the other? Fedora and Debian use the one from util-linux.

Yes, I agree, we should ideally use one from util-linux.
As discussed [here](https://github.com/flatcar-linux/coreos-overlay/pull/1660#discussion_r811243309) recently however, installing `util-linux[su]` opens up a can of worms.
That's why we still use su from shadow.

pothos:
Okay, maybe we can set it for the boards only to avoid having to deal with the SDK's pam settings (or we add some use flag to our baselayout package that is only set for the boards where sss stuff gets activated)

dongsupark:
Thanks for the review.
Will then merge the PR for the next Alpha.

> Okay, maybe we can set it for the boards only to avoid having to deal with the SDK's pam settings (or we add some use flag to our baselayout package that is only set for the boards where sss stuff gets activated)

+1

PR #1750, opened 2022-03-23
eclass/coreos-go.eclass: strip Go binaries by default

pothos:
The size contains not only of the /usr partition but also the /boot
partition require that we reduce the size of binaries as much as
possible.
Strip all Go binaries by default.

## How to use
check that tests pass and that it actually reduced the image size and vmlinuz size

## Testing done
[onging](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5220/cldsv/)

↑ not needed I think

### DIFF
--- a/eclass/coreos-go.eclass
+++ b/eclass/coreos-go.eclass
@@ -56,7 +56,7 @@ go_build() {
 
 	${EGO} build -v \
 		-p "$(makeopts_jobs)" \
-		-ldflags "${GO_LDFLAGS} -extldflags '${LDFLAGS}'" \
+		-ldflags "${GO_LDFLAGS} -s -w -extldflags '${LDFLAGS}'" \
 		${COREOS_GO_MOD:+-mod "${COREOS_GO_MOD}"} \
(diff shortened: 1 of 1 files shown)


jepio:
What about stack traces from go binaries? They become useless without these options, don't they?

pothos:
I think not:
```
package main

func main() {
	panic("hello")
}
```
compiles with `go build -ldflags "-s -w"` still gives
```
panic: hello

goroutine 1 [running]:
main.main()
	/var/tmp/main/main.go:4 +0x27
```

jepio (approved):
OK; but let's wait and see if this actually reduces the artifact size in any noticable way.

pothos:
The binaries are smaller which mostly has an effect for the ext2 `/usr` partition where compression isn't used. The 3 MiB difference we get for the ignition binary in the initramfs is just < 1 MiB effective size difference for the compressed vmlinuz file. Still I think it's worth doing this - we still have to add cryptsetup and clevis to the initramfs for full Ignition v3 support and it will be hard to squeeze all this in while still having some space left for future update bloat.

PR #1751, opened 2022-03-23
coreos-devel/mantle: Bump to latest commit

krnowak:


pothos (approved):
(also pick for Stable/Beta/Alpha)

PR #1757, opened 2022-03-24
.github/workflows: install lbzip2 to speed up sdk creation

jepio:
# .github/workflows: install lbzip2 to speed up sdk creation

Our github actions use cork to create an sdk chroot, which pulls down bzipped
archives. The runners have 2 CPUs, so this unpacking could be faster if we
installed lbzip2. Cork transparently uses lbzip2.

## How to use

## Testing done

### DIFF
--- a/.github/workflows/setup-flatcar-sdk.sh
+++ b/.github/workflows/setup-flatcar-sdk.sh
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+sudo apt-get install -y lbzip2
+
 CORK_VERSION=$(curl -sL https://api.github.com/repos/kinvolk/mantle/releases/latest | jq -r .tag_name | sed -e 's/^v//')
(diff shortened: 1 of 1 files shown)


jepio:
Tested here: https://github.com/flatcar-linux/coreos-overlay/actions/runs/2033183669, much faster :)

JAORMX (approved):

.github/workflows/setup-flatcar-sdk.sh
```diff
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+sudo apt-get install -y lbzip2
```
pothos, line 5: This may fail depending on whether the OS image is rebuilt often or not and if lbzip2 is updated. A `sudo apt update` first would make it a bit more robust.

jepio: I'll add a sudo apt update if we ever observe it failing.

PR #1769, opened 2022-03-29
coreos-base/afterburn: sync with 5.2.0

tormath1:
In this PR, we sync the ebuild with `coreos/afterburn-5.2.0` and apply Flatcar patches:

* `coreos-base/afterburn/files/0001-Revert-remove-cl-legacy-feature.patch`: to bring back the `cl-legacy` feature with  `vagrant_virtualbox` provider. `write_ssh_keys` keeps the upstream definition to avoid pull `update-ssh-keys` dependency.
* `coreos-base/afterburn/files/0002-util-cmdline-Handle-the-cmdline-flags-as-list-of-sup.patch`: to add the backward compatibility coreos/flatcar for the cmdline
* `coreos-base/afterburn/files/0003-encode-information-for-systemd-networkd-wait-online.patch`: see https://github.com/coreos/afterburn/pull/728

<hr>

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5362/cldsv/

## Note for reviewers

In `0001-Revert-remove-cl-legacy-feature.patch`, the patch has been reworked a bit especially the `vagrant_virtualbox` provider which was not compatible with recent codebase.

Closes: https://github.com/flatcar-linux/Flatcar/issues/387, https://github.com/flatcar-linux/Flatcar/issues/220


coreos-base/afterburn/files/coreos-metadata-sshkeys@.service
```diff
@@ -12,3 +12,4 @@ ExecStart=/usr/bin/coreos-metadata ${COREOS_METADATA_OPT_PROVIDER} --ssh-keys=%i
 [Install]
 DefaultInstance=core
 RequiredBy=multi-user.target
+Alias=afterburn-sshkeys@.service
```
pothos, line 15: I think we also need an ExecStartPost action here to append the created file in `authorized_keys.d/` to `authorized_keys`, now that update-ssh-keys is not used but the file is simply dropped inside the folder.

pothos: (or you restore the full cl-legacy features with calling update-ssh-keys)

coreos-base/afterburn/files/0001-Revert-remove-cl-legacy-feature.patch
```diff
@@ -0,0 +1,591 @@
+From ba861f48c0ca6abb0944c362b4bf9e79bfbb7bc8 Mon Sep 17 00:00:00 2001
```
pothos, line 1: Looks ok but I doubt this large patch will cleanly apply after a few months, thus the idea of using `sed` in `ExecStartPost`https://github.com/flatcar-linux/Flatcar/issues/387#issuecomment-1080532259

coreos-base/afterburn/files/coreos-metadata-sshkeys@.service
tormath1: Oh ok understood. To me `update-ssh-keys` was just a lib to write into `./authorized_keys_d/...` and afterburn was doing the merge later.
Indeed things happen there: https://github.com/flatcar-linux/update-ssh-keys/blob/012bc8d9d86f044b0c87ca749563879eefbde25c/src/lib.rs#L306-L309 and there: https://github.com/flatcar-linux/afterburn/blob/bb3cdd0b895e9803c843d1e1831f9c0613e35e02/src/providers/mod.rs#L90-L92

coreos-base/afterburn/files/0001-Revert-remove-cl-legacy-feature.patch
tormath1: Yes - I just find more consistent to get every changes applied in one place. Something we can do: reduce this patch to bring back only  `src/providers/vagrant_virtualbox/mod.rs` and `sed` the `AWS/GCP/AFTERBURN` prefixes.

coreos-base/afterburn/files/coreos-metadata-sshkeys@.service
```diff
@@ -8,7 +8,10 @@ Restart=on-failure
 RestartSec=10
 Environment=COREOS_METADATA_OPT_PROVIDER=--cmdline
 ExecStart=/usr/bin/coreos-metadata ${COREOS_METADATA_OPT_PROVIDER} --ssh-keys=%i
+ExecStartPost=/usr/bin/bash -c '/usr/bin/cat /home/%i/.ssh/authorized_keys.d/*'
+StandardOutput=append:/home/%i/.ssh/authorized_keys
```
pothos, line 12: This makes the file grow on each run. Maybe we can call `update-ssh-keys` directly since we still ship the binary?

tormath1: Since we're removing the `update-ssh-keys` dependency I was thinking of removing the CLI too since it seems quite outdated.
Not sure about the user impact though - so yes, it's maybe better for now to use it directly.

pothos (commented):
Looks good, but haven't done a manual test

Can you add an update changelog entry?

tormath1:
@pothos done for the changelog entry :)

tormath1:
CI totally passed on Equinix Metal too now.

tormath1:
@flatcar-linux/flatcar-maintainers just tested on all cloud-providers supported by Afterburn - CI is good :)

coreos-base/afterburn/files/0002-util-cmdline-Handle-the-cmdline-flags-as-list-of-sup.patch
```diff
@@ -0,0 +1,111 @@
+From db00a15023798d0eae711b78fddc7745efade51e Mon Sep 17 00:00:00 2001
+From: Mathieu Tortuyaux <mtortuyaux@microsoft.com>
+Date: Tue, 29 Mar 2022 16:01:19 +0200
+Subject: [PATCH 2/3] util/cmdline: Handle the cmdline flags as list of
+ supported names
+
+Flatcar Container Linux uses flarcar.oem.id as kernel parameter and has
+support for coreos.oem.id. Afterburn's ignition.oem.id is not supported
+and even if, it would not be the sole option.
+Handle both flarcar.oem.id and coreos.oem.id. The first name in the
```
krnowak, line 10: Feel free to ignore it, if regenerating the patch is too much of a PITA. :)

```suggestion
Handle both flatcar.oem.id and coreos.oem.id. The first name in the
```

coreos-base/afterburn/files/coreos-metadata.service
```diff
@@ -8,7 +8,9 @@ Restart=on-failure
 RestartSec=10
 Environment=COREOS_METADATA_OPT_PROVIDER=--cmdline
 ExecStart=/usr/bin/coreos-metadata ${COREOS_METADATA_OPT_PROVIDER} --attributes=/run/metadata/flatcar
+ExecStartPost=/usr/bin/sed --in-place "s/AFTERBURN/COREOS/ ; s/AWS/EC2/ ; s/GCP/GCE/" /run/metadata/flatcar
```
krnowak, line 11: Should those be with the `g` flag in case there are more instances to replace?
```suggestion
ExecStartPost=/usr/bin/sed --in-place "s/AFTERBURN/COREOS/g ; s/AWS/EC2/g ; s/GCP/GCE/g" /run/metadata/flatcar
```

krnowak (commented):
One question: have we tried to upstream the second patch? That way our diff would be only by adding some names to an array.

coreos-base/afterburn/files/coreos-metadata.service
tormath1: Good question - I think we can safely add the `g` flag. From what I seen in the codebase, it's always `AFTERBURN_$PROVIDER_$ATTRIBUTE`, so we _should_ not have redundancy on the same line.
However, it's better to preserve consistency if we want to avoid confusing situation while we maintain backward compatibility.

krnowak:
Oh, and by the way, seems like this is now a good candidate for dropping CROS_WORKON stuff from the ebuild. But that can be done later.

tormath1:
> Oh, and by the way, seems like this is now a good candidate for dropping CROS_WORKON stuff from the ebuild. But that can be done later.

Any tracking issue for this topic to keep a trace ?

> One question: have we tried to upstream the second patch? That way our diff would be only by adding some names to an array.

You're right, that would make sense to try to upstream this change - afterburn is not distro specific related so it would be a good change for other distro interested to ship afterburn.

krnowak:
> > Oh, and by the way, seems like this is now a good candidate for dropping CROS_WORKON stuff from the ebuild. But that can be done later.
> 
> Any tracking issue for this topic to keep a trace ?

Hm, not really. I remember talking about it when I was doing the one-pager for the portage-stable-as-gentoo-snapshot thing as a possible future work item.

krnowak (approved):

PR #1775, opened 2022-03-30
profiles: delete keywords for gnutls

dongsupark:
As we update gnutls to 3.7.3-r1 which is already stable, there is no need to accept keywords for gnutls.
Delete.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/315.

## Testing done

- Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) (see https://github.com/flatcar-linux/portage-stable/pull/315)

### DIFF
--- a/profiles/coreos/base/package.accept_keywords
+++ b/profiles/coreos/base/package.accept_keywords
@@ -19,8 +19,6 @@
 # needed for arm64 sdk
 =dev-util/pahole-1.20 ~arm64
 
-=net-libs/gnutls-3.7.1 ~amd64 ~arm64
-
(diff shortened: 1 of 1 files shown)


jepio (approved):

krnowak:
I'd like to notice that gnutls-3.7.3-r1 is not stable for arm64, so accept keyword for it should remain. The only reason why this is not a problem currently is that gnutls is apparently not even built on arm64. But this may change in future, when we will want to make sure that both amd64 and arm64 has more or less the same package set.

I have a fix for it in my package updates, 2021 edition branch, so this will be fixed eventually.

PR #1783, opened 2022-03-31
Profile cleanups for package updates, 2020 edition

krnowak:
Should be merged together with https://github.com/flatcar-linux/portage-stable/pull/320.

Please see the PR in portage-stable for more info.

CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5298/cldsv/

### DIFF
--- a/coreos/config/env/net-dns/bind-tools
+++ b/coreos/config/env/net-dns/bind-tools
@@ -1,2 +1,10 @@
-# Work around failing OpenSSL feature detection.
-EXTRA_ECONF="--without-ecdsa"
+# Work around lame function attribute checks when cross-compiling.
+# Linker emits a bunch of irrelevant warnings and that trips configure
+# script up into thinking that those attributes are not
+# supported. Upstream bind-tools already fixed that by being smarter
+# and grepping for -Wattributes instead, but we are not yet packaging
+# it. We also know that constructor and destructor attributes are
+# supported - they are available since at least gcc 3.4.
+
+EXTRA_ECONF+=" ax_cv_have_func_attribute_constructor=yes"
+EXTRA_ECONF+=" ax_cv_have_func_attribute_destructor=yes"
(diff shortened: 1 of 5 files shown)


krnowak:
CI passed, but still waiting for #318 to get in first.

coreos/config/env/net-dns/bind-tools
```diff
@@ -1,2 +1,10 @@
...
-EXTRA_ECONF="--without-ecdsa"
+# Work around lame function attribute checks when cross-compiling.
+# Linker emits a bunch of irrelevant warnings and that trips configure
+# script up into thinking that those attributes are not
+# supported. Upstream bind-tools already fixed that by being smarter
+# and grepping for -Wattributes instead, but we are not yet packaging
+# it. We also know that constructor and destructor attributes are
+# supported - they are available since at least gcc 3.4.
+
+EXTRA_ECONF+=" ax_cv_have_func_attribute_constructor=yes"
```
tormath1, line 9: I would almost be in favor of moving these options to the ebuild configuration stepf itself and move the ebuild to `::coreos-overlay` to avoid spreading the configuration across various places...

krnowak: I think that when I will focus more on the coreos-overlay packages, we will get to the point where we need to decide whether it is more important to keep the packages in portage-stable/gentoo at the expense of unified configuration, or to move them to coreos-overlay and modify them to our needs, but at a price of more maintainership effort needed.

For now, I'd say that I'd favor the former.

krnowak: Also, bind-tools got a better check in newer versions, but gentoo is still packaging 9.16.x, instead of 9.18.x.

jepio: I think this is the lesser evil and allows us to consume unchanged upstream ebuilds in portage-stable.

jepio (approved):
LGTM

PR #1784, opened 2022-03-31
ignition: Enable SELinux relabeling

JAORMX:
# ignition: Enable SELinux relabeling

This removes the LD Flag that disabled SELinux relabeling. It defaulted to true, thus enabling the feature.

Related Bug: https://github.com/flatcar-linux/Flatcar/issues/673

- [ ] CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5307/cldsv/
- [ ] changelog entry

### DIFF
--- a/sys-apps/ignition/ignition-9999.ebuild
+++ b/sys-apps/ignition/ignition-9999.ebuild
@@ -46,7 +46,7 @@ PATCHES=(
 
 src_compile() {
 	export GO15VENDOREXPERIMENT="1"
-	GO_LDFLAGS="-X github.com/flatcar-linux/ignition/v2/internal/version.Raw=${PV} -X github.com/flatcar-linux/ignition/v2/internal/distro.selinuxRelabel=false -X github.com/flatcar-linux/ignition/v2/internal/distro.writeAuthorizedKeysFragment=false" || die
+	GO_LDFLAGS="-X github.com/flatcar-linux/ignition/v2/internal/version.Raw=${PV} -X github.com/flatcar-linux/ignition/v2/internal/distro.writeAuthorizedKeysFragment=false" || die
 	go_build "${COREOS_GO_PACKAGE}/internal"
 }
 
--- a/sys-kernel/bootengine/bootengine-9999.ebuild
+++ b/sys-kernel/bootengine/bootengine-9999.ebuild
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="https://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="85d325684c5c7817aad230b801f302903e9c6f69" # flatcar-master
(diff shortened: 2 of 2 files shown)


sys-apps/ignition/ignition-9999.ebuild
```diff
@@ -46,7 +46,7 @@ PATCHES=(
 
 src_compile() {
 	export GO15VENDOREXPERIMENT="1"
-	GO_LDFLAGS="-X github.com/flatcar-linux/ignition/v2/internal/version.Raw=${PV} -X github.com/flatcar-linux/ignition/v2/internal/distro.selinuxRelabel=false -X github.com/flatcar-linux/ignition/v2/internal/distro.writeAuthorizedKeysFragment=false" || die
+	GO_LDFLAGS="-X github.com/flatcar-linux/ignition/v2/internal/version.Raw=${PV} -X github.com/flatcar-linux/ignition/v2/internal/distro.writeAuthorizedKeysFragment=false" || die
```
pothos, line 49: @tormath1 oh, I think writeAuthorizedKeysFragment should be true, right? Otherwise changes done by Ignition get lost when `update-ssh-keys` runs, or?

pothos: and we would need to trigger `update-ssh-keys` once because `internal/authorized_keys_d/authorized_keys_d.go` is gone

pothos: Here: https://github.com/flatcar-linux/Flatcar/issues/699

tormath1:
To verify it works as expected, we can run the following test:
```
sudo ./bin/kola run ... --remove=false --qemu-image ./flatcar_production_qemu_image.img cl.ignition.instantiated.enable-unit
```

This test will create the following service file:
```
/etc/systemd/system/echo@.service
```

which has the following label:
```
ls -aliZ /etc/systemd/system/echo@.service
29 -rw-r--r--. 1 root root system_u:object_r:unlabeled_t:s0 124 Mar 31 14:44 /etc/systemd/system/echo@.service
```

with this PR, result should be the following (obtained here after a `restorecon` command):
```
ls -aliZ /etc/systemd/system/echo@.service
29 -rw-r--r--. 1 root root system_u:object_r:etc_t:s0 124 Mar 31 14:44 /etc/systemd/system/echo@.service
```

One existing Mantle test could be updated consequently.

tormath1:
CI failed for the following tests (94 tests, 82 ok, 11 not ok):
- cl.disk.raid1.root
- cl.ignition.oem.reuse
- cl.disk.raid0.root
- cl.ignition.oem.regular
- cl.ignition.v1.btrfsroot
- cl.ignition.v1.ext4root
- cl.ignition.oem.wipe
- cl.ignition.v2.ext4root
- cl.ignition.v1.xfsroot
- cl.ignition.v2.xfsroot
- cl.ignition.v2.btrfsroot

Mainly for these two reasons:
https://gist.github.com/tormath1/f67ac94b4a035e7f6b17267a60de8726

JAORMX:
Unfortunately, I ran out of cycles for working on this.

PR #1790, opened 2022-04-01
app-admin/sudo: Sync with Gentoo upstream; updates to 1.9.8_p2

sayanchowdhury:
# app-admin/sudo: Sync with Gentoo upstream; updates to 1.9.8_p2

## Testing done

CI passed http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5306/cldsv/


tormath1:
do you think that would make sense to pull 1.9.9 (which comes with OpenSSLv3 support) or 1.9.10 ? Based on the changelogs it should not bring many turbulence :thinking:

app-admin/sudo/sudo-1.9.8_p2-r1.ebuild
```diff
@@ -200,8 +202,7 @@ src_install() {
 		doins "${T}"/ldap.conf.sudo
 		fperms 0440 /etc/ldap.conf.sudo
 
-		insinto /etc/openldap/schema
-		newins doc/schema.OpenLDAP sudo.schema
+		#Flatcar: we don't ship OpenLDAP schemas
```
krnowak, line 205: I try to avoid removing lines if I can just comment them out.

app-admin/sudo/README.md
```diff
@@ -0,0 +1,8 @@
+## Flatcar changes
+- Remove Perl Runtime Dependency
+- Remove OpenLDAP schema files for sudo
+```
+insinto /etc/openldap/schema
+newins doc/schema.OpenLDAP sudo.schema
+```
```
krnowak, line 7: If you comment these lines out instead of removing, then you won't need to add them here.

krnowak (commented):
Some nitpicks, but otherwise looks good. Whether you will want to update to 1.9.{9,10} it's up to you. We will eventually get them.

sayanchowdhury:
@tormath1 Good point. Let's jump to 1.9.9. Let's the latest version stabilize a bit more :)

app-admin/sudo/sudo-1.9.8_p2-r1.ebuild
sayanchowdhury: Sure. Will just comment it out here.

sayanchowdhury:
Closing this issue in favor of #2018

PR #1794, opened 2022-04-01
coreos-base/coreos-init: run update-ssh-keys once after Ignition

pothos:
This pulls in
https://github.com/flatcar-linux/init/pull/66
to fix the problem that Ignition keys would be lost as soon as
update-ssh-keys runs. This is done by placing Ignition's keys in as
files in the authorized_keys.d folder and calling update-ssh-keys after
Ignition ran.

## How to use

Backport to Alpha

## Testing done

[tests seem to pass so far](http://192.168.42.7:8080/job/os/job/manifest/5342/cldsv/)


krnowak (approved):
Could you bump the ebuild symlink to 2.13.0-r1?

PR #1801, opened 2022-04-08
sys-kernel/coreos-modules: build RAPL module

tormath1:
This enables support for the Intel Running Average Power Limit (RAPL)
technology via MSR interface, which allows power limits to be enforced
and monitored on modern Intel processors.

It can be useful for energy consumption monitoring tools.

src: https://github.com/torvalds/linux/blob/master/drivers/powercap/Kconfig

<hr>

See also: https://gist.github.com/tormath1/140a625178588ae9e5b76d7342012cfc

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5382/cldsv/


tormath1:
ARM64 CI is failing, I think it's expected since it's Intel specific thing - will move the config to `amd64` dedicated config.

krnowak (approved):

PR #1815, opened 2022-04-13
[RFC] dev-lang/python-oem: use python from python-build-standalone project

jepio:
# dev-lang/python-oem: use python from python-build-standalone project

Our python-oem is minimal and doesn't support pip or most of the standard
library. This is an attempt to see whether using a prebuilt python
distribution works to support Azure extensions.

It was necessary to add a symlink to 'python' executable and lib64 dir to avoid
having to change distro-oem and waagent ebuilds.

## How to use

Test on Azure, deploy VMBackup/VMSnapshot extension from Azure portal.

## Testing done


dev-lang/python-oem/python-oem-3.9.11.ebuild
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="This project produces self-contained, highly-portable Python distributions."
+HOMEPAGE="https://github.com/indygreg/python-build-standalone/"
+SRC_URI="
+	amd64? ( https://github.com/indygreg/python-build-standalone/releases/download/20220318/cpython-3.9.11+20220318-x86_64_v2-unknown-linux-gnu-install_only.tar.gz )
```
pothos, line 9: These are the binary releases, or? I think it's ok to use binary releases if we have some criteria but at the moment we normally compile from source (also possible for [this project](https://python-build-standalone.readthedocs.io/en/latest/building.html#linux)).
So that's more a general question worth discussing: under which circumstances are we fine consuming third-party binaries, and if it's an option, should we plan to do more of it to lower the maintenance burden and build times? E.g, there are binary releases for the Rust compiler or Go, why not use them, too?

jepio: Yes these are binary releases. Normally I prefer the build from source approach, but it wouldn't be possible to do that from the ebuild here:
> The execution environment must have access to a Docker daemon (all build operations are performed in Docker containers for isolation from the host system).
Though we could rebuild from source and host the binaries for ourselves, and then consume them.

> E.g, there are binary releases for the Rust compiler or Go, why not use them, too?
I would be for switching to dev-lang/rust-bin, the build times for it are ridiculous, and the rust compiler (including self bootstrap) is a major part of our SDK build time. For go we wouldn't have much benefit from using a pre-built toolchain.

pothos: Also given that we can revisit the choice of where the python binaries come from when migrating to sysext, I'm fine consuming the published binaries in this case here.

About Rust, yeah, specially the LLVM compilation takes ages…

jepio: Rust cross-compilation support is something we handle in dev-lang/rust and would need to be ported. I haven't looked deeper.

pothos: Yeah, or use `rustup` to install the regular toolchain into the SDK

dev-lang/python-oem/python-oem-3.9.11.ebuild
```diff
@@ -0,0 +1,32 @@
...
+DEPEND=""
+RDEPEND="${DEPEND}"
+BDEPEND=""
+
+S="${WORKDIR}"
+
+src_install() {
+	dodir /usr/share/oem
+	cp -a ${S}/python ${ED}/usr/share/oem
+	find ${ED} -name 'lib*.a' -delete
```
pothos, line 29:
```suggestion
	find ${ED}/usr/share/oem/python/ -name 'lib*.a' -delete
```

dev-lang/python-oem/python-oem-3.9.11.ebuild
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 Gentoo Authors
```
pothos, line 1: Since this is very different from `python-oem-3.9.8.ebuild` please remove `python-oem-3.9.8.ebuild` to avoid confusion

krnowak:
I'm not sure I like the fact that we are pulling binaries from some random place. How does this play with the secure supply chain initiative?

jepio:
That's why I wanted to have this discussion, this is not something I want to merge as is. At very least we should rebuild the binaries ourself.

Do you have ideas for alternative approaches that I could explore?

pothos:
> Do you have ideas for alternative approaches that I could explore?

Maybe we can prioritize the OEM sysext transition over Docker sysext and start with Azure?

PR #1819, opened 2022-04-13
Package updates, 2021 edition, part 1 - fallout fixes from the update of profiles in portage-stable

krnowak:
The update of the profiles in portage-stable dropped support for freebsd kernels, masked sys-libs/e2fsprogs-libs (in favor of a newer version of sys-fs/e2fsprogs package) and masked new versions of dev-vcs/git. This means that some use flags were dropped or packages got masked, so packages that were using them also needed an update.

Should be merged together with https://github.com/flatcar-linux/portage-stable/pull/323.

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5625/cldsv/

- sys-apps/dbus
  - 1.12.20-r2 -> 1.12.22-r1
  - new modification - disable user-sessions outright, earlier it was
    disabled by not enabling user-session use flag, but now it got
    merged into systemd use flag, which we enable. We didn't have user
    sessions before, so let's keep it like that. Enabling user
    sessions caused dead symlinks kola test failure.
- net-fs/samba
  - 4.12.9-r2 -> 4.15.4-r3
  - Flatcar modifications slightly reduced
    - we probably should still try to avoid using bundled ldb, I haven't tried that yet
- sys-libs/ldb
  - moved to portage-stable
- profiles and coreos/config:
  - clean ups of accept_keywords, mask, unmasks and stuff


dongsupark (approved):

profiles/coreos/targets/generic/make.defaults
```diff
@@ -76,6 +76,14 @@ INSTALL_MASK="${INSTALL_MASK}
   /usr/bin/cvtsudoers
 "
 
+# Do not install some of the tools from e2fsprogs.
+INSTALL_MSK="${INSTALL_MASK}
```
dongsupark, line 80: Do you mean `INSTALL_MASK`?

dongsupark (approved):
Looks good in general.

profiles/coreos/targets/generic/make.defaults
krnowak: Meh, yes. Thanks for spotting.

PR #1837, opened 2022-04-21
sys-apps/portage: refresh patches and ensure they are applied

jepio:
# sys-apps/portage: refresh patches and ensure they are applied

The patches have been quietly ignored since 2020, but we could really use at least the one that gets rid of
this message:

  !!! Section 'gentoo' in repos.conf is missing location attribute

## How to use

`sudo emerge -av portage`

## Testing done

`sudo emerge -av portage`


tormath1 (approved):
Thanks !

PR #1838, opened 2022-04-21
profiles: add keywords to cross-*/binutils as well

jepio:
# profiles: add keywords to cross-*/binutils as well

Crossdev currently uses binutils 2.36 (stable), while the SDK and sysroot both
build binutils 2.37 due to keywording. Kernel modules built within the
developer container fail to load due to relocation errors. Add the same
keywords to cross-*/binutils packages so that the versions match.

- [ ] https://github.com/flatcar-linux/scripts/pull/286

## How to use

Perform at least a stage4 bootstrap with sdk-new. Doesn't affect update_chroot because that prefers binpkgs in all cases.

## Testing done

Still testing in CI.

### DIFF
--- a/profiles/coreos/base/package.accept_keywords
+++ b/profiles/coreos/base/package.accept_keywords
@@ -30,6 +30,11 @@
 # Upgrade to GCC 10.3.0 to support latest glibc builds
 =sys-devel/binutils-2.37_p1 ~amd64 ~arm64
 =sys-libs/binutils-libs-2.37_p1 ~amd64 ~arm64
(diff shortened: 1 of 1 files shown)


jepio:
Testing here:
- all: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5475/cldsv/
- all-full: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5476/cldsv/

dongsupark (approved):

PR #1839, opened 2022-04-25
Add missing pahole dependencies

jepio:
# Add missing pahole dependencies

pahole is a build-time dependency of our kernel build, due to us setting
CONFIG_BTF_DEBUG_INFO. If pahole is missing, a `make modules_prepare` with our
kernel config results in symbols in the config changing. This will affect
people building kernel modules against coreos-sources in the developer
container, but not the SDK because pahole is already in sdk-depends.

## How to use

Test the resulting developer container with nvidia.service on a GPU instance (e.g. NC6s_v3 in EastUS).

## Testing done

Modified this line: https://github.com/flatcar-linux/coreos-overlay/blob/main/x11-drivers/nvidia-drivers/files/bin/install-nvidia#L7 to be `emerge -gkv pahole coreos-sources` and verified that this allows the built modules to be probed. Without this, the module fails to load with this error message "module: x86/modules: Skipping invalid relocation target, existing value is nonzero for type 1...".

Tested here: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5487/cldsv/

`

### DIFF
--- a/changelog/bugfixes/2022-04-25-fix-dev-container-module-builds.md
+++ b/changelog/bugfixes/2022-04-25-fix-dev-container-module-builds.md
@@ -1,0 +1,1 @@
+- Added pahole to developer container, without it kernel modules built against /usr/src/linux may fail to probe with an 'invalid relocation target' error ([PR#1839](https://github.com/flatcar-linux/coreos-overlay/pull/1839))
--- a/coreos-base/coreos-dev/coreos-dev-0.1.0.ebuild
+++ b/coreos-base/coreos-dev/coreos-dev-0.1.0.ebuild
@@ -21,6 +21,7 @@ RDEPEND="
 	coreos-base/emerge-gitclone
 	dev-lang/python
 	dev-util/strace
+	dev-util/pahole
 	dev-vcs/repo
 	net-analyzer/netperf
 	net-dialup/minicom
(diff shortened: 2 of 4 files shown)


pothos (approved):

pothos (commented):
A changelog entry for bugfix would be good, or?

jepio:
Can you check the changelog I added?

changelog/bugfixes/2022-04-25-fix-dev-container-module-builds.md
```diff
@@ -0,0 +1 @@
+- Added dev-util/pahole to developer container, without it kernel modules built against /usr/src/linux may fail to probe with an 'invalid relocation target' error ([PR#1839](https://github.com/flatcar-linux/coreos-overlay/pull/1839))
```
pothos, line 1:
```suggestion
- Added pahole to developer container, without it kernel modules built against /usr/src/linux may fail to probe with an 'invalid relocation target' error ([PR#1839](https://github.com/flatcar-linux/coreos-overlay/pull/1839))
```
I think the dev-util prefix confuses non-Gentoo users (it could be read as an alternative name for pahole)

jepio: ok, appreciate the feedback, removed the dev-util prefix.

PR #1840, opened 2022-04-25
nvidia driver build improvements

jepio:
# nvidia driver build improvements

This PR includes several enhancements to the setup-nvidia/install-nvidia scripts that we ship to build
Nvidia kernel modules on system. They include:
* faster decompression
* build against /lib/modules, not /usr/src/linux (skip emerging things)
* sysext compatible file layout (for potential future use)
* allow service to rerun after reboot and only load modules
* allow service to work after upgrades
* allow user to override driver version (TODO: add GPU docs flatcar-docs)

## How to use

Build image and test on Azure Standard_NC6 or Standard_NC6s_v3 instances in EastUS.

## Testing done

Tested according to "how-to-use", would appreciate testing on other platforms.

### DIFF
--- a/x11-drivers/nvidia-drivers/files/bin/install-nvidia
+++ b/x11-drivers/nvidia-drivers/files/bin/install-nvidia
@@ -13,5 +13,12 @@ cd "/nvidia/${NVIDIA_DRIVER_BASENAME}"
   --no-distro-scripts \
   --no-systemd \
   --skip-depmod \
-  --kernel-install-path="${PWD}" \
+  --kernel-install-path="${PWD}/install-mod" \
   --log-file-name="${PWD}/nvidia-installer.log" || true
+
+mkdir -p /lib/modules/${KERNEL_NAME}/video
+mkdir -p "${PWD}"/install-mod
+cp "${PWD}"/kernel/*.ko "${PWD}"/install-mod
+cp "${PWD}"/install-mod/*.ko /lib/modules/${KERNEL_NAME}/video/
+depmod -a
+cp /lib/modules/${KERNEL_NAME}/modules.* "${PWD}/install-mod/"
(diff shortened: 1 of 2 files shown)


x11-drivers/nvidia-drivers/files/bin/install-nvidia
```diff
@@ -2,17 +2,23 @@
...
   --no-distro-scripts \
-  --kernel-install-path="${PWD}" \
+  --no-systemd \
+  --skip-depmod \
+  --kernel-install-path="${PWD}/install-mod" \
   --log-file-name="${PWD}/nvidia-installer.log" || true
+
+mkdir -p /lib/modules/${KERNEL_NAME}/video
+mkdir -p "${PWD}"/install-mod
+cp "${PWD}"/kernel/*.ko "${PWD}"/install-mod
```
jepio, line 21: `nvidia-installer` fails when it can't modprobe immediately, which means it always fails in the dev container. Depending on installer version, the modules are copied to `kernel-install-path` anyway, or not. So we have to assume the module files are in `$PWD/kernel/*.ko` and fetch them from there.

jepio:
There is also a daemon that can be run `nvidia-persistenced`, that pins the GPU in the driver and makes one-off accesses (like nvidia-smi) much faster. I've toyed with 2 ideas: `nvidia-peristence.service`, or rework `nvidia.service` to effectively do:
```
[Service]
ExecStartPre=/usr/share/oem/bin/setup-nvidia
ExecStart=/opt/bin/nvidia-persistenced
TimeoutSec=10min
```
Does anyone have any suggestions about this?

pothos:
> Does anyone have any suggestions about this?

In the docs I didn't see power usage mentioned, would be interesting to know if this is a tradeoff like setting the `performance` CPU governor which has performance/latency benefits but will consume more power.

PR #1842, opened 2022-04-26
sys-apps/ignition: use upstream

tormath1:
use upstream ignition (coreos/ignition) and apply our patches on top of
it.

It's currently done in the same way with coreos/afterburn.

<hr>

It basically changes nothing to the Ignition we currently ship in beta and alpha except that we decouple the ebuild from the flatcar-linux/ignition fork.
In the end, this fork should be archived.

It's far from being flexible - but at least, we now carry an explicit list of patches. To upgrade Ignition, we could checkout on the upstream git repo to the targeted version, apply and adapt the patches and commit them.

## Testing done

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5511/cldsv/
- locally `ebuild ignition-2.13.0-r1.ebuild clean configure`
- changelog is not required


tormath1:
Yes `d5545707b879e507749efed89881f987bd0ec81e` is the most recent commit we have in common with the upstream - I simply run `git format-patch d5545707b879e507749efed89881f987bd0ec81e` from our fork - since all patches are currently applied on top of this commit.

The idea is to go in two steps:
1. Switch the ebuild to use coreos/ignition + apply patches during the `src_configure` step
2. Wait for the next Ignition release to include IMDS and others.

pothos (approved):

PR #1863, opened 2022-05-06
sys-apps/baselayout: Bump package version to force rebuild

pothos:
The SDK container includes pre-build boards which have the basic
packages installed in the versions when the SDK was built. However,
when a change was done to one of these packages, like "baselayout",
this change is ignored unless a package version bump was done.

Bump the package version to enforce the rebuild when using the SDK
container.

## Testing done

[ongoing](http://192.168.42.7:8080/job/container/job/packages/254/cldsv/)


pothos:
Doesn't work yet because the baselayout package can't be rebuilt as it wants to overwrite `/etc/hosts` which in the SDK container is a bind mount.

PR #1883, opened 2022-05-12
Enable arm64 version of sssd

mtougeron:
# Enable arm64 version of sssd

Enable arm64 version of sssd

## How to use

Verify that sssd is cross-compiled for arm64

This may be dependent on the samba changes in https://github.com/flatcar-linux/coreos-overlay/pull/1819

I'm following the guidance from https://github.com/flatcar-linux/Flatcar/issues/689#issuecomment-1124969883

## Testing done

### DIFF
--- a/coreos-base/coreos/coreos-0.0.1.ebuild
+++ b/coreos-base/coreos/coreos-0.0.1.ebuild
@@ -80,7 +80,6 @@ RDEPEND="${RDEPEND}
 		app-emulation/xenstore
 		net-fs/cifs-utils
 		sys-auth/realmd
-		sys-auth/sssd
 	)"
 
 RDEPEND="${RDEPEND}
@@ -170,6 +169,7 @@ RDEPEND="${RDEPEND}
 	sys-apps/usbutils
 	sys-apps/util-linux
 	sys-apps/which
+	sys-auth/sssd
 	sys-block/open-iscsi
 	sys-block/parted
 	sys-cluster/ipvsadm


coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -80,7 +80,6 @@ RDEPEND="${RDEPEND}
 		app-emulation/xenstore
 		net-fs/cifs-utils
 		sys-auth/realmd
-		sys-auth/sssd
```
jepio, line 83: By removing this item from the list, it won't be installed into any image. Instead you would need to move the entry into the list starting at 86 (and let's keep it sorted)

mtougeron: Updated. Thanks for your patience as I learn. :)

dongsupark:
Thanks for creating the PR.

However, it looks like the PR needs more work.

1. First of all, `dev-libs/ding-libs` is missing for arm64.

```
$ emerge-arm64-usr -pv sys-auth/sssd
!!! All ebuilds that could satisfy ">=dev-libs/ding-libs-0.2" for /build/arm64-usr/ have been masked.
- dev-libs/ding-libs-0.6.1-r1::portage-stable (masked by: ~arm64 keyword)
- dev-libs/ding-libs-0.6.1::portage-stable (masked by: ~arm64 keyword)

```

That happens because every version of ding-libs is still unstable for arm64.
You would probably want to add `=dev-libs/ding-libs-0.6.1-r1 ~arm64` to `profiles/coreos/arm64/package.accept_keywords`.


2. After that, you would still see failures like:

```
$ emerge-arm64-usr -pv sys-auth/sssd
!!! All ebuilds that could satisfy ">=sec-policy/selinux-sssd-2.20120725-r9" for /build/arm64-usr/ have been masked.
- sec-policy/selinux-sssd-9999::portage-stable (masked by: missing keyword)
- sec-policy/selinux-sssd-2.20210203-r1::portage-stable (masked by: ~arm64 keyword)
- sec-policy/selinux-sssd-2.20200818-r2::portage-stable (masked by: ~arm64 keyword)

```

That happens because every version of selinux-sssd is still unstable for arm64.
You would probably want to add `=sec-policy/selinux-sssd-2.20200818-r2 ~arm64` to `profiles/coreos/arm64/package.accept_keywords`.

Then the actual build could probably start pulling in a bunch of additional dependencies.
You would again bump into additional issues during the builds, because of hidden build issue being uncovered by that change.

mtougeron:
Can you point me in the right direction to find the tools you're using to build & test these changes? I'm kind of flying blind here. I'd like to be able to help out and do this change but honestly I'm not quite sure what I'm doing. :( I'm having a hard time finding the docs on the testing process for Flatcar that I can run myself before pushing changes to the PR.

mtougeron:
> I would suggest following the developer guide: https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-modifying-flatcar/

Thank you, I will read through this and give it a try.

PR #1899, opened 2022-05-18
sys-libs/ncurses: update to 6.3_p20220423

dongsupark:
Update `sys-libs/ncurses` to 6.3_p20220423, mainly to address [CVE-2022-29458](https://nvd.nist.gov/vuln/detail/CVE-2022-29458).

Based on Gentoo commit [a59133147f7a](https://github.com/gentoo/gentoo/commit/a59133147f7a).

Flatcar change is based on [09951dc3db0f](https://github.com/flatcar-linux/coreos-overlay/commit/09951dc3db0f).

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5656/cldsv/


sys-libs/ncurses/ncurses-6.3_p20220423.ebuild
```diff
@@ -99,7 +163,7 @@ src_configure() {
 		CXXFLAGS=${BUILD_CXXFLAGS} \
 		CPPFLAGS=${BUILD_CPPFLAGS} \
 		LDFLAGS="${BUILD_LDFLAGS} ${lbuildflags}" \
-		do_configure cross --without-shared --with-normal
+		do_configure cross --without-shared --with-normal --with-progs
```
krnowak, line 166: Does this install some extra stuff we didn't need before?

dongsupark: No, ncurses 6.3 does not install additional binaries or tools under /usr/bin etc.
Both 6.2 and 6.3 install binaries like `clear` or `tic`.
I think `--with-progs` has been always the default.

krnowak: Ok, thanks for the explanation.

PR #1900, opened 2022-05-18
Package updates, 2021 edition, part 2 - python update

krnowak:
Contents of this PR:
- Sync dev-lang/python-oem to 3.9.12 (so I haven't done the switch to 3.10 yet, even if it's stable now in Gentoo).
- Bump dev-python/distro-oem package to 1.7.0.
- Sync dev-python/boto.
- Pull in some more python packages to SDK, due to a big overhaul in a way the python packages are being built.
- Fix python dependency in selinux package
- Clean up some use flags (dropped the enabled-by-default or obsolete flags).

Should be merged together with https://github.com/flatcar-linux/portage-stable/pull/329.

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5711/cldsv/


coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
```diff
@@ -50,8 +50,13 @@ RDEPEND="${RDEPEND}
 	dev-libs/libyaml
 	dev-libs/nspr
 	dev-libs/protobuf
+	dev-python/flit_core
+	dev-python/gpep517
 	dev-python/mako
+	dev-python/poetry-core
 	dev-python/setuptools
+	dev-python/tomli
```
krnowak, line 58: I'll drop this line - it's a leftover. It's probably going to be pulled into SDK anyway, but there's no need for it to be an explicit dependency.

dongsupark (approved):
Looks good.
Please squash unnecessary commits before merging.

PR #1916, opened 2022-05-25
sys-apps/ignition: upgrade to 2.14.0

tormath1:
Ignition has released a new version, CHANGELOG is available here: https://github.com/coreos/ignition/releases/tag/v2.14.0

Here some high level changes for the patches:
* `ign-converter` is directly pulled to its latest version: 0.2.0 - so we can drop this two patches: https://github.com/flatcar-linux/coreos-overlay/blob/b9235d9188b489b6484792c53851febdd3cc71c5/sys-apps/ignition/ignition-9999.ebuild#L60-L61
* This function now takes a `io.Writer` as argument instead of `*os.File`: https://github.com/flatcar-linux/coreos-overlay/blob/b9235d9188b489b6484792c53851febdd3cc71c5/sys-apps/ignition/files/0013-Revert-drop-OEM-URI-support.patch#L179
* The [`Revert-drop-OEM-URI-support.patch`](https://github.com/flatcar-linux/coreos-overlay/blob/main/sys-apps/ignition/files/0013-Revert-drop-OEM-URI-support.patch) and [`0010-VMware-Fix-guestinfo.-.config.data-and-.config.url-v.patch`](https://github.com/flatcar-linux/coreos-overlay/blob/main/sys-apps/ignition/files/0010-VMware-Fix-guestinfo.-.config.data-and-.config.url-v.patch) were the most painful to port.

Related-To: https://github.com/flatcar-linux/Flatcar/issues/656

<hr>

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5773/cldsv/


sys-apps/ignition/files/0005-config-v3_4-convert-ignition-2.x-to-3.4-exp.patch
```diff
@@ -0,0 +1,81 @@
+From 85d70f5dab2014859a1248ced91271d67a813714 Mon Sep 17 00:00:00 2001
+From: Mathieu Tortuyaux <mtortuyaux@microsoft.com>
+Date: Wed, 25 May 2022 10:20:09 +0200
+Subject: [PATCH 05/15] config/v3_4: convert ignition 2.x to 3.4-exp
+
+if the version of ignition is 2.x we convert it to 3.4-exp using
+ign-converter.
+
+it shoud support any 2.x version (or at the least the last 2 releases)
```
krnowak, line 9:
```suggestion
it should support any 2.x version (or at the least the last 2 releases)
```

sys-apps/ignition/files/0005-config-v3_4-convert-ignition-2.x-to-3.4-exp.patch
```diff
@@ -0,0 +1,81 @@
...
++			map[string]string{} is required by the ign-converter
++			Ignition Spec 3 will mount filesystems at the mountpoint specified by path when running.
++			Filesystems no longer have the name field and files, links, and directories no longer specify the filesystem by name.
++			This means to translate filesystems (with the exception of root),
++			you must also provide a mapping of filesystem name to absolute path, e.g.
++			```
++			map[string]string{"var": "/var"}
++			```
++		*/
++		newCfg, err := v24tov31.Translate(cfg, map[string]string{})
```
krnowak, line 63: This comment is confusing me, because it talks about why the map is required, only to pass an empty one. Who should be responsible for filling the map? Is passing an empty map going to make the conversion to fail, because some required entry was missing from the map?

sys-apps/ignition/files/0006-internal-prv-cmdline-backport-flatcar-patch.patch
```diff
@@ -0,0 +1,40 @@
+From ca0c60b6dfbec852360abae7bed18f9e91883e5a Mon Sep 17 00:00:00 2001
+From: Mathieu Tortuyaux <mathieu@kinvolk.io>
+Date: Fri, 30 Jul 2021 14:21:31 +0200
+Subject: [PATCH 06/15] internal/prv/cmdline: backport flatcar patch
+
+this patch provides backward compatiblity for various config
```
krnowak, line 6:
```suggestion
this patch provides backward compatibility for various config
```

sys-apps/ignition/files/0010-VMware-Fix-guestinfo.-.config.data-and-.config.url-v.patch
```diff
@@ -1,7 +1,7 @@
-From b4e062a912db68e50f5db35bdb7704889a5aedc8 Mon Sep 17 00:00:00 2001
-From: =?UTF-8?q?Kai=20L=C3=BCke?= <kai@kinvolk.io>
-Date: Tue, 7 Apr 2020 17:36:35 +0200
-Subject: [PATCH 10/16] VMware: Fix guestinfo.*.config.data and *.config.url
+From d67ef76dcec2dafeba89a424c25c7f3f6e765bfa Mon Sep 17 00:00:00 2001
+From: Mathieu Tortuyaux <mtortuyaux@microsoft.com>
+Date: Wed, 25 May 2022 10:34:32 +0200
+Subject: [PATCH 10/15] VMware: Fix guestinfo.*.config.data and *.config.url
```
krnowak, line 4: Did you change the authorship here by mistake?

sys-apps/ignition/ignition-9999.ebuild
```diff
@@ -12,7 +12,7 @@ inherit coreos-go cros-workon systemd udev
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm64"
 else
-	CROS_WORKON_COMMIT="d5545707b879e507749efed89881f987bd0ec81e" # main
+	CROS_WORKON_COMMIT="fef845a7d5ba463b4a93a5ed03a76721164c4a84" # v2.14.0
```
krnowak, line 15: While at it - the `COREOS_GO_PACKAGE` variable above says `github.com/flatcar-linux/ignition/v2`. Should it say `coreos` rather?

Not sure that it matter, though - building the package worked so far anyway. But we should consider updating the ebuild to stop using `cros-workon` and `coreos-go` eclasses, I think. But not in this PR. Let's keep the scope of the PR limited to ignition update.

sys-apps/ignition/files/0005-config-v3_4-convert-ignition-2.x-to-3.4-exp.patch
tormath1: I get your point - this is required by `ign-converter` in the `Translate` method. Later in the code, the map is feed, for example `"root": "/"` is automatically added.
We also added some logic to generate values see this commit: https://github.com/flatcar-linux/ign-converter/commit/96f82d676ffd35e6dbf68078455ff02affd7b8b1.
If you want more information on that, there is the upstream issue: https://github.com/coreos/ign-converter/issues/30.

When `ign-converter` is used via its CLI, `fsMap` is manually provided by the user.

sys-apps/ignition/files/0010-VMware-Fix-guestinfo.-.config.data-and-.config.url-v.patch
krnowak: Oh yeah, applying failing patches is a PITA. Anyway, I remember seeing `Co-authored-by: Name Surname <email@stuff.com>` label. I think Github even understands it. It's used in the same way as `Signed-off-by` label.

pothos: With `git am -3` it's sometimes possible to resolve the conflicts but at some point I also find it easier to redo the patch manually.

tormath1: yeah - that's why this patch is mentioned in the PR description but only 2 on the whole set of patches were annoying to apply, otherwise it went well. :)

pothos:
Looks good! I think it makes sense to test all OEM platforms for this change before merging.

krnowak (approved):

pothos:
If you want, you can squash the commits because the changes belong together

PR #1948, opened 2022-06-09
sys-apps/ignition: add `ignition-rmcfg`

tormath1:
In this PR, we add the `ignition-rmcfg` command (in the root filesystem, not in the initramfs) to remove Ignition configuration from booted instance on VMWare and Virtualbox.

See also: https://github.com/advisories/GHSA-hj57-j5cw-2mwp

<hr>

We could add a Mantle test to verify Ignition has been correctly removed from VMWare guestinfo.

- [ ] CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5825/cldsv/


sys-apps/ignition/ignition-9999.ebuild
```diff
@@ -67,4 +67,8 @@ src_compile() {
 
 src_install() {
 	newbin ${GOBIN}/internal ${PN}
+
+	systemd_dounit "${FILESDIR}/ignition-delete-config.service"
```
pothos, line 71: This file should be part of the regular system, not of the initramfs, or? If it doesn't work to have it added here to the regular system, we could add it in the init repo (otherwise if it's for the initramfs the bootengine repo would fit better).

tormath1: Turns out that it's less trivial than expected - I added the service file to the init repository (https://github.com/flatcar-linux/init/pull/73).

Now comes the fun part:
* `/usr/bin/ignition`: has to be in the initramfs but not in the regular system
* `/usr/libexec/ignition-rmcfg`: has to be in the regular system 

Adding `sys-apps/ignition` to `coreos-base/coreos` dependency will work but we will have `/usr/bin/ignition` in the regular system.

Adding a use-flag like `remove` to have the following dependency `sys-apps/ignition[remove]` won't work because we will have a conflict between `coreos-kernel` and `coreos-base` (since both are installed at the same moment).

pothos: So, a second package or maybe move this to the initramfs and only use it when a ignition config is used because with cloud-init we wouldn't delete the userdata anyway?

tormath1: I wanted to avoid a second package - since `ignition-rmcfg` _is_ `ignition` but with a different name -> two ebuilds to maintain but I don't see yet other solutions.
Having `ignition-rmcfg` in the initramfs could work but one won't be able to mask the service to prevent the config from being removed.

sys-apps/ignition/ignition-9999.ebuild
```diff
@@ -67,4 +67,7 @@ src_compile() {
 
 src_install() {
 	newbin ${GOBIN}/internal ${PN}
+
+	exeinto "/usr/libexec"
+	newexe ${GOBIN}/internal "${PN}-rmcfg"
```
pothos, line 72: Does this file end up in the initramfs? We need to be careful because the arm64 initramfs almost hits the size limit.

changelog/security/2022-06-09-ignition.md
```diff
@@ -0,0 +1 @@
+- ignition ([CVE-2022-1706](https://nvd.nist.gov/vuln/detail/CVE-2022-1706))
```
pothos, line 1: Maybe a change entry would be good in addition?

sys-apps/ignition/ignition-9999.ebuild
tormath1: Yeah it does not. The initramfs is being built in the `sys-kernel/coreos-kernel` emerging, that's why we have a dependency for Ignition in the ebuild: https://github.com/flatcar-linux/coreos-overlay/blob/927d1afcc9feb157fd8c005a3d22b9fb3c9121a9/sys-kernel/coreos-kernel/coreos-kernel-5.15.44.ebuild#L20

The Ignition dracut's module setup explicitly installs the `ignition` binary: https://github.com/flatcar-linux/bootengine/blob/39caefd1da35d3600120b44ea5ee23ed22129b46/dracut/30ignition/module-setup.sh#L77-L78 so we're safe - `ignition-rmcfg` will never land in the initramfs as long as we don't explicitly pulling it.

changelog/security/2022-06-09-ignition.md
pothos: The change to enable it is done in this PR, so I think the operator notes link is very valuable here in case someone would have to opt out

tormath1: Ok, done. Thanks for the suggestion :)

changelog/changes/2022-06-14-ignition-rmcfg-vmware.md
```diff
@@ -0,0 +1 @@
+- VMWare: Added `ignition-delete-config.service` to remove Ignition config from VM metadata. Also see: https://coreos.github.io/ignition/operator-notes/#automatic-config-deletion
```
pothos, line 1:
```suggestion
- VMWare: Added `ignition-delete-config.service` to remove Ignition config from VM metadata, see also [here](https://coreos.github.io/ignition/operator-notes/#automatic-config-deletion)
```
Without the markdown formatting I'm not sure if it becomes a clickable link in the homepage?

pothos (approved):

PR #1993, opened 2022-06-28
wip: sec-policy/*: sync with gentoo

tormath1:
TBD

## How to use

TBD

## Testing done

TBD


JAORMX (approved):

krnowak:
I would only want to ask you to keep only one ebuild per package, so it will be clear which ebuild is actually patched up by us. Thanks!

krnowak:
> > I would only want to ask you to keep only one ebuild per package, so it will be clear which ebuild is actually patched up by us. Thanks!
> 
> I did or I missed one? If you just checked the Manifest that's normal, it's added in the TODO to regen the manifest otherwise it's in each "flatcar patches" commit (`removed non-used ebuilds`)

Ah, sorry. I just remembered that we used to have more ebuilds in those packages and Manifest files confused me, apparently. :) Thanks for taking care of it.

sys-libs/libsemanage/files/tmpfiles.d/libsemanage.conf
```diff
@@ -1,3 +1,4 @@
 #Type	Path				Mode	UID 	GID 	Age	Argument
 d	/etc/selinux/		-	-	-	-	-
 L	/etc/selinux/semanage.conf	-	-	-	-	../../usr/lib/selinux/semanage.conf
+z	/etc/selinux/semanage.conf
```
JAORMX, line 4: should we instead do:
```
Z	/etc/selinux/
```
?

JAORMX (changes requested):

PR #1957, opened 2022-06-14
sys-libs/glibc: update to 2.34

dongsupark:
Sync with Gentoo to update `sys-libs/glibc` to 2.34-r13, based on Gentoo commit [01ad1fecda61](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=01ad1fecda61778081c5c1eaa051d7f6f2f4456c).

To do that, we need to do the following as well:
* move `app-admin/adcli` to `app-crypt/adcli` 0.9.1 in portage-stable, as there is no need to keep Flatcar patches.
* update `sys-devel/gdb` to 11.2, as gdb 9 is not compatible with glibc 2.34.
  * Side issue: a fix for another cross build issue uncovered in gdb 11.2. (See https://github.com/gentoo/gentoo/pull/25896 for details)

This PR is still blocked by another issue with `cl.update.payload` failing always. Need a deeper look.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5852/cldsv/ (failed)

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)

PR #2001, opened 2022-07-01
preparation for glibc 2.34

dongsupark:
For updating glibc to [2.34](https://github.com/flatcar-linux/coreos-overlay/pull/1957), we need to do first do the followings:
* move `app-admin/adcli` to `app-crypt/adcli` 0.9.1 in portage-stable, as there is no need to keep Flatcar patches.
* update `sys-devel/gdb` to 11.2, as gdb 9 is not compatible with glibc 2.34.
  * Side issue: a fix for another cross build issue uncovered in gdb 11.2. (See https://github.com/gentoo/gentoo/pull/25896 for details. As Gentoo did not fix 11.2, but only >= 12.1, we still need to keep gdb in coreos-overlay).

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/340.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5976/cldsv/


krnowak (approved):
Looks good, but the changelog commit message should mention gdb instead of glibc.

dongsupark:
Fixed the commit message. Thanks.

PR #2026, opened 2022-07-13
sys-kernel/bootengine: Fix ignition kargs support by creating grub.cfg

pothos:
This pulls in
https://github.com/flatcar-linux/bootengine/pull/47
which creates the grub.cfg file if it does not exist when the Ignition
kargs directive is used, preventing an error when it tried to read the
current settings from it.

## How to use

Can be backported when using bootengine branches

## Testing done

see PR


pothos:
Backported to 3227 by using a new bootengine branch `flatcar-3227-backport`

PR #2027, opened 2022-07-13
Record SLSA provenance data

jepio:
# Record SLSA provenance data

Add post install hooks to capture provenance metadata for every built package. The hooks are sourced by default by every package built with the coreos/base profile (or inheriting from it), but enabled only when GENERATE_SLSA_PROVENANCE=true is defined. This is intended to be enabled for board packages (including toolchains).

The hooks need access to some metadata beyond what is available in the temporary build directory: hashes of commits of scripts/coreos-overlay/portage-stable and sdk version currently in use.

## How to use

Together with https://github.com/flatcar-linux/scripts/pull/378, build a full image including toolchains.

## Testing done

http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/6068/cldsv/

### DIFF
--- a/profiles/coreos/base/profile.bashrc.slsa-provenance
+++ b/profiles/coreos/base/profile.bashrc.slsa-provenance
@@ -1,0 +1,181 @@
+# Build provenance hooks
+# ======================
+# The functions below hook into every ebuild's execution and generate provenance files
+#   to meet the SLSA provenance requirements (https://slsa.dev/spec/v0.1/requirements#available).
+# All input files (source tarball / git commit hash) plus added patches / files,
+#   and all resulting installation binaries and files are captured.
+# The information is emitted in SLSA provenance 0.2 format (see https://slsa.dev/provenance/v0.2)
+
+
+# We only record provenance when a package is actually being built.
+# See profiles/coreos/base/profile.bashrc for cros_...
+cros_post_src_configure_enable_slsa_provenance_report() {
+    export __slsa_provenance_report="yes"
+}
+# --
+
+# Generate SLSA provenance 0.2 Subject information.
+# The information will cover all installation files shipped with a package.
+__slsa_provenance_subject() {
+    local parallel="$(grep -cE '^processor' /proc/cpuinfo)"
+    local comma=""
+
+    einfo "    Provenance: recording subject (output)"
+
+    echo '   "subject": ['
+
+    (
+        cd "$D"
+        find . -type f -print | sed 's:^./::' | xargs -P "$parallel" -L 1 sha512sum
+    ) | while read checksum filepath; do
+        echo -en "${comma}     {\"name\":\"/$filepath\", \"digest\":{\"sha512\":\"$checksum\"}}"
+        if [ -z "$comma" ] ; then
+            comma=',\n'
+        fi
+    done
+   echo -en "\n   ]"
+}
+# --
+
+__slsa_provenance_materials() {
+    local csum="" uri="" repo="" ebuild=""
+
+    local ebuilds_basedir="/mnt/host/source/src/third_party/"
+    local ebuild="${CATEGORY}/${PN}/${PF}.ebuild"
+    if [ -f "${ebuilds_basedir}/coreos-overlay/${ebuild}" ] ; then
+        repo="coreos-overlay"
+    elif [ -f "${ebuilds_basedir}/portage-stable/${ebuild}" ] ; then
+        repo="portage-stable"
+    else
+        eerror "SLSA provenance: Unable to detect ebuild repository for package '${ebuild}'"
+        exit 1
+    fi
+
+    einfo "    Provenance: recording ebuild material (input) '${repo}/${ebuild}'"
+    echo '   "materials": ['
+
+    # The ebuild. Since "configSource" in "invocation" cannot have more than one (top/level) entry
+    #  we add the ebuild and git repo checksum here, as a material.
+    csum="$(cat "/mnt/host/source/src/scripts/.git/modules/sdk_container/src/third_party/${repo}/ORIG_HEAD")"
+    uri="git+https://github.com/flatcar-linux/${repo}.git/${ebuild}"
+    echo -e "      { \"uri\": \"${uri}\","
+    echo -n "        \"digest\": {\"sha1\":\"${csum}\"} }"
+
+    # The main sources
+    if [ -n "${A}" ] ; then
+        # Package is built from downloaded source tarball(s)
+        # There can be multiple, and can be used conditionally based on use flags,
+        #  and even replaced with different local names ("http://... -> othername.tgz"). So
+        #  we go through what's actually used ($A), then find the corresponding source URI.
+        local src="" prev_uri="" rename="false" orig_name=""
+        for src in ${A}; do
+            local found="false"
+            for uri in ${SRC_URI}; do
+                if [ "${uri}" = "->" ] ; then
+                    rename="true"
+                    continue
+                fi
+                if [ "${src}" = "$(basename "${uri}")" ] ; then
+                    orig_name="${src}"
+                    if [ "${rename}" = "true" ] ; then
+                        uri="${prev_uri}"
+                        orig_name="$(basename "${uri}")"
+                    fi
+                    einfo "    Provenance: recording tarball material (input) '${src}' ('${orig_name}')"
+                    csum="$(sha512sum "${DISTDIR}/${src}" | cut -d' ' -f1)"
+                    echo -e ",\n      { \"uri\": \"${uri}\","
+                    echo -n "        \"digest\": {\"sha512\":\"${csum}\"} }"
+                    found="true"
+                fi
+                rename="false"
+                prev_uri="${uri}"
+            done
+            if [ "${found}" != "true" ] ; then
+                eerror "No SRC_URI found for source '${src}', unable to record provenance!"
+                exit 1
+            fi
+        done
+    elif [ -n "${EGIT_REPO_URI:-}" ] ; then
+        # package is built from repo checkout (git)
+        einfo "    Provenance: recording GIT material (input) '${EGIT_REPO_URI}'"
+        csum="${EGIT_COMMIT}"
+        uri="${EGIT_REPO_URI}"
+        echo -e ",\n      { \"uri\": \"${uri}\","
+        echo -n "        \"digest\": {\"sha1\":\"$csum\"} }"
+    fi
+
+    # Patches / files shipped with the ebuild (if any)
+    uri="git+https://github.com/flatcar-linux/${repo}.git/${CATEGORY}/${PN}/files"
+    csum="$(cat "/mnt/host/source/src/scripts/.git/modules/sdk_container/src/third_party/${repo}/ORIG_HEAD")"
+    if [ -d "${FILESDIR}" ] ; then
+        for file in $(cd "$FILESDIR" && find . -type f | sed 's:^./::') ; do
+            einfo "    Provenance: recording ebuild material (input) '${file}'"
+            echo -e ",\n      { \"uri\": \"${uri}/${file}\","
+            echo -n "        \"digest\": {\"sha1\":\"$csum\"} }"
+        done
+    fi
+
+    echo -ne '\n   ]'
+}
+# --
+
+__slsa_provenance_report() {
+    local scripts_hash="$(cat "/mnt/host/source/src/scripts/.git/ORIG_HEAD")"
+    local buildcmd="emerge"
+    # extract board from e.g. '/build/amd64-usr/build'. Empty if no board is set (SDK build).
+    local board="$(echo "${CROS_BUILD_BOARD_TREE:-}" | sed -n 's:^/build/\([^/]\+\)/.*:\1:p')"
+    if [ -n "$board" ] ; then
+        buildcmd="emerge-${board}"
+    fi
+
+    # FIXME: Supply SDK image ID and sha256 digest along with the version tag
+    local sdk_version="$(source /mnt/host/source/.repo/manifests/version.txt; echo ${FLATCAR_SDK_VERSION})"
+
+    # FIXME: add builder ID
+cat <<EOF
+{
+ "_type": "https://in-toto.io/Statement/v0.1",
+ "predicateType": "https://slsa.dev/provenance/v0.2",
+ "predicate": {
+   "buildType": "ghcr.io/flatcar-linux/flatcar-sdk-all:${sdk_version}",
+   "builder": {"id": "TODO - builder ID" },
+   "invocation": {
+      "configSource": {
+        "uri": "https://github.com/flatcar-linux/scripts",
+        "digest": {"sha1": "${scripts_hash}"}
+      }
+   },
+   "buildConfig": {
+     "commands": [
+        "git checkout ${scripts_hash}",
+        "git submodule init",
+        "git submodule update",
+        "./run_sdk_container ${buildcmd} =${CATEGORY}/${PF}"
+    ]
+   },
+EOF
+    __slsa_provenance_materials
+    echo ","
+    __slsa_provenance_subject
+    echo ""
+cat <<EOF
+  }
+}
+EOF
+}
+# --
+
+cros_post_src_install_generate_slsa_provenance_report() {
+    if [ "${__slsa_provenance_report:-no}" != "yes" ] ; then
+        return
+    fi
+
+    mkdir -p "$D/usr/share/SLSA/"
+    local report_file="${CATEGORY}_${PF}.json.xz"
+    local dest_dir="${D}/usr/share/SLSA/"
+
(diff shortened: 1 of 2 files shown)


profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,183 @@
...
+                    orig_name="${src}"
+                    if [ "${rename}" = "true" ] ; then
+                        uri="${prev_uri}"
+                        orig_name="$(basename "${uri}")"
+                    fi
+                    einfo "    Provenance: recording tarball material (input) '${src}' ('${orig_name}')"
+                    csum="$(sha512sum "${DISTDIR}/${src}" | cut -d' ' -f1)"
+                    echo -e ",\n      { \"uri\": \"${uri}\","
+                    echo -n "        \"digest\": {\"sha512\":\"${csum}\"} }"
+                    found="true"
```
krnowak, line 90: Should `break` come here too?
```suggestion
                    found="true"
                    break
```

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,183 @@
...
+                fi
+                rename="false"
+                prev_uri="${uri}"
+            done
+            if [ "${found}" != "true" ] ; then
+                eerror "No SRC_URI found for source '${src}', unable to record provenance!"
+                exit 1
```
jepio, line 97: Is `exit` valid here? I don't know how the emerge phase loop operates - will this abort the build?

profiles/coreos/base/profile.bashrc.slsa-provenance
krnowak: For a package, yes, but can SRC_URI have multiple entries for the same file? Like

```
SRC_URI="https://example.com/file.tar.gz https://another-example.com/file.tar.gz"
```

Do we want to capture both links?

profiles/coreos/base/profile.bashrc.slsa-provenance
krnowak: That's a good question. Emerge is a python script, that invokes a bash script (`bin/ebuild.sh` in portage) that ends up invoking the hooks. So I'd expect emerge to actually register the exit status of the bash script and act accordingly.

Maybe just use `die` instead of `eerror` and `exit`?

profiles/coreos/base/profile.bashrc.slsa-provenance
pothos: In the license json file all are stored because some may be dead links - I think we can also point to https://mirror.release.flatcar-linux.net/ because most tar balls are available there and actually downloaded from there and not from the SRC_URI.

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: I think you're right. Replaced it with a `die`.

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+cros_post_src_install_generate_slsa_provenance_report() {
+    if [ "${__slsa_provenance_report:-no}" != "yes" ] ; then
+        return
+    fi
+
+    mkdir -p "$D/usr/share/SLSA/"
+    local report_file="${CATEGORY}_${PF}.json.xz"
+    local dest_dir="${D}/usr/share/SLSA/"
+
+    __slsa_provenance_report | xz -9cz > "${T}/${report_file}"
```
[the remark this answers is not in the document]
pothos: I suggest sticking to bz2

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+    local parallel="$(grep -cE '^processor' /proc/cpuinfo)"
+    local comma=""
+
+    einfo "    Provenance: recording subject (output)"
+
+    echo '   "subject": ['
+
+    (
+        cd "$D"
+        find . -type f -print | sed 's:^./::' | xargs -P "$parallel" -L 1 sha512sum
```
krnowak, line 33: Hm, this might hurt the reproducibility of a build, because the generated lines may be in different order on every run. Not sure if it's important.

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+
+    einfo "    Provenance: recording ebuild material (input) '${repo}/${ebuild}'"
+    echo '   "materials": ['
+
+    # The ebuild. Since "configSource" in "invocation" cannot have more than one (top/level) entry
+    #  we add the ebuild and git repo checksum here, as a material.
+    csum="$(cat "/mnt/host/source/src/scripts/.git/modules/sdk_container/src/third_party/${repo}/ORIG_HEAD")"
+    uri="git+https://github.com/flatcar-linux/${repo}.git/${ebuild}"
+    echo -e "      { \"uri\": \"${uri}\","
+    echo -n "        \"digest\": {\"sha1\":\"${csum}\"} }"
```
krnowak, line 64: Couple points:

- I think URI should contain the specific commit this ebuild was taken from, so for instance, it should be in form like `https://raw.githubusercontent.com/flatcar-linux/coreos-overlay/3829e4d0dd8adb5c2af47f543723c3a525b5e743/sys-apps/systemd/systemd-250.3.ebuild`
- ORIG_HEAD contains a previous state of HEAD. It should use HEAD. Also, ORIG_HEAD does not even exist on first update of submodules (I don't have it here).
- It's confusing to record commit checksum of the repo and use URI to an ebuild. Shouldn't it be rather checksum of the ebuild file?

```suggestion
    # The ebuild. Commit checksum is contained within an URI to the ebuild.
    csum="$(cat "/mnt/host/source/src/scripts/.git/modules/sdk_container/src/third_party/${repo}/HEAD")"
    uri="https://raw.githubusercontent.com/flatcar-linux/${repo}/${csum}/${ebuild}"
    # I don't know where exactly we can find the ebuild file in this context.
    esum=$(sha512sum "${PATH_TO_THE_CURRENT_EBUILD_FILE}" | cut -f1 -d' ')
    echo -e "      { \"uri\": \"${uri}\","
    echo -n "        \"digest\": {\"sha512\":\"${esum}\"} }"
```

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+        einfo "    Provenance: recording GIT material (input) '${EGIT_REPO_URI}'"
+        csum="${EGIT_COMMIT}"
+        uri="${EGIT_REPO_URI}"
+        echo -e ",\n      { \"uri\": \"${uri}\","
+        echo -n "        \"digest\": {\"sha1\":\"$csum\"} }"
+    fi
+
+    # Patches / files shipped with the ebuild (if any)
+    uri="git+https://github.com/flatcar-linux/${repo}.git/${CATEGORY}/${PN}/files"
+    csum="$(cat "/mnt/host/source/src/scripts/.git/modules/sdk_container/src/third_party/${repo}/ORIG_HEAD")"
```
krnowak, line 110: Same as above.

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+            echo -n "        \"digest\": {\"sha1\":\"$csum\"} }"
+        done
+    fi
+
+    echo -ne '\n   ]'
+}
+# --
+
+__slsa_provenance_report() {
+    local scripts_hash="$(cat "/mnt/host/source/src/scripts/.git/ORIG_HEAD")"
```
krnowak, line 124: This probably should use HEAD, but this maybe in form of `ref: refs/heads/main`, in such case a value from `.git/refs/heads/main` should be taken.

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+    # extract board from e.g. '/build/amd64-usr/build'. Empty if no board is set (SDK build).
+    local board="$(echo "${CROS_BUILD_BOARD_TREE:-}" | sed -n 's:^/build/\([^/]\+\)/.*:\1:p')"
+    if [ -n "$board" ] ; then
+        buildcmd="emerge-${board}"
+    fi
+
+    # FIXME: Supply SDK image ID and sha256 digest along with the version tag
+    local sdk_version="$(source /mnt/host/source/.repo/manifests/version.txt; echo ${FLATCAR_SDK_VERSION})"
+
+    # FIXME: add builder ID
```
krnowak, line 135: Are these things to be fixed before merging this PR?

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: good idea, i'll add a sort to the output

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: @t-lo any thoughts on the 3rd bullet point?

t-lo: Both is valid in SLSA, the uri determines the semantics of the checksum.
I used the git checksum as it seemed to be more useful for reproducing the build, but Krzesimir's first bullet point addresses this. So yes, if we can find the path to the ebuild and generate its sha512sum that would work for me.

profiles/coreos/base/profile.bashrc.slsa-provenance
t-lo: This needs to be fed from the outside e.g. by build automation (i.e. scripts in `ci-automation/`); we could use a placeholder variable here which is assumed to be filled by automation. So it would be a stub but we could address it in the scope of this PR.

profiles/coreos/base/profile.bashrc.slsa-provenance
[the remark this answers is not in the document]
jepio: switched to bz2

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+EOF
+}
+# --
+
+cros_post_src_install_generate_slsa_provenance_report() {
+    if [ "${__slsa_provenance_report:-no}" != "yes" ] ; then
+        return
+    fi
+
+    mkdir -p "$D/usr/share/SLSA/"
```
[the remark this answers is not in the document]
jepio: dropped this line

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,182 @@
...
+        return 0
+    fi
+    export __slsa_provenance_report="yes"
+}
+# --
+
+# Generate SLSA provenance 0.2 Subject information.
+# The information will cover all installation files shipped with a package.
+__slsa_provenance_subject() {
+    local parallel="$(grep -cE '^processor' /proc/cpuinfo)"
```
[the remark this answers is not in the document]
jepio: applied

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: > I think URI should contain the specific commit this ebuild was taken from, so for instance, it should be in form like https://raw.githubusercontent.com/flatcar-linux/coreos-overlay/3829e4d0dd8adb5c2af47f543723c3a525b5e743/sys-apps/systemd/systemd-250.3.ebuild

I've switched uri to contain the commit hash according to https://spdx.github.io/spdx-spec/package-information/#77-package-download-location-field.

> ORIG_HEAD contains a previous state of HEAD. It should use HEAD. Also, ORIG_HEAD does not even exist on first update of submodules (I don't have it here).

Fixed that.

> It's confusing to record commit checksum of the repo and use URI to an ebuild. Shouldn't it be rather checksum of the ebuild file?

The csum field now contains the actual file checksum, but the the uri contains the commit hash. I think that covers everything.

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: yup, fixed this

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: this uses HEAD now and resolved the ref

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: I'm not sure what this field is supposed to mean. This https://github.com/slsa-framework/slsa-github-generator/blob/main/PROVENANCE_FORMAT.md#builder claims it's meant to be the "job definition"

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,194 @@
...
+
+    einfo "    Provenance: recording subject (output)"
+
+    echo '   "subject": ['
+
+    (
+        cd "$D"
+        find . -type f -print | sed 's:^./::' | xargs -P "$parallel" -L 1 sha512sum | sort -k2
+    ) | while read checksum filepath; do
+        echo -en "${comma}     {\"name\":\"/$filepath\", \"digest\":{\"sha512\":\"$checksum\"}}"
```
krnowak, line 35: This prepends `/` to the "$filepath". I can assume that this results in a proper absolute path. I can't verify it, since build 6068 is already gone from storage.

krnowak (approved):

krnowak:
Just downloaded the image from build 6124 and did `xzcat /usr/share/dev-libs_glib-2.68.4.json.xz | less` to see the output. There seem to be a problem with checksums - all of them in materials and in invocation seem to be empty.

profiles/coreos/base/profile.bashrc.slsa-provenance
```diff
@@ -0,0 +1,194 @@
...
+
+cros_post_src_install_generate_slsa_provenance_report() {
+    if [ "${generate_slsa_provenance_report:-no}" != "yes" ] ; then
+        return
+    fi
+
+    local report_file="${CATEGORY}_${PF}.json.bz2"
+    local dest_dir="${D}/usr/share/SLSA/"
+
+    __slsa_provenance_report | lbzip2 -9cz > "${T}/${report_file}"
```
krnowak, line 190: Maybe use jq to format the report nicely:
```suggestion
    __slsa_provenance_report | jq | lbzip2 -9cz > "${T}/${report_file}"
```

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: correct, `$filepath` is initially relative to `${D}` so with the prepended `/` the entry becomes absolute. Here's a random part of the glibc report:
```
      {
        "name": "/usr/share/i18n/locales/pa_IN",
        "digest": {
          "sha512": "c7868c23a8f9a48118b150b5327565d4b4f0eca1afcf6061f5068f65a5594b8b17434ec8a16427dd693429a4d43cbb906fc27df6262872dc1199cd83904958c7"
        }
      },
      {
        "name": "/usr/share/i18n/locales/pa_PK",
        "digest": {
          "sha512": "b26b6a0716cdef6c96cda1be9c51b2ee0959e54b8afac0076022a97dae88921e8e94fa25907d01f21dfd57e2daa213f92f5eccf36cb6298e1b8afc340277b856"
        }
      },
```

jepio:
> Just downloaded the image from build 6124 and did `xzcat /usr/share/dev-libs_glib-2.68.4.json.xz | less` to see the output. There seem to be a problem with checksums - all of them in materials and in invocation seem to be empty.

I looked at the latest build 6199, and everything looks the way I expect. The file/ebuild hashes are correct, but the commit hashes are blank, because those depend on the container pipeline. I think this is still suitable for merge with the disclaimer that it will only be useful when releases are produced by the container pipeline. But atleast it won't suffer as much bitrot.

profiles/coreos/base/profile.bashrc.slsa-provenance
jepio: done

krnowak:
> > Just downloaded the image from build 6124 and did `xzcat /usr/share/dev-libs_glib-2.68.4.json.xz | less` to see the output. There seem to be a problem with checksums - all of them in materials and in invocation seem to be empty.
> 
> I looked at the latest build 6199, and everything looks the way I expect. The file/ebuild hashes are correct, but the commit hashes are blank, because those depend on the container pipeline. I think this is still suitable for merge with the disclaimer that it will only be useful when releases are produced by the container pipeline. But atleast it won't suffer as much bitrot.

Thanks for pointing me to the newer build. Now looks alright, besides those commit hashes in URIs.

krnowak (approved):

PR #2038, opened 2022-07-18
profiles: enable symlink for nmap

tormath1:
For compatiblity, it's good to have `nc` in the PATH too.

<hr>

See: https://github.com/flatcar-linux/portage-stable/pull/345


pothos (commented):
How about a changelog entry that `nc` is available now as symlink to `ncat` and thus doesn't support `-q`?

PR #2071, opened 2022-08-02
sys-kernel/coreos-sources: Regenerate patches

krnowak:
The first hyperv patch needed a regeneration, while the last one could
be dropped as it was already backported to 5.15.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/80/cldsv/

CC: @jepio

### DIFF
--- a/sys-kernel/coreos-sources/coreos-sources-5.15.58.ebuild
+++ b/sys-kernel/coreos-sources/coreos-sources-5.15.58.ebuild
@@ -40,4 +40,6 @@ UNIPATCH_LIST="
 	${PATCH_DIR}/z0003-PCI-hv-Make-the-code-arch-neutral-by-adding-arch-spe.patch \
 	${PATCH_DIR}/z0004-PCI-hv-Add-arm64-Hyper-V-vPCI-support.patch \
 	${PATCH_DIR}/z0005-Drivers-hv-vmbus-Propagate-VMbus-coherence-to-each-V.patch \
+	${PATCH_DIR}/z0006-PCI-hv-Avoid-the-retarget-interrupt-hypercall-in-irq.patch \
+	${PATCH_DIR}/z0007-PCI-hv-Remove-unused-hv_set_msi_entry_from_desc.patch \
 "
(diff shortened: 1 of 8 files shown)


krnowak:
Will need backporting to 3277, 3227 and 3305.

jepio:
I don't see that patch 5 has been backported?

krnowak:
It was probably rewritten with a different fix - https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=92a3499933c17b25b7c59658183bc7d23a581f5b

jepio:
That's a different patch that we used to carry as well but dropped a while ago: https://github.com/flatcar-linux/coreos-overlay/commit/351e702b29e25439ffcf17bf111930d145399774

jepio (commented):
The patches refreshes look OK, but we shouldn't drop z0005. Can you make sure that we run the tests on azure (and azure arm64) instances? The patches are backports of arm64 support.

jepio:
Patch 0003 is hitting another collision, this time with https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=5c676a214d6359ef84a673ca791716f8c76555cf, we need to move the hv_msi_prepare definition from there.

krnowak:
Yeah, the patches were applied out of order, so I already dropped a duplicate `hv_msi_get_int_vector` function, but it looks like more work is needed. Will also bring back the patch.

jepio:
It's not that it doesn't apply - but the result does not compile due to duplicate definitions.

krnowak:
Meh, was too hasty with running the job and forgot to set it up to run tests on azure. Will rerun the tests when the build is done.

tormath1 (commented):
LGTM from a build PoV - rebased one of my branch on it and it builds fine :)

krnowak:
Cool, arm64 still fails, though.

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/89/cldsv/

dongsupark:
Finally Azure CI passed for both amd64 and arm64.

dongsupark (approved):

krnowak:
Cherry-picked to 3227, 3277 and 3305.

PR #2096, opened 2022-08-18
dev-libs/cyrus-sasl: Move to portage-stable

krnowak:
This will be updated to version 2.1.28 there which contains the fix we were carrying ourselves.

The cross-compilation issue is fixed with env var override.

There is also a small profile cleanup (removing a redundant entry).

Should be merged together with https://github.com/flatcar-linux/portage-stable/pull/350.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/173/cldsv/


krnowak:
CI passed.

PR #2111, opened 2022-08-23
Toolchain update

krnowak:
- Some profiles and coreos/config cleanup related to updated packages in portage-stable PR.
- Needed to modify sssd with another small patch to fix the build.
- Also moved GNU make here, so I could add a patch to fix building binutils.
- Updated glibc to latest 2.33 revision from Gentoo.

CI passed: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/204/cldsv/

Needs to be merged together with https://github.com/flatcar-linux/portage-stable/pull/351 and https://github.com/flatcar-linux/scripts/pull/419.


sys-libs/glibc/README.md
```diff
@@ -1,9 +0,0 @@
-# GLibc
```
dongsupark, line 1: `sys-libs/glibc/README.md` has disappeared, and not added again later.
Is that intended?

dongsupark (approved):

sys-libs/glibc/README.md
krnowak: Hm, no. Something must went wrong. Will fix it.

krnowak: It's back.

PR #2122, opened 2022-08-30
coreos-base/oem-packet: Use i386 serial settings for x86_64 to fix m3

pothos:
The m3.small.x86 instance type had no serial console output because
ttyS0 was used because the GRUB CPU check didn't trigger. It seems that
most instances had i386 reported but this new one not (maybe EFI is
used here?).
Extend the GRUB check to cover both i386 and x86_64 when setting up the
serial console. For arm64 this still shouldn't be needed and the
defaults worked so far.

## How to use

Test again on other instance types that the serial console works as expected.

Backport

## Testing done

On m3.small, c3.medium, s3.xlarge


krnowak (approved):

pothos:
It seems the Equinix Metal provisioning logic already appends `set linux_console="console=tty0 console=ttyS1,115200n8"` to the grub.cfg file because the if condition wasn't met. Not sure how - I guess by injecting an ignition argument or through a custom image?

PR #2129, opened 2022-09-05
app-emulation/containerd: fix checksums of containerd 1.5.13

dongsupark:
CI fails with LTS flatcar-3033 when building containerd 1.5.13, because its upstream Github tarball suddenly started to have a different file size and different checksums.

```
[2022-09-02T23:05:07.727Z] !!! Fetched file: containerd-1.5.13.tar.gz VERIFY FAILED!
[2022-09-02T23:05:07.727Z] !!! Reason: Filesize does not match recorded size
[2022-09-02T23:05:07.727Z] !!! Got:      7514581
[2022-09-02T23:05:07.727Z] !!! Expected: 7514582
[2022-09-02T23:05:07.727Z] Refetching... File renamed to
'/mnt/host/source/.cache/distfiles/containerd-1.5.13.tar.gz._checksum_failure_.d2qh50kb'
```

Regerate checksums to fix the build issues.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/261/cldsv/


krnowak (approved):
Interesting, the size actually went down by one byte.

tormath1:
> Interesting, the size actually went down by one byte.

yeah, like before...

krnowak:
> > Interesting, the size actually went down by one byte.
> 
> yeah, like before...

Oh, I thought it grew by one byte last time, because they needed to add one more letter to the abbreviated commit hash somewhere in the tarball. Which maybe was the case but it compressed better?

PR #2140, opened 2022-09-13
Update vim to v9.0.0469

krnowak:
~These PRs use the 9999 ebuild for vim and pick the specific tag to use with EGIT_COMMIT variable. It's a bit of a hack, but it should not require us to wait for Gentoo to bump the package to the desired version.~

Here I'm moving vim-core and vim to coreos-overlay and bump the version.

Also, do some cleanups in profiles. As a part of it, I enabled the `minimal` USE flag in vim package (it was enabled before only for vim-core).

Should be merged with https://github.com/flatcar-linux/portage-stable/pull/357.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/238/cldsv

Image diff: http://jenkins.infra.kinvolk.io:8080/job/container/job/image_changes/127/console

There seem to be a lot of files installed in `/usr/share/vim/vim90` - not much different from before the update, where vim had lot of files in `/usr/share/vim/vim82`.

Also size diff - probably comes from gettext that we started putting in prod image for envsubst.

```
-Usr: 1137 MiB (inc. sparse files/hardlinks)
+Usr: 1138 MiB (inc. sparse files/hardlinks)
```


krnowak:
CI passed.

dongsupark (approved):
It looks good. Thanks!

Though I am not sure if we would need to install files like `/usr/share/icons/hicolor/scalable/apps/gvim.svg` or a bunch of files under `/usr/share/vim/vim90/keymap/`, `/usr/share/vim/vim90/lang/`. There is for sure a room for improvement.

krnowak:
> It looks good. Thanks!
> 
> Though I am not sure if we would need to install files like `/usr/share/icons/hicolor/scalable/apps/gvim.svg` or a bunch of files under `/usr/share/vim/vim90/keymap/`, `/usr/share/vim/vim90/lang/`. There is for sure a room for improvement.

Yeah, but I'm not a vim user to be able to tell if those files are needed or not. I would expect that enabling `minimal` USE flag on both vim-core and vim would result in minimal installation, but apparently we will need to minimize it even more.

dongsupark:
Moved to Release 2022-10-24.

PR #2153, opened 2022-09-14
app-admin/mayday: Use renamed package name

pothos:
This pulls in https://github.com/flatcar/mayday/pull/10 to update the package name after the github org move. It also changes the homepage to use our repo instead of the archive.

## How to use

Backport

## Testing done

none, I would say we only start the test build after all changes are prepared, it won't work otherwise


krnowak (changes requested):
`COREOS_GO_PACKAGE="github.com/coreos/mayday"` in line 8 should use flatcar.

pothos:
> `COREOS_GO_PACKAGE="github.com/coreos/mayday"` in line 8 should use flatcar.

Thanks, updated

krnowak (approved):

PR #2196, opened 2022-09-29
sys-kernel/coreos-module: use strip-debug instead of strip-unneeded

jepio:
# sys-kernel/coreos-module: use strip-debug instead of strip-unneeded

With `--strip-unneeded` some static symbols are also stripped from modules, making stacktraces incomplete, and making it harder to debug kernel issues. Switch to the default setting of `--strip-debug`, which keeps symbols intact and does not appear to lead to a measurable size increase of the /usr partition.

## How to use

`emerge-amd64-usr coreos-modules coreos-kernel`

## Testing done

Booted image and checked:
```
$ grep kjournald /proc/kallsyms
0000000000000000 t kjournald2   [jbd2]
```

### DIFF
--- a/changelog/changes/2022-09-30-kernel-strip.md
+++ b/changelog/changes/2022-09-30-kernel-strip.md
@@ -1,0 +1,1 @@
+- Switched from `--strip-unneeded` to `--strip-debug` when installing kernel modules, which makes kernel stacktraces more accurate and makes debugging issues easier ([coreos-overlay#2196](https://github.com/flatcar/coreos-overlay/pull/2196))
--- a/sys-kernel/coreos-modules/coreos-modules-5.15.70.ebuild
+++ b/sys-kernel/coreos-modules/coreos-modules-5.15.70.ebuild
@@ -38,7 +38,7 @@ src_install() {
 	# The linux-firmware package will be used instead.
 	# Stripping must be done here, not portage, to preserve sigs.
 	kmake INSTALL_MOD_PATH="${D}/usr" \
-		  INSTALL_MOD_STRIP="--strip-unneeded" \
+		  INSTALL_MOD_STRIP="--strip-debug" \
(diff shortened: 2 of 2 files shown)


pothos:
> Changelog entries added

Maybe it's good to include one here for future reference about which release includes this change

pothos:
Is there any size difference on `/boot`?

jepio:
Will add the changelog, and check the size difference on boot. I understand that we're close to the limit. This only affects modules, so the effect would be through the modules embed within the initramfs. Modules themselves are xz compressed.

jepio:
So from http://jenkins.infra.kinvolk.io:8080/job/container/job/image_changes/212/console and http://jenkins.infra.kinvolk.io:8080/job/container/job/image_changes/203/console:

||3346.0.0 | 3376.0.0-nightly | this PR|
|-|-|-|-|
|Boot|54|54|55|
|Usr|1137|1138|1139|
|Sum|1203|1204|1206|

pothos:
Good, for arm64 there was no raise reported (Boot: 58 MiB)

changelog/changes/2022-09-30-kernel-strip.md
```diff
@@ -0,0 +1 @@
+- Switched from `--strip-unneeded` to `--strip-debug` when installing Kernel modules, which has a size impact on the image but makes Kernel stacktraces more accurate and makes debuggging issues easier ([coreos-overlay#2196](https://github.com/flatcar/coreos-overlay/pull/2196))
```
pothos, line 1:
```suggestion
- Switched from `--strip-unneeded` to `--strip-debug` when installing kernel modules, which makes kernel stacktraces more accurate and makes debugging issues easier ([coreos-overlay#2196](https://github.com/flatcar/coreos-overlay/pull/2196))
```

pothos (approved):

PR #2207, opened 2022-10-07
Enable nf_conntrack_bridge and nft_meta_bridge kernel modules

lukasstockner:
# Enable nf_conntrack_bridge and nft_meta_bridge kernel modules

This allows to use conntrack rules for bridges in nftables and to match on bridge interface names.

## How to use

- Build kernel
- Run resulting image, note that `modprobe nf_conntrack_bridge` works now
- Set up nftables chain that uses conntrack (e.g. the [stateful bridge firewall example](https://wiki.nftables.org/wiki-nftables/index.php/Bridge_filtering#Example:_Stateful_bridge_firewall)), note that the rules are applied successfully.

## Testing done

Built image locally, booted on bare metal, tested stateful bridge firewalling.


changelog/changes/2022-10-07-kernel-enable-bridge-conntrack.md
```diff
@@ -0,0 +1 @@
+- Added CONFIG_NF_CONNTRACK_BRIDGE (for nf_conntrack_bridge) and CONFIG_NFT_BRIDGE_META (for nft_meta_bridge) to the kernel config
```
pothos, line 1:
```suggestion
- Added `CONFIG_NF_CONNTRACK_BRIDGE` (for nf_conntrack_bridge) and `CONFIG_NFT_BRIDGE_META` (for nft_meta_bridge) to the kernel config
```
Thanks, since this is for users to read more than developers, it would be good to also summarize the effect/benefit of those

pothos: (like in the git commit message body)

pothos:
I've started a [test build](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/498/cldsv/) just in case we run into something unexpected and also check the image size and content changes. Later I'll share the link to the built images.

changelog/changes/2022-10-07-kernel-enable-bridge-conntrack.md
pothos: Updated the suggestion

pothos (approved):

lukasstockner:
> @lukasstockner can you apply the suggestion?

Sure, done - thanks for improving it!

pothos:
Thanks, it will be part of the next major Alpha release

PR #2211, opened 2022-10-10
sys-apps/dbus: Update to 1.14.4

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/324/cldsv

Nothing interesting in changes.


krnowak:
CI passed for amd64 (for arm64 is still running, failed earlier because builder ran out of space).

PR #2213, opened 2022-10-11
changelog: Add an entry for go 1.18.7 update

krnowak:
Merged #2208, but forgot to add the security changelog.


krnowak:
https://nvd.nist.gov/vuln/detail/ links didn't work for those CVEs, so used https://cve.mitre.org/cgi-bin/cvename.cgi ones instead.

dongsupark (approved):
Looks good.

FYI, you can continue writing NVD links, even when the CVEs are not available on NVD yet.
In some cases, the NVD links become available after many weeks or months.

krnowak:
> Looks good.
> 
> FYI, you can continue writing NVD links, even when the CVEs are not available on NVD yet. In some cases, the NVD links become available after many weeks or months.

Ok, thanks for the hint. I'll use the NVD links next time.

sayanchowdhury:
We mistakenly added the CVE to the `updates` folder instead of the `security` folder. We would need to keep this track in the future.

PR #2221, opened 2022-10-13
github: Deduplicate kernel and ca-certificates workflows

krnowak:
Instead of duplicating workflows for each branch we have, have just one with a matrix. It should automatically handle the fact of deprecating the old LTS channel.

Also, the ca-certificates and kernel actions should now apply the correct label (alpha or stable instead of flatcar-XXXX).

## How to use

Fork the repo, have my branch as the one used for picking up actions from, make sure that the branch has some old versions of ca certificates and kernel, and run the workflow manually.

## Testing done

See PRs in https://github.com/krnowak/coreos-overlay/pulls.

Nothing changeloggable and nothing with the image generation.


.github/workflows/cacerts-apply-patch.sh
```diff
@@ -2,14 +2,13 @@
...
 
+BRANCH_NAME="cacerts-${VERSION_NEW}-${TARGET}"
+
 prepare_git_repo
 
-if ! checkout_branches "${VERSION_NEW}-${TARGET}"; then
-  UPDATE_NEEDED=0
+if ! checkout_branches "${BRANCH_NAME}"; then
   exit 0
 fi
```
dongsupark, line 13: We set `UPDATE_NEEDED` in order not to push unnecessary commits to existing open PRs that were created for the same branch.
We have a history of having troubles with noisy commits for existing open PRs, when we tried to touch `UPDATE_NEEDED` in that code.
If we would want to remove that part like above, I would recommend testing the specific scenarios using this PR.

.github/workflows/kernel-apply-patch.sh
```diff
@@ -4,14 +4,13 @@ set -euo pipefail
...
 
 . .github/workflows/common.sh
 
 prepare_git_repo
 
-if ! checkout_branches "linux-${VERSION_NEW}-${TARGET}"; then
-  UPDATE_NEEDED=0
+if ! checkout_branches "${BRANCH_NAME}"; then
   exit 0
 fi
```
dongsupark, line 15: same here

dongsupark (commented):
+1 for reducing the lines of code.

See below:

.github/workflows/figure-out-branch.sh
```diff
@@ -0,0 +1,71 @@
...
+channel_name="${1}"
+skip=0
+link=''
+branch=''
+label=''
+case "${channel_name}" in
+    main)
+        branch='main'
+        ;;
+    lts-old)
```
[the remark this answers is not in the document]
krnowak: `lts-old` will become a no-op when we deprecate LTS-2021. And will become meaningful again, after introducing LTS-2023. At this point, `lts-old` will handle LTS-2022. For this reason, I'm rather suggesting to keep it. We don't need to do anything to this github workflow when new LTS major release is made or deprecated.

.github/workflows/cacerts-apply-patch.sh
krnowak: UPDATE_NEEDED is set to zero, then we exit the script. The whole point of this variable is to be a substitution at the bottom of the script where we do `echo ::set_output` stuff. I'll rerun the workflows on my fork - there are already those branches, so next workflow run should just stop here.

krnowak: As a matter of fact, the [action for kernel updates run on my fork 4 hours ago](https://github.com/krnowak/coreos-overlay/actions/runs/3248096616) and all the jobs stopped at this checkout_checkout branches function with the message "Target branch already exists. exit." and no PR was created/updated. I've started a run for ca-certificates and guess what - got another set of 6 PRs because [ca-certificates 3.84 got released yesterday](https://firefox-source-docs.mozilla.org/security/nss/nss_3_84.html#mozilla-projects-nss-nss-3-84-release-notes). :) But reruninng it again yielded no new PRs nor any PR updates.

krnowak:
Updated to bump the action versions and to use the recommended way of setting output.

PR #2222, opened 2022-10-14
sys-libs/glibc: Update to 2.35

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/368/cldsv

- `catchsegv` and `libSegFault` were dropped
- Convenient symlink to dynamic linker is installed in `/usr/bin/ld.so`.
- There are some changes in the way shared libs are installed, it used to be `libfoo.so -> libfoo.so.1 -> libfoo-2.33.so`, now it mostly is `libfoo.so -> libfoo.so.1` or even in some cases it's just  `libfoo.so.1` (`libfoo.so` symlink is gone). Possibly some of those libs are now empty, as their functionality is moved to glibc itself.

- [arm64](http://jenkins.infra.kinvolk.io:8080/job/container/job/image_changes/319/console)
- [amd64](http://jenkins.infra.kinvolk.io:8080/job/container/job/image_changes/320/console)


dongsupark:
Have not looked into its detail.
But let's do the update to 2.35 at least after one Alpha release with 2.34, probably 2022-10-24.
If we would update directly from 2.33 to 2.35 in a single Alpha release, then a potential regression caused by the glibc updates could become a lot more difficult to debug.

krnowak:
Sure. I'll put it on hold then.

sys-libs/glibc/glibc-2.35-r8.ebuild
```diff
@@ -818,12 +823,7 @@ upgrade_warning() {
...
 # pkg_pretend
 
-# Flatcar: Skip sanity checks at pretend time because we don't ship a compiler
-#  in the OS image. This test fails when installing the glibc binpkg and no
-#  compiler is present.
 pkg_pretend() {
-	einfo "Flatcar: Skipping sanity_prechecks for binpkg installation. src_unpack will take care of compile-time prechecks."
-	# sanity_prechecks
 	upgrade_warning
 }
```
krnowak, line 828: This change is not needed any more: https://github.com/gentoo/gentoo/blob/master/sys-libs/glibc/glibc-2.35-r8.ebuild#L826-L828

dongsupark (approved):
Looks good.
Again I cannot see the CI results, as it ran more than a week ago.

krnowak:
I'll rerun the CI.

krnowak:
CI passed.

PR #2246, opened 2022-10-20
coreos-base/oem-ec2-compat: pull ssh keys from metadata server

tormath1:
In this PR, we're enabling the `coreos-metadata-sshkeys@.service` for the OpenStack image in order to pull SSH keys from the OpenStack metadata server if no user configuration is provided.

## How to use

Boot an instance on OpenStack without Ignition provided SSH keys. 

## Testing done

Manually tested with https://github.com/flatcar/mantle/pull/388

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/585/cldsv/ 

<hr>

Should fix: https://github.com/flatcar/Flatcar/issues/817

### DIFF
--- a/coreos-base/oem-ec2-compat/oem-ec2-compat-0.1.2-r2.ebuild
+++ b/coreos-base/oem-ec2-compat/oem-ec2-compat-0.1.2-r2.ebuild
@@ -63,4 +63,8 @@ src_install() {
 	if use ec2 ; then
 		newins "${FILESDIR}/base/base-ec2.ign" base.ign
 	fi
+
+	if use openstack; then
+		newins "${FILESDIR}/base/openstack.ign" base.ign
+	fi
 }
--- a/changelog/changes/2022-10-19-openstack-metadata.md
+++ b/changelog/changes/2022-10-19-openstack-metadata.md
@@ -1,0 +1,1 @@
+- OpenStack: enabled `coreos-metadata-sshkeys@.service` to provision SSH keys from metadata. ([Flatcar#817](https://github.com/flatcar/Flatcar/issues/817), [coreos-overlay#2246](https://github.com/flatcar/coreos-overlay/pull/2246))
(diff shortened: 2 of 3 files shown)


krnowak (approved):

coreos-base/oem-ec2-compat/oem-ec2-compat-0.1.2-r2.ebuild
```diff
@@ -63,4 +63,6 @@ src_install() {
 	if use ec2 ; then
 		newins "${FILESDIR}/base/base-ec2.ign" base.ign
 	fi
+
+	use openstack && newins "${FILESDIR}/base/openstack.ign" base.ign
```
pothos, line 67: Does the return code of `src_install` matter or can it be non-0?

tormath1: `newins` calls `die` if there is an error during the installation of the file: this operation should succeed or die.

krnowak: I think the question is whether the return value of `src_install` has any meaning. This function's return value will the return value of `use openstack && newins "…"`, which will be 1 for non-openstack images (because `use openstack` will return 1).

krnowak: I think I'd recommend rewriting this as `if use openstack; then newins "…"; fi` for consistency.

tormath1: Yeah I tend to follow this documentation: https://devmanual.gentoo.org/ebuild-writing/use-conditional-code/index.html
> For single-statement conditions, the use foo && blah (or use foo || blah for negatives) form is often more readable.

But yes, in this case it's wrong to return `use openstack && newins "…"` as it will be 1 for non-openstack image (AWS). Thanks!

krnowak (approved):

tormath1:
cherry-picked to:
- flatcar-3227
- flatcar-3346
- flatcar-3374

PR #2268, opened 2022-11-02
net-misc/openssh: Bump to 9.1

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/367/cldsv


krnowak:
Blocked on [weekly updates](https://github.com/flatcar/portage-stable/pull/380/files) to get in - we need an updated `sys-devel/gcc-config`.

krnowak:
CI passed.

dongsupark (approved):
This PR looks good.

Note, openssh 9.0 or newer has potential incompatibility issues like using SFTP by default instead of legacy scp/rcp protocol.
That could be a source of bug reports.
Hope the issues could be found before hitting Stable.

PR #2269, opened 2022-11-03
dev-lang/go: add 1.19.3, update 1.18.8, remove 1.17

dongsupark:
* Add Go [1.19.3](https://go.dev/doc/devel/release#go1.19.3), and update the default Go version to 1.19.
Note, we still keep `COREOS_GO_VERSION=go1.18` in `containerd`, `docker`, `docker-cli`, `docker-proxy`, `docker-runc` following the default version of the upstream repos.

* Update Go 1.18.7 to [1.18.8](https://go.dev/doc/devel/release#go1.18.8).
Note, the security issue of the release does not affect Flatcar, as that affects only Windows.

* Go 1.17 is no longer used anywhere. Clean up.

* Update list of required docs of `dev-lang/go` as the source tree of Go 1.19 or newer does not have files like `AUTHORS` or `CONTRIBUTORS`. We need to remove the files from the list of required docs. Also add `CONTRIBUTING.md` to the list of docs, as all Go versions have the file.

* Make `coreos-base/hard-host-depends` pull in `dev-lang/go:1.19` instead of `1.17`, to avoid build failures.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/371/cldsv/


krnowak (commented):
Could you also please update `go-releases-main.yml`? There is a list of major versions to update - it still has 1.17 there and no 1.19.

dongsupark:
Done updating go-releases-main.yml.

dongsupark:
have just realized that the changelog URL was broken.
It should be not https://go.dev/doc/devel/release#1.19.3 but https://go.dev/doc/devel/release#go1.19.3.
Fixed also `go-apply-patch.sh`.

krnowak:
> have just realized that the changelog URL was broken. It should be not https://go.dev/doc/devel/release#1.19.3 but https://go.dev/doc/devel/release#go1.19.3. Fixed also `go-apply-patch.sh`.

Maybe that has changed over time. But well spotted and thanks for fixing it.

PR #2277, opened 2022-11-10
sys-apps/ignition: fix OEM detection

tormath1:
In case the partition was `OEM`, the `btrfs` formatting was not forced if `WipeFilesystem` was not provided.

<hr>

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/725/cldsv/

Closes: https://github.com/flatcar/Flatcar/issues/885


pothos:
> In case the partition was OEM, the btrfs formatting was not forced if WipeFilesystem was not provided.

I think this would be clearer:

> In case the OEM partition was specified with the name "OEM", the btrfs format was not forced because it only considered the name "oem".

sys-apps/ignition/files/0017-translation-support-OEM-and-oem.patch
```diff
@@ -0,0 +1,161 @@
...
++							Name: "OEM",
++							Mount: &types2_4.Mount{
++								Device: "/dev/disk/by-label/OEM",
++								Format: "ext4",
++							},
++						},
++					},
++				},
++			},
++			fsFormat: "btrfs",
```
krnowak, line 77: Just to understand one bit: does the test assume that initially OEM filesystem is using btrfs?

tormath1: so Flatcar uses `btrfs` for OEM partition since this commit: https://github.com/flatcar/scripts/pull/131/files#diff-61886cc88e76de6ba4a72062d3bf81cce88bc97a8fa6ab6f874a214e9f92ba0dR54

This unittest test is mainly to ensure that there's no regression on the `OEM` / `oem` initial issue.

We also have Mantle test for this: https://github.com/flatcar/mantle/blob/4b3631f743d130b1d1c20ddf3036fe8648967066/kola/tests/ignition/oem.go#L25-L67

krnowak: Cool, thanks.

krnowak (approved):

tormath1:
cherry-picked to:
- flatcar-3417
- flatcar-3402
- flatcar-3374

PR #2305, opened 2022-11-28
sys-kernel/coreos-modules: Enable CONFIG_EFI_VARS on arm64

pothos:
For arm64 CONFIG_PSTORE is not set and the EFI variable pstore backend is disabled.
Enable CONFIG_EFI_VARS which should in turn enable the other settings.

## Testing done

[started](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/820/cldsv/)


jepio:
When debugging the ext4 kernel issue, i upgraded to kernel 6.0 and found that this config has been removed in a newer kernel. Would you be able to look up on lkml why that is and what's the replacement?

pothos:
Thanks, maybe if it's gone we might have to set CONFIG_EFI_VARS_PSTORE explicitly and check the value of CONFIG_EFI_VARS_PSTORE_DEFAULT_DISABLE (since we ship systemd-pstore.service I think we are quite safe enabling it by default because they get cleaned on boot).

pothos:
Anyway, `CONFIG_EFI_VARS` couldn't be enabled in arm64 due to `depends on EFI && (X86 || IA64)`

jepio:
what about `CONFIG_EFIVAR_FS`?

pothos:
That's set according to https://stable.release.flatcar-linux.net/arm64-usr/current/flatcar_production_image_kernel_config.txt

PR #2307, opened 2022-11-29
Changes for weekly updates

krnowak:
Some profile cleanups, and pull in `dev-util/b2` instead of `dev-util/boost-build` into SDK for `dev-libs/boost`. We updated boost in portage-stable and it started to depend on the former instead of the latter.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/429/cldsv/

Should be merged together with https://github.com/flatcar/portage-stable/pull/392


krnowak:
CI passed.

dongsupark (approved):

PR #2325, opened 2022-12-09
sys-apps/portage: Drop our patches, clean up modifications and move to portage-stable

krnowak:
Should be merged together with https://github.com/flatcar/portage-stable/pull/398 and https://github.com/flatcar/scripts/pull/606.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/513/cldsv/

Some modifications that we were making in the ebuild could be done in profiles (disabling rsync, not installing/clobbering the default repos.conf). The patches we carried so far seem to be unnecessary as we have managed without them for a short time before @jepio noticed that these are not applied and made them to be applied again. With that done, we can move the package to portage-stable and have it updated automatically there.


krnowak:
CI passed.

krnowak:
Did a package refresh and a CI rerunning, all green.

PR #2333, opened 2022-12-12
Changes for weekly updates (update glibc to 2.36)

krnowak:
Should be merged together with https://github.com/flatcar/portage-stable/pull/399.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/456/cldsv

glibc changes:

- dropped clone3 keywords (probably enabled by default now)
- experimental-loong USE flag dropped
- new hash-sysv-compat USE flag (something to enable if there are problems with DT_HASH being absent in glibc.so file)
- some variables got renamed from `__stuff` to `glibc__stuff`
- fix kernel headers version check (do it in proper phase)


krnowak:
CI passed.

dongsupark (approved):

PR #2343, opened 2022-12-16
sys-apps/systemd: Update to 250.7

krnowak:
CI passed: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/929/cldsv/

Why to 250.7 and not to 250.8? I picked the latest Gentoo ebuild for 250.x before it was dropped from the repository, and it was for 250.7. Looks like 250.8 never made to Gentoo.

Next step could be updating to 251.x after this PR gets merged.


krnowak:
> Can you please run CI for all cloud providers, not only qemu & DigitalOcean? In the past [flatcar/Flatcar#685](https://github.com/flatcar/Flatcar/issues/685) prevented us from updating systemd to >= 250.4, due to a networkd issue in Digital Ocean. To be completely sure, we should check if that still exists in any other cloud provider.

Sure, done. I have run it for Digital Ocean already, because I heard that there were issues with this platform. But anyway, rerunning for all the platforms. To see the jobs, please use the CI link in PR message.

krnowak:
All platforms green.

pothos:
Maybe add a changelog update entry even if we would directly update to 251 without an intermediate release but at least we could cherry pick this update here to other branches when needed.

krnowak:
Added the changelog.

pothos (approved):

pothos:
Thanks, depending on what we channels we release next this could even be backported to the next Alpha and Beta release.

PR #2351, opened 2022-12-21
sys-apps/ignition: bring back noop OEM

tormath1:
it mainly brings back Vagrant which was failing with Ignition 2.14.0 even if no Ignition is provided.

<hr>

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/947/cldsv/

_NOTE_: It's ok to me to backport on all channels: the added platforms were not just working before.

Related to: https://github.com/flatcar/Flatcar/issues/916

### DIFF
--- a/changelog/bugfixes/2023-01-03-ignition.md
+++ b/changelog/bugfixes/2023-01-03-ignition.md
@@ -1,0 +1,1 @@
+- Added back Ignition support for Vagrant [coreos-overlay#2351](https://github.com/flatcar/coreos-overlay/pull/2351)


krnowak (approved):
I'm wondering if this shouldn't be fixed by using `noop.FetchConfig` for the unregistered platforms, but I suppose this is outside the scope of this PR.

krnowak:
> > I'm wondering if this shouldn't be fixed by using `noop.FetchConfig` for the unregistered platforms, but I suppose this is outside the scope of this PR.
> 
> What do you mean by that?

Basically if somewhere in Ignition code there should be a snippet like (instead of trying to list all the old platforms):

```go
if config, ok := configs.getConfigForPlatformOrSomething(platformName); ok {
    config.fetch(…)
} else {
    # platform not registered, assume noop.FetchConfig will be enough for its needs
    noop.FetchConfig(…)
}
```

But wrt. Ignition, I'm an uninformed keyboard puncher here, so I might be spouting nonsense. :)

krnowak:
> @krnowak I'll go ahead for now - this is something we can suggest and track directly into the upstream `coreos/ignition` repo. :)

Sure, sounds good.

PR #2362, opened 2023-01-02
Profile and app/shells/bash updates for the weekly update

krnowak:
Should be merged together with https://github.com/flatcar/portage-stable/pull/401.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/535/cldsv

- app-shells/bash:
  - from 5.1_p8 to 5.1_p16-r2
  - added sig-verify stuff (unused still by us)
  - update done just in case it's needed for the app-alternatives work (gets pulled in by app-alternatives/sh)
  - changes: http://git.savannah.gnu.org/cgit/bash.git/log/?h=9439ce094c9aa7557a9d53ac7b412a23aa66e36b

This also cleans up our handling of split-usr USE flag, so it affect either all packages or none, instead only those listed. The cleanup allows us to drop one more modification from policycoreutils ebuild.

Another thing is adding a user patch for app-arch/ncompress to fix an installation of the uncompress symlink.


krnowak:
CI passed.

pothos (approved):

dongsupark (approved):

PR #2372, opened 2023-01-10
coreos-base/coreos-init: Pull in fixed sshd_config

krnowak:
According to sshd_config manual, only lines starting with a hash and empty lines are treated as comments. Which means that comments coming after the key value pairs are not really comments and sshd complains about them like:

/etc/ssh/sshd_config line 6: keyword PrintLastLog extra arguments at end of line
/etc/ssh/sshd_config line 7: keyword PrintMotd extra arguments at end of line
/etc/ssh/sshd_config: terminating, 2 bad configuration options

Pulls in https://github.com/flatcar/init/pull/85.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1039/cldsv


krnowak:
CI passed.

jepio (approved):

PR #2374, opened 2023-01-12
dev-lang/python-oem: update to 3.9.16

dongsupark:
Update `dev-lang/python-oem` to 3.9.16, the same version as `dev-lang/python` in portage-stable.
Based on Gentoo commit [8c8c938e158a](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=8c8c938e158a9e10f1c5a4280c16e435e354122e).

Apply Flatcar modifications, based on [32b448f8ccbf](https://github.com/flatcar/coreos-overlay/commit/32b448f8ccbf5f814739f834151803532272b1f9).
Unlike python <= 3.9.12, it is now unnecessary to replace `virtual/awk` with `app-alternatives/awk`, because the upstream python ebuild already has `app-alternatives/awk`.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1064/cldsv/


dongsupark:
CI passed

krnowak:
I see that you have run the tests on azure. Cool. You could have done also the same for arm64, but I assume that it should still be fine.

Anyway, thanks for doing it - updating python-oem is rather a hassle.

dev-lang/python-oem/python-oem-3.9.16.ebuild
```diff
@@ -104,9 +68,7 @@ src_unpack() {
 }
 
 src_prepare() {
-	# Ensure that internal copies of expat and libffi are not used.
-	rm -r Modules/expat || die
-	rm -r Modules/_ctypes/libffi* || die
+	# Ensure that internal copies of zlib are not used.
```
krnowak, line 71: I think you can drop this comment from our modifications - looks like internal copy of zlib is no more.

krnowak (approved):
Looks good, but please see one comment below.

krnowak (approved):

dongsupark:
Azure arm64 also passed http://jenkins.infra.kinvolk.io:8080/job/container/job/test/6204/

PR #2402, opened 2023-01-19
sys-kernel/coreos-sources: Add backport of bugfix for #847 for flatcar-3374

dongsupark:
Users reported a deadlock in ext4 that occurs under loads after kernel 5.15.72. We debugged and found that this issue is also present upstream (6.x) and found a fix. The fix has been validated to fix the issue.

Apply the backport to our kernel sources, so that users can be unblocked from updating.
This will be included in Stable 3374.2.3 with Kernel 5.15.86. Although the upstream stable Kernel 5.15.87 already has the fix, we should stay with 5.15.86 at least for this Stable channel release, to keep a certain degree of stability.

See also https://github.com/flatcar/Flatcar/issues/847, https://github.com/flatcar/coreos-overlay/pull/2315.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1121/cldsv/


jepio (approved):

dongsupark:
CI passed, fixed changelog.

PR #2403, opened 2023-01-20
Rework initrd usr mount to start Ignition/Afterburn from there

pothos:
This pulls in
https://github.com/flatcar/bootengine/pull/52
to reduce the size of the unified kernel image by starting Ignition/ Afterburn from the /usr partition.
This also requires to install the Ignition binary to /usr.

## Testing done

See linked PR

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) ← do we need one?

### DIFF
--- a/profiles/coreos/base/make.defaults
+++ b/profiles/coreos/base/make.defaults
@@ -89,9 +89,8 @@ INSTALL_MASK="
   /usr/share/portage/config/repos.conf
 "
 
-# Prevent 'ignition' binary from being installed from sys-apps/ignition binary package.
+# Prevent binaries from being installed to rootfs from binary packages
 PKG_INSTALL_MASK="
-  /usr/bin/ignition
 "
 
 # Keep the default languages small.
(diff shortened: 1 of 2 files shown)


profiles/coreos/base/make.defaults
```diff
@@ -89,9 +89,8 @@ INSTALL_MASK="
   /usr/share/portage/config/repos.conf
 "
 
-# Prevent 'ignition' binary from being installed from sys-apps/ignition binary package.
+# Prevent binaries from being installed to rootfs from binary packages
 PKG_INSTALL_MASK="
```
krnowak, line 93: Could this be dropped?

krnowak (approved):

profiles/coreos/base/make.defaults
pothos: For reference I kept it around

krnowak: Fair enough.

pothos: It's a handy mechanism to filter out unnecessary binaries without modifying an upstream ebuild file, maybe we should use it more

PR #2413, opened 2023-01-25
profiles: Pick defaults for app-alternatives during bootstrap

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/559/cldsv


krnowak:
@thesamesam: Excuse me for pulling you in here, but is this expected that Gentoo derivatives fix bootstrap issues with app-alternatives like this PR does?

krnowak:
Can't fix it by adding stuff to BOOTSTRAP_USE, because if I add `reference` flag for `app-alternatives/bzip2` and `app-alternatives/gzip` then it also gets picked up by the `app-alternatives/bison` for which I specified the `bison` use flag. The build then fails, because `app-alternatives` packages expect exactly one USE flag to be enabled.

Trying to add those flags to `package.use.force` instead.

krnowak:
2-phase SDK build in CI passed.

dongsupark (approved):

PR #2425, opened 2023-02-02
sys-apps/systemd: Update to 252.5

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/590/cldsv


pothos:
Maybe we should set `SYSTEMD_JOURNAL_COMPACT=0` for systemd-journald for the next months at least to give logging containers more time to update to the new format. Otherwise a changelog entry is needed. Or I misunderstood `HEADER_INCOMPATIBLE_COMPACT` as journal flag being on by default?

krnowak:
> Maybe we should set `SYSTEMD_JOURNAL_COMPACT=0` for systemd-journald for the next months at least to give logging containers more time to update to the new format. Otherwise a changelog entry is needed. Or I misunderstood `HEADER_INCOMPATIBLE_COMPACT` as journal flag being on by default?

I think you got it right. I suppose we could add a snippet to init repo that extends the journal unit with the env variable. Would this work?

Also, there's [a note about dropping support for split-usr and unmerged-usr](https://lists.freedesktop.org/archives/systemd-devel/2022-September/048352.html), but I haven't yet looked in details.

There's also this, which triggered some failures when Sayan tried to build systemd:
```
        * Private shared libraries (libsystemd-shared-nnn.so,
          libsystemd-core-nnn.so) are now installed into arch-specific
          directories to allow multi-arch installs.
```

pothos:
We could test it with booting your build and then `flatcar-update` to some old version and see if it can read the journal from the previous boot.

I've looked at the split/unmerged usr deprecation and I think we are good.

krnowak:
@pothos: Do you have any idea what creates a symlink from `/usr/lib64/systemd` to `/usr/lib/systemd`? With the systemd update, `/usr/lib64/systemd` stopped being a symlink, and became a separate directory, where the private systemd libraries are installed. And the image has now also `/usr/lib64/systemd.backup.0000` which is a symlink to `/usr/lib/systemd`. I'd like to get rid of it.

pothos:
I set up symlinks in https://github.com/flatcar/coreos-overlay/blob/main/coreos-base/coreos-init/coreos-init-9999.ebuild#L59 for aligning amd64 with arm64 because one was using lib64 and one was using lib as lib folder. The compat symlinks are for the case that something still uses the old path. We could check if we want to set up compat symlinks for a subdirectory or particular files instead if we have a conflict now with the systemd top folder.

pothos:
As far as I remember, `/usr/lib64/` was a symlink to `/usr/lib` and I split it up as usually done. However, that meant that some paths with `/usr/lib64/systemd/` were working before and to keep them working I had to add the compat symlinks. I fixed most places I found like `coreos-base/coreos-oem-gce/files/manglefs.sh` in 19f1adecef1325a513f0d597ba6b0ba8a4f28ef2 but there could still be usage in the wild. I'm not sure yet for what files/dirs under `/usr/lib/systemd/` we should set up symlinks from `/usr/lib64/systemd/`. Maybe we can also live without them and state that this is a breaking change for users.

krnowak:
I went with installing the compat symlinks in baselayout, so the private libs will be effectively still installed in /usr/lib/systemd. That also eliminated the need to fiddle with gce OEM's manglefs script.

As of journal file format - I did a downgrade to beta and journalctl could print old logs. It printed a warning about unsupported stuff. Thus I added another patch to systemd that adds SYSTEMD_JOURNAL_COMPACT=0 to journald units.

sys-apps/baselayout/baselayout-9999.ebuild
```diff
@@ -242,4 +242,15 @@ pkg_postinst() {
...
 		ln -sf "${func}" "${ROOT}/etc/init.d/functions.sh"
 	fi
+	# install compat symlinks in production images, not in SDK
+	# os-release symlink is set up in scripts
+	if ! use cros_host; then
+		local compat libdir
+		for compat in modules systemd flatcar coreos kernel modprobe.d pam pam.d sysctl.d udev ; do
+			for libdir in $(get_all_libdirs) ; do
+				if [[ "${libdir}" == 'lib' ]]; then continue; fi
+				ln -sfT "../lib/${compat}" "${ROOT}/usr/${compat}"
```
pothos, line 252: I don't follow, wouldn't it be `/usr/${libdir}/${compat}`?

krnowak: Meh.

krnowak: Fixed.

krnowak:
> As of journal file format - I did a downgrade to beta and journalctl could print old logs. It printed a warning about unsupported stuff. Thus I added another patch to systemd that adds SYSTEMD_JOURNAL_COMPACT=0 to journald units.

I meant to write that journalctl could NOT print old logs.

pothos:
Is the meson option `status-unit-format-default=combined` something we could set as downstream patch here?

krnowak:
Yes, we could patch `meson_options.txt` file to set the default value for `'status-unit-format-default'` option to `combined`. But I'd like to do it in a separate PR to get this PR ready for alpha ASAP.

pothos:
Yes, let's leave it for later, I would put `-Dstatus-unit-format-default=combined` below the `-Ddefault-kill-user-processes=false` line in the ebuild file where the Flatcar changes are.

krnowak:
For now I had to split the symlink installation between coreos-init and baselayout, because some symlink become dangling in oem gce image. Later I'll maybe add a use flag to pick a set of compat symlinks to install, so SDK won't have any of that, generic image all of them, and OEM image only some of those. Or we will bite the bullet and drop them altogether.

pothos:
Looks good, I (re)started the tests for all platforms on amd64

krnowak:
CI passed.

krnowak:
For the record, I checked manually if journal logs are still readable after downgrading from this developer build to beta and they are now.

pothos (approved):
Thanks a lot!

PR #2460, opened 2023-02-17
net-misc/openssh: update to 9.2_p1-r2

dongsupark:
Update `net-misc/openssh` to 9.2_p1-r2, mainly to address [CVE-2023-25136](https://nvd.nist.gov/vuln/detail/CVE-2023-25136).
Based on Gentoo commit [a773d7e0cdb9](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=a773d7e0cdb9579343c1a4437ed4ce97f9d91d1f).

Apply Flatcar patches based on [5a227154b57c](https://github.com/flatcar/coreos-overlay/commit/5a227154b57ce9a39ab08a436f28045b5fc8cf12).

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/643/cldsv/


krnowak (approved):
It's interesting to see that the only clear modification that we make here is a removal of rate-limiting on a socket. I think it's something we can do through an override file in the init repo, so this package could be moved to portage-stable. But this is a material for a future PR.

PR #2481, opened 2023-02-24
sys-apps/ignition: bump to 2.15.0

tormath1:
With this release, 3.4-exp configuration is stabilized to 3.4 and 3.5-exp is available. A bit of turbulence while upgrading the patches  (especially for the qemu fw_cfg one)

<hr>

 

- [x] http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1406/cldsv/ (all providers are tested)

### DIFF
--- a/changelog/updates/2023-02-24-ignition.md
+++ b/changelog/updates/2023-02-24-ignition.md
@@ -1,0 +1,1 @@
+- ignition ([2.15.0](https://coreos.github.io/ignition/release-notes/#ignition-2150-2023-02-21))


sys-apps/ignition/files/0001-sed-s-coreos-flatcar.patch
```diff
@@ -1,13 +1,13 @@
-From 4b86d64fe196d9cb5ca43548234d521907a1cb4d Mon Sep 17 00:00:00 2001
```
krnowak, line 1: Do we still need to carry this patch? I'm not sure what is the point of this rename, especially that we are not using our ignition fork any more.

krnowak (approved):
Looks good. Had a question about one patch, though.

sys-apps/ignition/files/0006-config-v3_4-convert-ignition-2.x-to-3.4.patch
```diff
@@ -1,24 +1,24 @@
...
  1 file changed, 37 insertions(+)
 
-diff --git a/config/v3_4_experimental/config.go b/config/v3_4_experimental/config.go
-index 097dbd83..e666c876 100644
---- a/config/v3_4_experimental/config.go
-+++ b/config/v3_4_experimental/config.go
+diff --git a/config/v3_4/config.go b/config/v3_4/config.go
+index 0a01a560..32770331 100644
+--- a/config/v3_4/config.go
++++ b/config/v3_4/config.go
```
pothos, line 19: Why is this 3.4, I think it should be 3_5_experimental because that's the version parsed by `config/config.go`

sys-apps/ignition/files/0001-sed-s-coreos-flatcar.patch
pothos: The old fork is still used for the translation

sys-apps/ignition/files/0006-config-v3_4-convert-ignition-2.x-to-3.4.patch
pothos: (If that is the case I wonder how the kola tests for spec 1 and 2 passed)

pothos: Having the translation step somewhere in the middle of the `ParseCompatibleVersion` chain is a bit confusing even if it now works. I suggest to either do it at the top as before or try to move it to `config/v3_0/config.go` which would be nice but is more work because the translator doesn't use 3.0 as far as I remember.

pothos (approved):

sys-apps/ignition/files/0006-config-v3_4-convert-ignition-2.x-to-3.4.patch
pothos: Thanks!

PR #2489, opened 2023-02-28
profiles: Enable systemd[elfutils] USE flag for board

jepio:
# profiles: Enable systemd[elfutils] USE flag for board
Elfutils is already part of the usr partition, but currently not enabled in systemd-coredump. Systemd-coredump therefore fails with:

  elfutils disabled, parsing ELF objects not supported.

Enable the elfutils flag for systemd to make this work.

## How to use

Dump core and check journal for messages.

## Testing done

Haven't tested, but will try it out once CI builds images.

### DIFF
--- a/changelog/changes/2023-03-08-systemd-elfutils.md
+++ b/changelog/changes/2023-03-08-systemd-elfutils.md
@@ -1,0 +1,1 @@
+- Enabled elfutils support in systemd-coredump. A backtrace will now appear in the journal for any program that dumps core ([coreos-overlay#2489](https://github.com/flatcar/coreos-overlay/pull/2489))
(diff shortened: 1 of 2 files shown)


jepio:
CI here: http://192.168.42.7:8080/job/container/job/packages_all_arches/1404/console (sorry for the IP, my DNS stopped allowing the normal entry).

jepio:
Manually tested the image and coredump now tries to prints a backtrace into the journal, nothing more.

dongsupark (approved):
:+1: 

Please add changelog.

Let's backport to Stable, Beta, Alpha, as the issue has been there since many versions, right?

PR #2490, opened 2023-02-28
Weekly profile updates and python-oem update

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/685/cldsv/

Needs to be merged together with https://github.com/flatcar/portage-stable/pull/423.

Blocked on ~https://github.com/flatcar/baselayout/pull/27 and~ ~https://github.com/flatcar/scripts/pull/672~.

Summary of changes made in this PR:
- profile updates
  - accept keyword updates mostly
  - mask for GCC dropped to allow an update to gcc 12
- added stage1 hook that updates baselayout in stage1
  - this is to get a proper setup of the PATH environment variable for the updated portage
  - portage dropped support for the ROOTPATH variable, which contained paths like `/usr/bin/` and `/bin`, so when it prepared an environment for running some helper bash scripts, the PATH variable in the environment was missing `/usr/bin` and `/bin`; as such, a script with shebang like `#!/usr/bin/env bash` failed to find bash.
  - the stage1 hook functionality is added to scripts in https://github.com/flatcar/scripts/pull/672
- dropped unnecessary user patch for gcc
- update python-oem to match the version in portage-stable
- pull in a newer version of baselayout
  - the newer version stops using ROOTPATH, and sets up PATH properly
  - the stage1 hook is still needed, though, as the stage1 is using an older version of baselayout, which breaks portage
- move sys-apps/texinfo to portage-stable
  - the modifications we made were pointless
    - something about CCACHE we stopped using
    - not building some portions of texinfo is not enough of a justification to keep maintaining our modifications, especially that none of the texinfo things end up in production image
- disable stack variable initialization hardening in kernel
  - it's not a change I like, but needed for now to keep cl.misc.falco test passing
  - for more details, please read the commit message
- glibc CET fixes
  - seed GCC has CET enabled, but for stage1 (and probably stage2) we build toolchains with CET disabled. Normally works except glibc, for which we need to explicitly pass a GCC option disabling the implicitly-enabled CET (-fcf-protection stuff).
  - will want to upstream it to Gentoo


krnowak:
Ready for review.

sayanchowdhury (approved):
LGTM 👍🏽

PR #2500, opened 2023-03-07
sys-apps/ignition: relabel created files (selinux)

tormath1:
In this PR, we pull:
* https://github.com/flatcar/bootengine/pull/57: adding `setfiles` to the initramfs, it's used by Ignition to relabel created files
* Added default Ignition upstream behavior to relabel files

Testing:
- [x] http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1442/cldsv/
- [x] Created a file with Ignition, assert it's labelled

Note:
- Reworked the translation patch, to map `oem` filesystems with regular randomly generated path (like others fs) - as stated in: https://github.com/coreos/ignition/issues/1452#issuecomment-1224386412

(cc @JAORMX)

### DIFF
--- a/sys-apps/ignition/files/0005-config-add-ignition-translation.patch
+++ b/sys-apps/ignition/files/0005-config-add-ignition-translation.patch
@@ -1,4 +1,4 @@
-From 54fa31c31ed431678adc89b6dfa450022d2f96fa Mon Sep 17 00:00:00 2001
+From 9596af9b8a3d1f999a8252a8a1a17f0bf6040f14 Mon Sep 17 00:00:00 2001
 From: Mathieu Tortuyaux <mtortuyaux@microsoft.com>
 Date: Mon, 26 Sep 2022 12:12:04 +0200
 Subject: [PATCH 05/18] config: add ignition translation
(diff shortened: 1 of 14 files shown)


sys-apps/ignition/files/0005-config-add-ignition-translation.patch
```diff
@@ -176,12 +176,8 @@ index 00000000..347d148c
 +
 +	if _, ok := fsMap[name]; !ok {
 +		addedSuffixCounter += 1
-+		if name == "oem" {
-+			fsMap[name] = "/usr/share/oem"
-+		} else {
-+			// generate a new path
-+			fsMap[name] = "/tmp/" + name + "-ign" + strconv.FormatUint(addedSuffixCounter, 10)
```
pothos, line 183: I don't understand why this is deleted (for the OEM filesystem we want to use the existing mount point)

pothos: Ah, I read the linked comment in the PR description… not really sure if the result would still be correct.
Reminds me that we have to move the `/usr/share/oem` mount point anyway to be `/oem` or similar because of systemd-sysext dropping it when run manually.

JAORMX (approved):

sys-apps/ignition/files/0005-config-add-ignition-translation.patch
pothos: This events of oem mount handling in the initrd are a bit messy currently, maybe your patch would even help a bit with the downside that it would have two mount points but I think that's fine!
I've tested it and found some unlabeled files/dirs but I think that's ok for now.

JAORMX: Normally, what kind of files are in `/usr/share/oem`? Probably that's not something you'd normally find in either the selinux policies coming from the selinux project nor the fedora one. so you might need a custom policy depending on the files that are there.

pothos: Ignition creates the `/etc/systemd/` folder and the preset file `/etc/systemd/system-preset/20-ignition.preset` unlabeled.

pothos: Maybe because policies are not loaded in the initrd?

pothos: With selinux not being loaded I would assume that every action in the initrd does not result in correct labels, or? And even if it would be loaded, the path `/sysroot/etc/…` is different from `/etc` and the labeling wouldn't work either. Maybe the setfiles step should be changed to cover the directories created on the way from `MkdirForFile`?
Anyway, to me it looks like `setfiles` being  a no op in this case because it gets called before the file gets created by `s.EnableUnit(unitString)`/`s.DisableUnit(unitString)`.

pothos: Ah, no, the reabeling is done in the and and uses the first missing part, then it should be ok.

pothos: Note: Using `/oem` as OEM mountpoint should work and it won't be a read-only path. Using /tmp/something is also ok.

PR #2506, opened 2023-03-10
use sysext in azure OEM images

krnowak:
app-emulation/wa-linux-agent: Start changing the ebuild to resemble a normal ebuild, without the OEM hacks. This replaces dependency on dev-lang/python-oem and dev-python/distro-oem with dev-lang/python and dev-python/distro, respectively.

dev-lang/python-oem, dev-python/distro-oem: Drop the unnecessary packages.

profiles, coreos-devel/board-packages: Drop things related to the dropped python packages. Maybe we should start pulling in dev-lang/python instead?

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1472/cldsv

Related scripts PR: https://github.com/flatcar/scripts/pull/683


app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
```diff
@@ -0,0 +1,229 @@
...
+         set_bin_files(data_files, dest=agent_bin_path)
+-        set_conf_files(data_files, dest="/usr/share/oem",
+-                       src=["config/coreos/waagent.conf"])
+         set_logrotate_files(data_files)
+         set_udev_files(data_files)
+-        set_files(data_files, dest="/usr/share/oem",
+-                  src=["init/coreos/cloud-config.yml"])
++        if int(version.split('.')[0]) > 3530:
++            set_conf_files(data_files, dest="/etc",
++                           src=["config/coreos/waagent.conf"])
```
pothos, line 218: The reason this file is not under `/usr` as part of the sysext images is that this file needs to be writable, or?

app-emulation/wa-linux-agent/wa-linux-agent-2.6.0.2-r1.ebuild
```diff
@@ -1,36 +1,40 @@
...
-	doins -r "${S}/azurelinuxagent/"
+PATCHES=(
+    "${FILESDIR}/0001-flatcar-changes.patch"
+)
 
-	insinto "/usr/share/oem"
+src_install() {
+	distutils-r1_src_install
+	rm -f "${D}/etc/waagent.conf"
+	insinto /etc
```
pothos, line 38: Where will this file end?

pothos:
While we have a workaround in Flatcar to support systemd units symlinks in `TARGET.wants/` for enabling a service, this is not upstream behavior and we should rather add a target drop-in unit with `Upholds=` - this will ensure that the service is started after a manual `systemd-sysext refresh`.

app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
krnowak: You know, this patch is something I'd like to eventually take to upstream. But I was not really sure what direction the changes there should take. Should setup.py in this case be only for us, the distro maintainers, or should it be usable also for people wanting to install a newer version of WALinuxAgent on their Flatcar instances. I'm not sure if these two use cases are reconcilable. I'm leaning towards the former, so maybe I should update the patch to install waagent.conf to `/usr/share/flatcar/etc`?

app-emulation/wa-linux-agent/wa-linux-agent-2.6.0.2-r1.ebuild
krnowak: That's a good question. This won't be a part of the generic production image and probably also not a part of sysext image (since we only take files from `/usr` into the raw image), so maybe I should actually be putting it in `/usr/share/flatcar/etc`. Or (better IMO to keep packages generic) will need an extra logic for the oem-sysexts, that moves `/etc` to `/usr/share/flatcar/etc` before creating the raw image. I think this could work if `/etc` is set up after sysexts, but not really sure if it works like that now - I have a bit of a chicken-and-egg feeling here…

pothos: If we know that this won't be written to, I would rather put this file under `/usr/share/waagent/` or one of the paths under `/usr` that this waagent already uses. But if that doesn't work I would we could split this up into two ebuild files by creating a new ebuild file for the generic image that installs only this file under `/etc` and our existing logic will move it to `/usr/share/flatcar/etc`.

pothos: We can also preserve the current behavior and copy it to `/etc` if it does not exist (`ExecStartPre` because tmpfiles won't pick the rule up)

pothos:
> While we have a workaround in Flatcar to support systemd units symlinks in `TARGET.wants/` for enabling a service, this is not upstream behavior and we should rather add a target drop-in unit with `Upholds=` - this will ensure that the service is started after a manual `systemd-sysext refresh`.

The services should also be put in the sysext. The files `coreos-base/oem-azure/files/units/waagent.service` and `coreos-base/oem-azure/files/base/base.ign` should be removed (assuming we use the upstream waagend.service definition, otherwise this should be a downstream patch?) and not installed to disk anymore.

krnowak:
There is also an issue with the `/usr/share/oem/oem-release` file. If `coreos-base/oem-azure` package is used solely for filling the sysext rootfs, then `/usr/share/oem/oem-release` will end up being inside sysext. Which means that [the sysext won't be even activated](https://github.com/flatcar/bootengine/blob/flatcar-master/dracut/99setup-root/initrd-setup-root-after-ignition#L11-L15).

I see two ways out of it:
- two packages, one for production image, one for sysext
  - we probably could even try having a generic `coreos-base/oem-common` (or `coreos-base/oem-release`) package that installs the `oem-release` file.
- one package + special sauce in scripts that creates the oem-release file with specific contents in the production image.

krnowak:
> * two packages, one for production image, one for sysext
>   * we probably could even try having a generic `coreos-base/oem-common` (or `coreos-base/oem-release`) package that installs the `oem-release` file.

I'll try this one first.

krnowak:
> While we have a workaround in Flatcar to support systemd units symlinks in `TARGET.wants/` for enabling a service, this is not upstream behavior and we should rather add a target drop-in unit with `Upholds=` - this will ensure that the service is started after a manual `systemd-sysext refresh`.

I'm not sure I understand, could you expand on it, please? Drop in for multi-user.target that adds `Upholds=waagent.service`?

pothos:
> I'm not sure I understand, could you expand on it, please? Drop in for multi-user.target that adds `Upholds=waagent.service`?

Yes, then `systemctl daemon-reload` done by `systemd-sysext` is enough to ensure that it runs. The wants-symlinks only work if we also restart all active targets which is a workaround I added for Flatcar but upstream recommends to use `Upholds` instead. We need to start the service after a `systemd-sysext refresh` because this temporarily unloads the service unit files, meaning the execution will be stopped.

Here an example:
```
# cat test.service 
[Unit]
Description=test
[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=echo hello

# cat multi-user.target.d/10-test-service.conf 
[Unit]
Upholds=test.service

# systemctl daemon-reload
# systemctl status test.service
```
This leads to test.service being started.

krnowak:
Cool, thanks! I appreciate it.

app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
krnowak: So, in the end I'm installing this file in `/usr/share/waagent/waagent.conf`. and in the `waagent.service` unit file I have an `ExecStartPre` line which creates a symlink in `/etc/waagent.conf` that points to `/usr/share/waagent/waagent.conf`.

app-emulation/wa-linux-agent/wa-linux-agent-2.6.0.2-r1.ebuild
krnowak: I went with `ExecStartPre` handling the waagent.conf stuff. I also placed our config into the patch for eventual upstreaming, so this whole `src_install` function is now gone.

pothos:
To migrate existing servers we need to add the list of old OEM files in `/etc/` (from the 'base' Ignition config) and the OEM partition to https://github.com/flatcar/bootengine/blob/d3cc0f4b1dce6a5084a8a909810efc30c367020b/dracut/99setup-root/initrd-setup-root-after-ignition#L59

krnowak:
> To migrate existing servers we need to add the list of old OEM files in `/etc/` (from the 'base' Ignition config) and the OEM partition to https://github.com/flatcar/bootengine/blob/d3cc0f4b1dce6a5084a8a909810efc30c367020b/dracut/99setup-root/initrd-setup-root-after-ignition#L59

Thanks, noted. I have not been yet thinking about the migration path. First wanted to get the fresh deployments to work.

PR #2542, opened 2023-03-20
sys-kernel/bootengine: Prevent OEM mount point from showing up early

pothos:
This pulls in
https://github.com/flatcar/bootengine/pull/58
to prevent the OEM mount point from showing up before Ignition mounts. It also makes it simpler to write files to the OEM partition without having to specify the partition first as initrd mount point.

Fixes https://github.com/flatcar/Flatcar/issues/979

## Testing done

See linked PR


pothos:
Backported as bugfix for flatcar-3510

PR #2553, opened 2023-03-27
sys-kernel/bootengine: fix journalctl --user issue

tormath1:
This pulls:
- https://github.com/flatcar/bootengine/pull/59
<hr>

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1573/cldsv/

### DIFF
--- a/changelog/bugfixes/2023-03-23-journalctl-user.md
+++ b/changelog/bugfixes/2023-03-23-journalctl-user.md
@@ -1,0 +1,1 @@
+- Fixed `journalctl --user` permission issue ([Flatcar#989](https://github.com/flatcar/Flatcar/issues/989))


krnowak:
You know, backporting to alpha doesn't make much sense, unless it's going to be promoted for beta.

pothos:
I think bugfixes are always fine for a branched Alpha and Beta. For Beta in this case it involves using the backport branch, however. For Stable it depends since there is some risk in directly committing such a backport without prior testing - I would rather not do it since it's not a regression in the Stable's major version because this stuff was broken since 2020.

pothos:
Picked it for the Alpha to make it easier to get the other bootengine bug fix in, too.

PR #2558, opened 2023-03-30
Use tmpfiles to set up /var/log/journal/ from initrd

pothos:
This pulls in https://github.com/flatcar/bootengine/pull/60 and https://github.com/flatcar/baselayout/pull/29 to fix the persisting of the journal on first boot.

## How to use

Backport to Alpha

## Testing done

See linked bootengine PR


krnowak (commented):
I'm not sure that the change in bootengine is necessary. Also, the PR probably will need to wait until the "code freeze" is lifted.

pothos:
I think we are good to merge because the freeze hasn't started yet

krnowak (approved):

PR #2564, opened 2023-04-04
/usr/share/oem -> /oem

krnowak:
TODO


krnowak:
Will file a PR to scripts after Big Merge.

PR #2580, opened 2023-05-29
Update k8s binaries s3 paths

MosheFriedland:
# Update k8s binaries s3 paths

Updated k8s binaries s3 paths to new ones

## How to use
Pass different cluster version

## Testing done

Verified the paths and their contents. Didn't try to create nodes and connect them to an EKS cluster (should I?)


NadavYahav (approved):

MosheFriedland:
Unfortunately, I now see that it requires more than just adding these paths in order to be able to support running with K8S versions >=1.24.
Closing this PR

